PR #39, opened 2019-02-05
[RFR] Create UDF functions for Excel

SimonBiggs:
Creates user defined functions for Excel using xlwings. See xlwings documentation for more details:

https://docs.xlwings.org/en/stable/udfs.html


SimonBiggs:
Thanks Matt :)

PR #44, opened 2019-02-14
[RFR] Add docker machine images

SimonBiggs:
Will explain this more in a bit. But essentially I am creating a docker orthanc server that receives DICOM files, manipulates the header, and then forwards the DICOM files onwards to another server.

Specifically I am manipulating the electron density tags within the DICOM structure set so as to attach relative electron densities for the couch structures for being sent to DoseCHECK. I am also manipulating the machine name so that the machine name matches that within DoseCHECK.


src/pymedphys/dcm/_level2/header_tweaks.py
```diff
@@ -0,0 +1,119 @@
...
+        try:
+            if value == getattr(item, key):
+                new_sequence.pop(i)
+        except AttributeError:
+            pass
+
+    return new_sequence
+
+
+def adjust_rel_elec_density(dcm, adjustment_map):
```
SimonBiggs: Monaco doesn't export the electron densities of a couch. I needed to create a DICOM server to stand between Monaco and DoseCHECK and append the electron densities of the couch to the structure set.

src/pymedphys/docker/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+import os
+import subprocess
+
+
+HERE = os.path.dirname(__file__)
+
+
+def server_cli(args):
```
SimonBiggs: This boots up the orthanc docker server by running the command:

```bash
pymedphys docker server
```

src/pymedphys/docker/nginx/nginx.conf
```diff
@@ -0,0 +1,87 @@
+events {
+    worker_connections 1024;
+}
+
+stream {
+    upstream backend {
+        server rccc-physicssvr:104;
```
SimonBiggs: There are some RCCC specific items in here. In the long term I desire to pull these out into a configuration file, which can then be edited via a GUI.

src/pymedphys/docker/orthanc/scripts/morph_header_and_forward.lua
```diff
@@ -0,0 +1,50 @@
...
+      local target = assert(io.open(received_filepath, 'wb'))
+      target:write(dicom)
+      target:close()
+
+      RestApiDelete('/instances/' .. instanceId)
+    end
+
+    if tags['SOPClassUID'] == structure_UID then
+      print('Converting a structure set')
+      os.execute('pymedphys dicom adjust-rel-elec-density ' .. received_filepath .. ' ' .. converted_filepath .. ' "Couch Edge" 1.1 "Couch Foam Half Couch" 0.06 "Couch Outer Half Couch" 0.5')
```
SimonBiggs: Here are some hard coded couch values. I'm still finding my feet with the Lua language which is what Orthanc uses for scripting. In the future these hardcoded values need to be pulled out into a configuration file.

test-watch.py
```diff
@@ -7,7 +7,7 @@ def main():
     except FileNotFoundError:
         pass
 
-    os.system('ptw -- --testmon')
+    os.system('ptw -- --testmon -vv --maxfail=1')
```
SimonBiggs: The test driven development was being slowed down by having the tests continue after one was found to fail. This helped the process, and also gave more verbose debugging printouts.

tests/test_dcm/test_header_tweaks.py
```diff
@@ -0,0 +1,208 @@
...
+    adjusted_dicom_file = adjust_machine_name(original_dicom_file, new_name)
+
+    assert adjusted_dicom_file != original_dicom_file
+    assert adjusted_dicom_file == expected_dicom_file
+
+    pydicom.write_file(ORIGINAL_DICOM_FILENAME, original_dicom_file)
+    subprocess.check_call(
+        'pymedphys dicom adjust-machine-name'.split() +
+        [ORIGINAL_DICOM_FILENAME, ADJUSTED_DICOM_FILENAME, new_name]
+    )
```
SimonBiggs: Testing the CLI as well as the python interface.

tests/test_dcm/test_header_tweaks.py
```diff
@@ -0,0 +1,208 @@
...
+                'ROIName': 'to_be_changed 3'
+            },
+        ],
+        'RTROIObservationsSequence': [
+            {
+                'ReferencedROINumber': 1,
+                'ROIPhysicalPropertiesSequence': [
+                    {
+                        'ROIPhysicalProperty': 'EFFECTIVE_Z',
+                        'ROIPhysicalPropertyValue': 6
```
SimonBiggs: Want to make sure this isn't overwritten.

tests/test_dcm/test_header_tweaks.py
```diff
@@ -0,0 +1,208 @@
...
+            },
+            {
+                'ReferencedROINumber': 10,
+            },
+            {
+                'ReferencedROINumber': 99,
+                'ROIPhysicalPropertiesSequence': [
+                    {
+                        'ROIPhysicalProperty': 'REL_ELEC_DENSITY',
+                        'ROIPhysicalPropertyValue': 0
```
SimonBiggs: Want to make sure that this is overwritten.

tests/test_dcm/test_header_tweaks.py
```diff
@@ -0,0 +1,208 @@
...
+        + adjustment_map_flat)
+
+    print(command)
+
+    subprocess.check_call(command)
+
+    cli_adjusted_dicom_file = pydicom.read_file(
+        ADJUSTED_DICOM_FILENAME, force=True)
+
+    assert str(expected_dicom_file) == str(cli_adjusted_dicom_file)
```
SimonBiggs: By using `str` here pytest produces a really nice diff print out displaying where the DICOM files are different.

tests/test_dcm/test_header_tweaks.py
Centrus007: I think a broader "dicom edit" tool for any tag might be more useful here. Not urgent or necessary now - just a thought :)

Centrus007:
LGTM.

The DICOM tag editing functions would be better generalised, I think. But something to add later - not necessary now :)

SimonBiggs:
Thanks Matt :)

SimonBiggs:
> The DICOM tag editing functions would be better generalised, I think. But something to add later - not necessary now :)

I think that's true, but I can't quite think how to do it neatly. For example machine name is nested under every beam, it's not just one item. Also electron density editing is quite specific, as in it needs to be specifically added corresponding to a structure name on another dicom structure.

Yeah, I won't be giving it a shot in the near future. Might be easier to see how to generalise once a few more use cases come out.

Something I would like to do is "anonymised all dicom files within a directory into a new directory" CLI function.

PR #54, opened 2019-02-16
[WIP] 50: tidy up dicom dose and coords

Centrus007:
Okay, @SimonBiggs, I don't expect you to get to this any time soon, but that's okay. It might be a WIP for a bit while we hash out a few things, but I'd really like your input as this progresses.

Summary of changes:
- I think there are fairly strong advantages (logistically and structurally) to encapsulating as many aspects of a DICOM Dose file as potentially useful into a single data type. I.e. an object within which all useful parameters are directly and conveniently accessible. The user could simply pass this object to something like a gamma function and not mess around with individual elements (like dose values separated from coordinates) that are unnecessarily separated in code. To this end, I'm tentatively introducing a new DICOM Dose object constructed with a simple DICOM RT Dose filepath.
- I've renamed all of the coordinate extraction functions to `extract_xyz` rather than `extract_coords`. The "coordinates" those functions extract are not *really* coordinates, but simply the axis values that constitute the DICOM dose grid. I'm thinking the word "coordinates" is better reserved for actual *coordinate triplets* (in 3D). E.g. in the case of DICOM dose files, a `coords` variable should include one coordinate triplet for each dose voxel. Ideally, for a dose grid of shape *(k, j, i)*, I think coords ought to have shape *(3, k, j, i)*,  or perhaps at least be an *(x, y, z)* tuple whose elements have shape *(k, j, i)*. In that way the *x* coordinate of the dose voxel at  index *(a, b, c)* would either be `coords[0, a, b, c]`, or `coords[0][a, b, c]`
- To that end, I created a `convert_xyz_to_dicom_coords` function. It returns `coords`, a tuple of x, y and z meshgrids that (should - still fairly untested) such that the *x*, *y*, and *z* coordinates for the dose voxel at `dcmdose.values[23, -13, 12]` are simply `dcmdose.coords[0][23, -13, 12]`, `dcmdose.coords[1][23, -13, 12]` & `dcmdose.coords[2][23, -13, 12]` respectively.
   - I'm relatively sure that these, when flattened, will play quite nicely with the coordinate transforms in `coordinatestransforms.py`. A neat way to get coordinates in other coordinate systems, I think.
- I updated the troublesome `test_dcmdose` test functions to have `.json` baselines. Relatively sure all is in order there.


src/pymedphys/dcm/_level1/dcmdose.py
```diff
@@ -735,17 +658,3 @@ def contour_to_points(contours):
     contour_points = np.concatenate(resampled_contours, axis=1)
 
     return contour_points
-
-
-def check_dcmdose(dcm):
```
Centrus007: Unneccessary inclusion. `pydicom` already includes checks like this. A simple check for modality is fine, i think.

src/pymedphys/dcm/_level1/dcmdose.py
```diff
@@ -508,7 +431,7 @@ def extract_dicom_patient_coords(dcm):
     else:
         z = np.flip(-position[2] + np.array(dcm.GridFrameOffsetVector))
 
-    return Coords(x, y, z)
```
Centrus007: Not much to be gained from the nametuple, I don't think.

src/pymedphys/dcm/_level1/dcmdose.py
```diff
@@ -76,104 +96,6 @@ def load_dose_from_dicom(dcm, set_transfer_syntax_uid=True, reshape=True):
     return dose
 
 
-def extract_dose(dcm):
```
Centrus007: Effectively replaced by `DicomDose` class

src/pymedphys/dcm/_level1/dcmdose.py
```diff
@@ -41,12 +40,33 @@
...
+        dcm = pydicom.dcmread(dcm_filepath)
+        if dcm.Modality != "RTDOSE":
+            raise ValueError("The input DICOM file is not an RT Dose file")
+        self.dcm = dcm
+
+        self.values = dcm.pixel_array * dcm.DoseGridScaling
+        self.units = dcm.DoseUnits
+        self.x, self.y, self.z = extract_dicom_patient_xyz(dcm)
+        self.coords = convert_xyz_to_dicom_coords((self.x, self.y, self.z))
+        self.mask = None
```
Centrus007: Meant for masking dose by an ROI from a DICOM RT Structure File. Code (eventually) forthcoming.

src/pymedphys/dcm/_level1/dcmdose.py
```diff
@@ -41,12 +40,33 @@
...
+    def __init__(self, dcm_filepath):
+        dcm = pydicom.dcmread(dcm_filepath)
+        if dcm.Modality != "RTDOSE":
+            raise ValueError("The input DICOM file is not an RT Dose file")
+        self.dcm = dcm
+
+        self.values = dcm.pixel_array * dcm.DoseGridScaling
+        self.units = dcm.DoseUnits
+        self.x, self.y, self.z = extract_dicom_patient_xyz(dcm)
+        self.coords = convert_xyz_to_dicom_coords((self.x, self.y, self.z))
```
Centrus007: May add a `coord_system` attribute later, with some native function to convert `self.coords` and track the current system. Being always able to refer to `self.ds` for patient orientation.

src/pymedphys/dcm/_level1/dcmdose.py
```diff
@@ -41,12 +40,33 @@
...
 IMPORTS = get_imports(globals())
 
-Coords = namedtuple('Coords', 'x y z')
-Dose = namedtuple(
-    'Dose', 'values units type summation heterogeneity_correction')
-
 # pylint: disable=C0103
 
+class DicomDose:
+
```
Centrus007: docstring forthcoming

src/pymedphys/dcm/_level1/dcmdose.py
SimonBiggs: Some of the code to achieve this should be able to be found at https://github.com/pymedphys/pymedphys/blob/master/src/pymedphys/dcm/_level1/dcmdose.py#L654-L686

src/pymedphys/dcm/_level1/dcmdose.py
```diff
@@ -41,12 +40,32 @@
...
+    def __init__(self, dcm_filepath):
+        dcm = pydicom.dcmread(dcm_filepath)
```
SimonBiggs: One preference on my end, have it by default be initialised from a pydicom object, not file path. But then include a `from_filepath` `@classmethod`

This would mean the following two methods would be possible to create this object:

```
dicom_dose = DicomDose(pydicom_object)
dicom_doss = DicomDose.from_filepath(dicom_filepath_string)
```

src/pymedphys/dcm/_level1/dcmdose.py
```diff
@@ -365,7 +286,8 @@ def extract_iec_fixed_coords(dcm):
        https://arxiv.org/ftp/arxiv/papers/1406/1406.0014.pdf
     """
 
-    check_dcmdose(dcm)
+    if dcm.Modality != "RTDOSE":
+        raise ValueError("The input DICOM file is not an RT Dose file")
```
SimonBiggs: This here seems repeated a few times. I think pulling this out into a function would be of value to keep things DRY.

src/pymedphys/dcm/_level1/dcmdose.py
SimonBiggs: Happy to lean on your judgement here.

tests/data/dcmdose/expected_dicom_patient_xyz.json
```diff
@@ -0,0 +1 @@
...000005, 32.548100000000005, 34.548100000000005, 36.548100000000005, 38.548100000000005, 40.548100000000005, 42.548100000000005, 44.548100000000005, 46.548100000000005, 48.548100000000005, 50.548100000000005, 52.548100000000005, 54.548100000000005, 56.548100000000005, 58.548100000000005], [-1259.5, -1257.5, -1255.5, -1253.5, -1251.5, -1249.5, -1247.5, -1245.5, -1243.5, -1241.5, -1239.5, -1237.5, -1235.5, -1233.5, -1231.5, -1229.5, -1227.5, -1225.5, -1223.5, -1221.5, -1219.5, -1217.5, -1215.5, -1213.5, -1211.5, -1209.5, -1207.5, -1205.5, -1203.5, -1201.5, -1199.5, -1197.5, -1195.5, -1193.5, -1191.5, -1189.5, -1187.5, -1185.5, -1183.5, -1181.5, -1179.5, -1177.5, -1175.5, -1173.5, -1171.5, -1169.5, -1167.5, -1165.5, -1163.5, -1161.5, -1159.5, -1157.5, -1155.5, -1153.5, -1151.5, -1149.5, -1147.5, -1145.5, -1143.5, -1141.5, -1139.5, -1137.5, -1135.5, -1133.5, -1131.5, -1129.5, -1127.5, -1125.5, -1123.5, -1121.5, -1119.5, -1117.5, -1115.5, -1113.5, -1111.5, -1109.5, -1107.5, -1105.5, -1103.5, -1101.5, -1099.5, -1097.5, -1095.5, -1093.5, -1091.5, -1089.5, -1087.5, -1085.5, -1083.5, -1081.5, -1079.5, -1077.5, -1075.5, -1073.5, -1071.5, -1069.5, -1067.5, -1065.5, -1063.5, -1061.5, -1059.5]]}
```
SimonBiggs: When making json files it's a good idea to include the indent parameter, say `indent=2`. Makes the file much more human readable :). 

Copy pasting this into http://jsonprettyprint.com should do this now that the file is already made though.

tests/test_dcm/test_dcmdose.py
```diff
@@ -23,123 +23,147 @@
...
+            'values': test_dcmdose.values.tolist(),
+            'units': test_dcmdose.units,
+            'x': test_dcmdose.x.tolist(),
+            'y': test_dcmdose.y.tolist(),
+            'z': test_dcmdose.z.tolist(),
+            'coords': test_dcmdose.coords.tolist(),
+            'mask': test_dcmdose.mask
+        }
+        with open(baseline_dcmdose_dict_filepath, 'w') as fp:
+            json.dump(expected_dcmdose_dict, fp)
```
SimonBiggs: `indent=2` helpful here (see other comment)

SimonBiggs:
> I think there are fairly strong advantages (logistically and structurally) to encapsulating as many aspects of a DICOM Dose file as potentially useful into a single data type. I.e. an object within which all useful parameters are directly and conveniently accessible. 

After reading your code a little and thinking about it a bit I agree with you. It makes a lot of sense as a mental model also. Attaching related functions as methods onto this DicomDose object would be of value also.

For example, one should be able to write `DicomDose.gamma(3,3)` for a 3%, 3mm gamma comparison.

src/pymedphys/dcm/_level1/dcmdose.py
SimonBiggs: Also, as a protocol for resolving discussions, for me this is resolved, bit if a click resolved I fear you might miss the comment. So maybe, we only click resolve if the other person was the last to comment? What do you think?

src/pymedphys/dcm/_level1/dcmdose.py
SimonBiggs: When this happens, my preference is that the coords transformation function returns a new DicomDose object instead of making this change you refer to inplace.

src/pymedphys/dcm/_level1/dcmdose.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2016-2018 Simon Biggs, Matthew Jennings
+# Copyright (C) 2016-2019 Matthew Jennings and Simon Biggs
```
SimonBiggs: Worth updating the copyright like this when you make a significant contribution to a file.

src/pymedphys/dcm/_level1/dcmdose.py
```diff
@@ -41,12 +40,32 @@
...
+        self.dcm = dcm
+
+        self.values = dcm.pixel_array * dcm.DoseGridScaling
+        self.units = dcm.DoseUnits
+        self.x, self.y, self.z = extract_dicom_patient_xyz(dcm)
+        self.coords = convert_xyz_to_dicom_coords((self.x, self.y, self.z))
+        self.mask = None
+
+    def save_to_dicom(self, filepath):
+        self.dcm.save_as(filepath)
```
SimonBiggs: Maybe just call this function `save_as`? And, I think there is value in thinking of this object as a pydicom object with dose specific extras. So maybe we should expose the pydicom class methods that we envisage useful. `save_as` is certainly one.

What do you think about creating a base dicom wrapper class that exposes the pydicom methods that we want, including having utilities like from_file etc. It would also include some of our other helper methods like from_dict and to_dict. Then we create DicomDose, DicomStructure etc that inherits from our base PyDicomWrapper object. PyDicomWrpper should not inherit from anything within pydicom, just be initialised with a pydicom object and then forward through the pydicom methods.

Let me know what you think :). It's seeming quite nice in my head.

src/pymedphys/dcm/_level1/dcmdose.py
Centrus007: Seems reasonable. Do you want to try resolving one of these just to see how it manifests?

src/pymedphys/dcm/_level1/dcmdose.py
Centrus007: Yep, you're absolutely right - I had it in mind while writing `convert_xyz_to_dicom_coords()`.

src/pymedphys/dcm/_level1/dcmdose.py
Centrus007: I don't think I mind too much, but why is that? That would be less memory efficient and the original version of the object would be easily retrievable through a simple conversion back.

Basically, it seems like a mutable thing to me...

src/pymedphys/dcm/_level1/dcmdose.py
Centrus007: Fair enough. I forgot to mention it but I did think it was worth supporting both.

I'm not sure I see a scenario where you wouldn't call `DicomDose()` from a file. It seems unnecessary to force to user to perform a `pydicom.dcmread()` first, when the entire `dcm` is stored within the `DoseObject` anyway.

src/pymedphys/dcm/_level1/dcmdose.py
Centrus007: Alright :)

tests/data/dcmdose/expected_dicom_patient_xyz.json
Centrus007: I can adjust easily enough :)

src/pymedphys/dcm/_level1/dcmdose.py
SimonBiggs: People using the object not understanding the difference between mutable and immutable can fall into some really awkward bugs when working with something mutable and assuming it is immutable.

For example:

```python
dose_dicom1 = DoseDicom(...)
dose_dicom2 = DoseDicom(...)

collection = [dose_dicom1, dose_dicom2]

collection[0].convert_coords(something)

dose_dicom1.dosomething # (but now it has changed)
```

src/pymedphys/dcm/_level1/dcmdose.py
SimonBiggs: When using it within our own library we might want to pass it around. As a preference thing I sort of picture it as extending an object that already exists, so it feels like it should compose with that original object :).

tests/data/dcmdose/expected_dicom_patient_xyz.json
SimonBiggs: Another thing that really helps is `np.round(array, decimals=4).tolist()` instead of going straight to list. That fixes up those floating point deviations.

src/pymedphys/dcm/_level1/dcmdose.py
SimonBiggs: I made that mistake with the standard pydicom object, that's why I needed to put a deepcopy on that template object.

With the RAM point, I imagine conversion of coordinates should only ever need to be done once or twice to an object. So shouldn't be too much of an issue here.

SimonBiggs:
Also, I didn't say, and I should have. This is an awesome idea :). Thanks Matt. Awesome work :).

SimonBiggs:
I think we should merge this in and you can iterate further in another branch?

src/pymedphys/dcm/_level1/dcmdose.py
Centrus007: I'm don't think I understand your first point re: passing the object around, but I don't really have a preference. 

Do you envisage a situation where a DicomDose object could be constructed from a pydicom Dataset that did not come from a file? If so, I think your suggestion to construct from a pydicom Dataset (rather than a file) makes sense.

src/pymedphys/dcm/_level1/dcmdose.py
Centrus007: I don't think that's a reason to insist on immutable objects. That exact same point could be made for many python objects that we regularly use throughout the PyMedPhys code, like lists and numpy arrays.

To be honest, in the code snippet you provided, I would *expect* `dose_dicom1` to have changed. Indeed, with code like that, I would've *intended* for it to change. I actually think the immutable case is the more confusing .

The mutable version better matches intuition. When my car has an empty fuel tank and I go to the petrol station to fill it up, I don't discard my empty-tanked car for a replica with a full tank!

> With the RAM point, I imagine conversion of coordinates should only ever need to be done once or twice to an object. So shouldn't be too much of an issue here.

Perhaps you're right, but I think it's a bit brave to assume this. I can absolutely see a batch processing application here...

src/pymedphys/dcm/_level1/dcmdose.py
Centrus007: Will do! :)

src/pymedphys/dcm/_level1/dcmdose.py
Centrus007: I don't think that's a reason to insist on immutable objects. That exact same point could be made for many python objects that we regularly use throughout the PyMedPhys code, like lists and numpy arrays.

The mutable version better matches intuition. When my car has an empty fuel tank and I go to the petrol station to fill it up, I don't discard my empty-tanked car for a replica with a full tank!

> With the RAM point, I imagine conversion of coordinates should only ever need to be done once or twice to an object. So shouldn't be too much of an issue here.

Perhaps you're right, but I think it's a bit brave to assume this. I can absolutely see a batch processing application here...

---

I think I might have a way to circumvent both of our concerns.

It does make sense to perform a deepcopy on the `dcm` at DicomDose construction, but then I'd still prefer to keep it mutable from there. In that way, the following isn't a risk:

```python
dcm = pydicom.dcm_read("my_dicom_file.dcm")

dcmdose = DicomDose(dcm)
dcmdose.convert_coords(...)

dcm.do_something() # but dcm has changed as a result of methods called on dcmdose
```

If `dcmdose.dcm` (i.e. the pydicom Dataset used to construct the DicomDose object) is copied before saving it as an attribute, the above will not occur (`dcm` will not have changed).

I should then be okay for methods like `dcmdose.convert_coords()` to simply edit `dcmdose.dcm` rather than return a deepcopy of `dcmdose.dcm` with the edits. There is no risk of the above confusion *and* you avoid memory waste and possible associated performance issues (e.g. in batch processing).

SimonBiggs:
Just replying by email. That logic seems good to me. In my botched pull
request from the other branch I did something like that. Essentially made a
deep copy on property assignment.

On this end I was mostly declaring preference. Feel free to override with
your judgement.

src/pymedphys/dcm/_level1/dcmdose.py
SimonBiggs: One case off the top of my head, it would be nice to be able to extend "dcm from dict" to be able to have a bumpy array within the pixel array dictionary field. Then we could make a DicomDose object from a dictionary.

PR #55, opened 2019-02-17
[RFR] Beginning work on DICOM Collection class

SimonBiggs:
Here is a quick mock up of what a base `Dicom` class might look like. @Centrus007 keen for your thoughts.

```ipython
In [1]: from pymedphys.dcm import Dicom

In [2]: dicom = Dicom.from_dict({
    'Manufacturer': 'PyMedPhys', 
    'PatientName': 'Python^Monte'
})

In [3]: dicom
Out[3]: 
(0008, 0070) Manufacturer                        LO: 'PyMedPhys'
(0010, 0010) Patient's Name                      PN: 'Python^Monte'

In [4]: dicom.anonymise()
Out[4]: 
(0008, 0070) Manufacturer                        LO: 'PyMedPhys'
(0010, 0010) Patient's Name                      PN: ''
```


src/pymedphys/dcm/_level2/dcmanonymise.py
```diff
@@ -64,21 +64,12 @@ def anonymise_dicom(dcm, delete_private_tags=True, tags_to_keep=None,
...
-        If `dcm` is not an instance of `pydicom.dataset.FileDataset`
+        An anonymised copy of the input DICOM file as a `pydicom Dataset`
     """
 
     if tags_to_keep is None:
         tags_to_keep = []
 
-    if not isinstance(dcm, pydicom.dataset.FileDataset):
-        raise TypeError("The input argument is a member of {}. "
-                        "It must be a pydicom FileDataset.".format(type(dcm)))
```
SimonBiggs: This excludes the ability of using `pydicom.Dataset`. Is it okay if I remove this raise?

SimonBiggs: I also believe that `pydicom.FileDataset` is a superset of `pydicom.Dataset`...

SimonBiggs:
@Centrus007 I'm okay if this doesn't get merged/used. Up to you if you think it is a good way to go about it or not.

SimonBiggs:
Gee, I really did that one too quick. A few test failures there :/

src/pymedphys/dcm/_level2/dcmanonymise.py
Centrus007: Yep no worries at all. I added that before I had a clear understanding of pydicom's structure.

src/pymedphys/dcm/_level2/dcmstruct.py
```diff
@@ -32,14 +32,72 @@
...
+        np.array(item[1::3])
+        for item in contours_by_slice_raw]
+    z = [
+        np.array(item[2::3])
+        for item in contours_by_slice_raw]
+
+    return x, y, z
+
+
+def pull_structure(string, dcm_struct):
```
SimonBiggs: I shifted structure logic into the level 2 structure dicom file, it made more sense there anyway. That freed me up to have dicom dose depend on the structure file instead of the other way round. If I didn't move that logic I would have had to make a level 5 as well as level 4. We are free to make as many levels as needed, but often some simple refactoring can lessen the number of levels.

This would also probably be why the tests are failing. In my haste between packing I didn't actually transfer any of the imports across between the files. Oops.

src/pymedphys/dcm/_level3/pydicom_wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @property
+    def dcm(self) -> pydicom.Dataset:
+        return deepcopy(self._dcm)
+
+    @dcm.setter
+    def dcm(self, dcm: pydicom.Dataset):
+        self._dcm = deepcopy(dcm)
+
+    def save_as(self, filepath):
```
SimonBiggs: For reasons of symmetry, I actually think this might be better named `to_file`. I think that's neater than forcing ourselves to go lockstep with pydicom's API.

src/pymedphys/dcm/_level3/pydicom_wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+        self._dcm.save_as(filepath)
+
+    def anonymise(self):
+        anonymised = anonymise_dicom(self.dcm)
+
+        return self.__class__(anonymised)
```
SimonBiggs: By calling class here instead of `Dicom` this method should appropriately undergo inheritance. As in, if DicomDose inherits from this, an anonymised DicomDose object will be returned instead of a plain Dicom object.

src/pymedphys/dcm/_level3/pydicom_wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+        return cls(dcm)
+
+    def __repr__(self):
+        return self._dcm.__repr__()
+
+    @property
+    def dcm(self) -> pydicom.Dataset:
+        return deepcopy(self._dcm)
+
+    @dcm.setter
```
SimonBiggs: By using a getter setter property I can make it so whenever dcm is set or retrieved it goes via a deepcopy. I think this is in line with your compromise between mutable and immutable. I'm happy to agree on that point.

src/pymedphys/dcm/_level3/pydicom_wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+        return cls(dcm)
+
+    @classmethod
+    def from_dict(cls, dictionary):
+        dcm = dcm_from_dict(dictionary)
+
+        return cls(dcm)
+
+    def __repr__(self):
+        return self._dcm.__repr__()
```
SimonBiggs: This here forwards through how the class is represented to pydicom's default method.

src/pymedphys/dcm/_level3/pydicom_wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @classmethod
+    def from_file(cls, filepath):
+        dcm = pydicom.read_file(filepath, force=True)
+
+        return cls(dcm)
+
+    @classmethod
+    def from_dict(cls, dictionary):
+        dcm = dcm_from_dict(dictionary)
```
SimonBiggs: As mentioned elsewhere, I could imagine extending this method to also be able to create pixel_array objects on the pydicom object. I think that would be a really neat way of making simple little transperant test DicomDose objects. It's really helped TTD development to be able to quickly and easily make dcm objects with a dictionary.

src/pymedphys/dcm/_level3/pydicom_wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+IMPORTS = get_imports(globals())
+
+
+class Dicom():
+    def __init__(self, dcm):
+        self.dcm = dcm
+
+    @classmethod
+    def from_file(cls, filepath):
+        dcm = pydicom.read_file(filepath, force=True)
```
SimonBiggs: By having this as a class method it means it is called by:

```
dicom = Dicom.from_file('filepath')
```

layers.yml
```diff
@@ -20,6 +20,7 @@ Internal Levelled Modules:
         - pymedphys.utilities
         - pymedphys.xarray
     layers:
+        - (_level4)
```
SimonBiggs: I needed to add an extra level here because this was the first time a module needed a level4

SimonBiggs:
You'll also not there's no testing in this pull request. I pictured writing a test much like the example I gave at the top. You'll notice it wasn't written first :(. I rush and I revert to old habits.

src/pymedphys/dcm/_level2/dcmstruct.py
Centrus007: Gotcha.

src/pymedphys/dcm/_level3/pydicom_wrapper.py
Centrus007: Agreed. Or even `read_from_file()` / `write_to_file`?

src/pymedphys/dcm/_level3/pydicom_wrapper.py
Centrus007: Cool :)

src/pymedphys/dcm/_level3/pydicom_wrapper.py
Centrus007: I think so, so long as we don't enforce that object methods use these getter/setters.

src/pymedphys/dcm/_level3/pydicom_wrapper.py
Centrus007: Yep, I can totally see this.

Centrus007:
> You'll also not there's no testing in this pull request. I pictured writing a test much like the example I gave at the top. You'll notice it wasn't written first :(. I rush and I revert to old habits.

Hahaha, I'll forgive you! :)

Centrus007:
I'm quite liking this approach! I'll leave it unmerged until the tests pass.

src/pymedphys/dcm/_level3/pydicom_wrapper.py
SimonBiggs: Yup certainly. Object methods can use the private `self._dcm` to bypass this deepcopy.

src/pymedphys/dcm/_level3/pydicom_wrapper.py
SimonBiggs: Yup, sounds good to me.

SimonBiggs:
> I'm quite liking this approach! I'll leave it unmerged until the tests pass.

I think the fix should just be moving some dependencies across. I don't think I'll get a chance to touch it for a few days. If you wanted to take it across the line that wouldn't bother me.

Centrus007:
> I think the fix should just be moving some dependencies across. I don't think I'll get a chance to touch it for a few days. If you wanted to take it across the line that wouldn't bother me.

No worries; I'll see how I go for time but will probably give it a go. :)

src/pymedphys/dicom/_level2/dose.py
```diff
@@ -37,32 +37,14 @@
...
-        self.values = ds.pixel_array * ds.DoseGridScaling
-        self.units = ds.DoseUnits
-        self.x, self.y, self.z = extract_dicom_patient_xyz(ds)
-        self.coords = convert_xyz_to_dicom_coords((self.x, self.y, self.z))
-        self.mask = None
-
-    def save_to_dicom(self, filepath):
-        self.ds.save_as(filepath)
-
-
```
SimonBiggs: Moved this function to `dicom_collection.py`

src/pymedphys/dicom/_level2/header_tweaks.py
```diff
@@ -31,28 +31,29 @@
...
 IMPORTS = get_imports(globals())
 
 
-def adjust_machine_name(ds, new_machine_name):
+def adjust_machine_name(dicom_dataset, new_machine_name):
     """Change the machine name within the DICOM header
     """
 
-    new_ds = deepcopy(ds)
+    new_dicom_dataset = deepcopy(dicom_dataset)
```
SimonBiggs: I'm trialling being more verbose `ds` --> `dicom_dataset`

src/pymedphys/dicom/__init__.py
```diff
@@ -29,7 +29,7 @@
 Available Functions
 -------------------
 >>> from pymedphys.dicom import (
-...     anonymise_dicom,
+...     anonymise_dicom_dataset,
```
SimonBiggs: @Centrus007 Up to you if you are happy or not with this name change. Let me know.

src/pymedphys/dicom/_level3/dicom_collection.py
```diff
@@ -0,0 +1,111 @@
...
+    pass
+
+
+class DicomCollection:
+    """A DICOM collection can hold one DICOM CT, one DICOM structure, one
+    DICOM plan and one DICOM dose. These DICOM files must align via UID
+    declaration.
+
+    Not all types are required to create a DicomCollection.
+    """
```
SimonBiggs: More thinking required on how best to implement this object...

SimonBiggs:
Some methods on DicomCollection that I believe would make sense:

* from_directory
* to_directory
* from_zip
* to_zip
* add_file
* add_dataset

The DICOM files would be assigned to their relative DicomObjects within the DicomCollection based on their header information.

We should then also have simpler non-dicom plan, structure, dose, and CT objects. Each of these objects can be added to the DicomCollection by their respective functions and a UID and other needed DICOM header tags would be created:

* add_dose(dose)
* add_plan(plan)
* add_CT(CT)
* add_structures(structures)

src/pymedphys/dicom/__init__.py
Centrus007: Yep that's better.

I'm imagining a dicom anonymise cli that calls a `anonymise_dicom_file` with a filepath parameter, which runs `pydicom.dcmread()` on the filepath and calls this.

src/pymedphys/dicom/_level2/header_tweaks.py
Centrus007: All good.

Is `dataset` not sufficient? We are working in the `dicom` pymedphys space. Avoiding redundant labelling is a specific recommendation in the [Python Best of the Best Practices Github guide](https://gist.github.com/sloria/7001839)...

But honestly, I don't really care at all. It would be a stretch to say `dataset` vs. `dicom_dataset` is even a very slight preference of mine.

Centrus007: Somewhat relevant: presumably we'll be using `dataset` (rather than `dicom_dataset`) in our custom Dicom classes?

src/pymedphys/dicom/_level3/dicom_collection.py
```diff
@@ -0,0 +1,111 @@
...
+
+class DicomStructure(DicomBase):
+    pass
+
+
+class DicomPlan(DicomBase):
+    pass
+
+
+class DicomCollection:
```
Centrus007: In the interest of maintaining nomenclature parallel to the DICOM Standard, would this correspond to a Study (implying that the class be named `DicomStudy`? See http://www.ssrpm.ch/old/dicom/neumann1.pdf

Maybe not a perfect correspondence, since I think DICOM CT *strictly* lies outside of an RT Study. But the CT is implicit in the Structure Set, I believe. So I think it's still a pretty good match.

Centrus007:
This is looking great! But I need to turn in before I fall asleep on my keyboard. I'll review asap.

Centrus007:
> We should then also have simpler non-dicom plan, structure, dose, and CT objects. Each of these objects can be added to the DicomCollection by their respective functions and a UID and other needed DICOM header tags would be created:
> 
> - 
> - add_dose(dose)
> - add_plan(plan)
> - add_CT(CT)
> - add_structures(structures)

Very cool idea!

src/pymedphys/dicom/_level3/dicom_collection.py
SimonBiggs: Yes, I agree, and maybe we need an extra layer or two. We should possibly have a DicomImage, and maybe instead of DicomCT we should have DicomSeries which would be a series of CT images. It would be worth thinking of whether or not it might be worthwhile adjusting how these pixel values are stored. Let's think on that when we're trying to use it, after it is all made.

We should also have a DicomView class which doesn't actually copy the data across but just acta as a pointer to another DicomObject.

A DicomStudy would be an inter-related  single Series, Plan, Structure, Dose. We need to put some work here to be memory efficient. Allow DicomStudy to contain the objects themselves or simply views to the objects.

A DicomCollection would still exist, but it would allow an arbitrary number of Dicom Studies within it. Any Dicom object with the same UID as another would be verified for equality, an error raised if not the same Dicom dataset, and then a view would be created instead of another full object. That way looking up each Dicom study in the collection would still give all the info, and still look like that CT series is in every DicomStudy in the collection. As well as editing the CT slices on one study in the collection should edit the slices on every other study in that collection also.

Multiple different doses would look like multiple different studies, but the CT series, structures, plan would all be a shared view between them.

... to make this simpler maybe we shouldn't do that defensive copy...

SimonBiggs: If we don't make a defensive copy then we won't need to distinguish between view and otherwise.

src/pymedphys/dicom/_level2/header_tweaks.py
SimonBiggs: > Somewhat relevant: presumably we'll be using dataset (rather than dicom_dataset) in our custom Dicom classes?

yup

src/pymedphys/dicom/_level3/dicom_collection.py
```diff
@@ -0,0 +1,135 @@
...
+    @units.setter
+    def units(self, units_used):
+        self.dataset.DoseUnits = units_used
+
+    # TODO: Need to refactor the following to still be a view to the dataset
+    # but not needlessly call the entire function.
+    @property
+    def x(self):
+        x_value, _, _ = extract_dicom_patient_xyz(self.dataset)
+        return x_value
```
SimonBiggs: Because I am no longer protecting the `dataset` structure with a defensive copy, need to be very sure to only have pointers to that `dataset` object instead of pulling the data out. Otherwise honest changes to the `dataset` would result in an internally inconsistent object.

This way of implementing views is not ideal. I think we should consider inverting the dependency here.

The logic for the each individual coord should belong within the object, and then `extract_dicom_patient_xyz` should depend on this object and just be a helper wrapper function.

What do you think @Centrus007?

Centrus007: This is actually a complicated problem.

Of course, *one* way around it is to instantiate DicomBase with a **filepath** rather than a pydicom dataset (like I had it originally, hehe).

Centrus007: Although, if we encourage the use of `from_file()` in examples, docs, etc. I think this problem is largely circumvented.

Centrus007: Yeah, I do think `self.dataset` ought to be a property. That will help.

Centrus007: Sorry for the barrage, I was typing as I was thinking. I just want to add the following for clarity:

> Otherwise honest changes to the dataset would result in an internally inconsistent object.

I don't think we should permit changes to `self.dataset` at all. I think we should enforce (if possible) or at least strongly hint that a new DicomXXX object should be created if we want to make changes to its underlying pydicom dataset

src/pymedphys/dicom/_level3/dicom_collection.py
SimonBiggs: @Centrus007 I guess what I am thinking of here is the case where we have multiple DicomStudy objects which all have pointers to the same DicomSeries and DicomStructure. I don't want the internal dataset object to be duplicated across every DicomStudy. Instead each dataset within the DicomStudy should just be a view.

Each DicomStudy would be a hierarchical object with the DicomSeries (CT) at its base, DicomStructure on top of that, then DicomPlan, then DicomDose.

I envisage a common use case would be `DicomCollection.from_directory(directory_path)`. If there were multiple plan and dose files for just one CT series I would like a DicomStudy for every Plan/Dose combination. That's a lot of DicomStudy instances. I want each DicomSeries within that DicomStudy to only be a pointer. I don't want to use a whole new chunk of RAM for duplicating the DicomSeries data.

It also would be handy to be able to update some feature of the DicomCollection and have every linked DicomObject with the same UID all update in sync. We get this "update in sync" if internally they are actually all pointing to the same address space of RAM.

src/pymedphys/dicom/_level3/dicom_collection.py
Centrus007: > In this current iteration a change to dataset here is just fine. x, y, z, coords, values, units will all continue to mimic what's in dataset even if it is changed by the user.

Yes I know, your comment seemed to be asking how to *avoid* the current implementation, which does feels ugly/inefficient (having to call extraction functions upon every call to, say, `self.x`).

My thinking was that simply reading DicomXXX attributes will be a more common use-case than *changing* them. If so, my thoughts were to return a new DicomXXX instance for any change required. Or at least to *copy* and replace `self.dataset` and run any dependent attribute-setting methods. Maybe that is the worse of the two trade offs...

> I envisage having a machine ID property for example on DicomPlan, it can be accessed, and changed either via the dataset or via the object property.

I really don't like this optionality. It's safe (for the moment), but unintuitive in my opinion. I do think that the average programmer would look at this:

```python
ds = pydicom.dcmread(ds_filepath)
dose = DicomDose(ds)
```
and assume that `dose` and `ds` are now logically separate entities. Note that this does not mean dose has to be *immutable*, just that any changes are best done via `dose.dataset = x`, and not have `ds = x` alter `dose.dataset`.

Its very debatable, but that's my thinking at the moment.

Centrus007: The above comment notwithstanding, I think I'm happy to merge :)

SimonBiggs: I agree. Hmm. What about a separate trade off, have an optional `no_copy` parameter that defaults to false?

Centrus007: Actually not a bad idea. Incidentally, Numpy itself does this for its arrays. `numpy.array()` has a `copy` parameter that defaults to True. See https://docs.scipy.org/doc/numpy/reference/generated/numpy.array.html

SimonBiggs: Awesome, well let's fall in line with numpy standard and have also have a copy parameter that defaults to True?

Centrus007: Agreed.

Shall we merge this in and implement the copy later? Or squeeze it in now?

SimonBiggs:
Squeeze it in now. Are you able to do that? I'm not able to sit at a
computer at this moment.

SimonBiggs:
Actually I do have a moment, have you begun? I can do it just now

Centrus007:
I haven't sorry. In between chores at the moment...

SimonBiggs:
No need to be sorry at all :), not at all :), I was just umming and ahhing if I should dive in, only to find out you were part way through it :)

SimonBiggs:
Awesome, I'm ready to merge :)

SimonBiggs:
I am tacking on a few changes to `README.rst` to this pull request, sorry about that.

changelog.md
```diff
@@ -8,10 +8,30 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 
 ### Breaking Changes
-- nil
+- `anonymise_dicom` has been renamed to `anonymise_dicom_dataset`
 
 ### New Features
-- nil
+- Implementing a suite of Dicom objects, currently a work in progress:
+  - `DicomBase`, a base DICOM class that wraps `pydicom`'s `Dataset` object.
+    This class includes aditions such as an anonymisation method.
```
Centrus007:
```suggestion
    This class includes additions such as an anonymisation method.
```

changelog.md
```diff
@@ -8,10 +8,30 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+  - `DicomImage`, designed to hold a single DICOM image slice. Might someday
+    contain methods such as `resample` and the like.
+  - `DicomSeries`, a series of `DicomImage` objects creating a CT dataset.
+  - `DicomStructure`, designed to house DICOM structure datasets.
+  - `DicomPlan`, a class that holds RT plan DICOM datasets.
+  - `DicomDose`, a class that to hold RT DICOM dose datasets. It has helper
+    functions and parameters such as coordinate transforms built into it.
+  - `DicomStudy`, a class designed to hold an interelated set of `DicomDose`,
```
Centrus007:
```suggestion
  - `DicomStudy`, a class designed to hold an interrelated set of `DicomDose`,
```

src/pymedphys/dicom/_level3/dicom_collection.py
```diff
@@ -0,0 +1,162 @@
...
+            dataset.is_little_endian = True
+
+        if dataset.is_implicit_VR is None:
+            dataset.is_implicit_VR = True
+
+        self.dataset = dataset
+
+    @classmethod
+    def from_file(cls, filepath):
+        dataset = pydicom.read_file(filepath, force=True)
```
Centrus007: I think `pydicom.dcmread(filepath)` is the standard file reading function. See https://pydicom.github.io/pydicom/stable/ref_guide.html

Is there a reason you've opted for `pydicom.read_file(filepath)`?

src/pymedphys/dicom/_level3/dicom_collection.py
```diff
@@ -0,0 +1,162 @@
...
+        return cls(dataset)
+
+    def __repr__(self):
+        return self.dataset.__repr__()
+
+    def __eq__(self, other):
+        return self.dataset.__eq__(other.dataset)
+
+    def anonymise(self, inplace=False):
+        to_copy = not inplace
```
Centrus007: This looks a little messy. Why not have `def anonymise(self, copy=True):` and avoid `to_copy = not inplace`?

src/pymedphys/dicom/_level3/dicom_collection.py
```diff
@@ -0,0 +1,162 @@
...
+    @property
+    def values(self) -> np.ndarray:
+        return self.dataset.pixel_array * self.dataset.DoseGridScaling
+
+    @property
+    def units(self):
+        return self.dataset.DoseUnits
+
+    @units.setter
+    def units(self, units_used):
```
Centrus007: Not sure I like having this function. A change in units would require a change in values (from those at instantiation) and there's no check of that here...

tests/test_dicom/test_dicom_base.py
```diff
@@ -0,0 +1,84 @@
+# Copyright (C) 2019 Simon Biggs
```
Centrus007: Thank you for adding these!

src/pymedphys/dicom/_level3/dicom_collection.py
SimonBiggs: Nope, no good reason. I'll swap over.

src/pymedphys/dicom/_level3/dicom_collection.py
SimonBiggs: Yup, let's remove it.

tests/test_dicom/test_dicom_base.py
SimonBiggs: They were well overdue :)

tests/test_dicom/test_dicom_base.py
```diff
@@ -0,0 +1,86 @@
...
+        'PatientName': 'Python^Monte'
+    })
+
+    dicom.to_file(temp_file)
+
+    new_dicom = DicomBase.from_file(temp_file)
+
+    # TODO: Without the str this was passing locally but not on CI. Further
+    # investigation needed.
+    assert str(new_dicom) == str(dicom)
```
SimonBiggs: @Centrus007 Not sure why this was ...

src/pymedphys/dicom/_level3/dicom_collection.py
Centrus007: Ah, I see. Fair enough :)

SimonBiggs: :)

SimonBiggs:
I'm ready for a merge I think. This CI system is bugging me, I want to change gear and see if I can solve https://github.com/pymedphys/pymedphys/issues/102 before bed...

tests/test_dicom/test_dicom_base.py
Centrus007: Curious... do you think we need to define our own `__eq__()`?

Centrus007: > do you think we need to define our own __eq__() or __cmp__()

Actually, I think we ought to do this for any custom class, due to the way python treats object equality (i.e. it includes object ids in the comparison. See [this SO post](https://stackoverflow.com/questions/390250/elegant-ways-to-support-equivalence-equality-in-python-classes). Also see the [Python docs](https://docs.python.org/2/library/stdtypes.html#comparisons).

Centrus007: Ah right. I should've looked before I leapt.

SimonBiggs: Haha, well it certainly was Easier To Ask Forgiveness, you pythonista :)

Centrus007: XD

Issue #56, opened 2019-02-17
Rename dcm module to dicom


SimonBiggs:
With the new file structure we don't need to have the module name in front of every file path. This also means that short module names aren't really needed anymore.

I believe `dcm` as a name is too cute. I think it would make far more sense to call the module `dicom`.

Centrus007:
Agreed.

Centrus007:
On a related note, we've typically been using the variable name "dcm" to represent a pydicom Dataset. This is in contrast with pydicom convention, where "ds" (short for "dataset, I believe) is preferred.

Should we replace all `dcm` variables with `ds` in the library?

SimonBiggs:
What about in the pydicom docs, does it use dcm there? I'm sure I picked it
up from somewhere.

I'd be happy to change, but I'd like to confirm that we are actually
breaking convention before we do.

SimonBiggs:
Also :) if there's no rush on your end I'd be happy to do it when I do get
a chance

Centrus007:
The link I provided above is from the PyDicom docs. But for more specific evidence, see:

- From the Pydicom Complete API Reference, **Examples** under the `class pydicom.dataset.Dataset` entry ([link here](https://pydicom.github.io/pydicom/stable/api_ref.html#pydicom.dataset.Dataset)):
```python
>>> ds = Dataset()
>>> ds.PatientName = "CITIZEN^Joan"
>>> ds.add_new(0x00100020, 'LO', '12345')
>>> ds[0x0010, 0x0030] = DataElement(0x00100030, 'DA', '20010101')
```
- From the Pydicom Complete API Reference, **Examples** under the `pydicom.filereader.dcmread()` function entry ([link here](https://pydicom.github.io/pydicom/stable/api_ref.html#pydicom.filereader.dcmread)):
```python
>>> ds = pydicom.dcmread("rtplan.dcm")
>>> ds.PatientName
```

Centrus007:
In fact, even PyDicom's dataset editing functions name the returned dataset `ds`. For example, see:

* From the Pydicom Complete API Reference, the docstring for the`pydicom.filewriter.correct_ambiguous_vr()` function ([link here](https://pydicom.github.io/pydicom/stable/api_ref.html#pydicom.filewriter.correct_ambiguous_vr)):

> Iterate through `ds` correcting ambiguous VR elements (if possible).
[...]
Returns `ds:pydicom.dataset.Dataset

Centrus007:
[...]

PR #64, opened 2019-02-18
[WIP]: (Attempt to) convert all pydicom Dataset variables named `dcm` to `ds` and convert all use of "dcm" in module names, directories, function names etc. to "dicom"

Centrus007:


Centrus007:
I've also had a go at the (much bigger) task of addressing #56. Passed tests locally and docs looked all in order.

So far I've mostly omitted notebooks, so many will now break. I'll try to tackle them tomorrow. There are also many .json files that will break, but I wasn't really sure where they came from / how important they were (were they from your GUI experimentation a couple of months ago?)

Centrus007:
Since I made a rather large set of changes since your approval, I'll leave this unmerged until you get a chance to see it :)

Centrus007:
> I am partially of the opinion to not worry too much about updating the notebooks. But I am torn, what do you think?

I think it'll be relatively easy to adjust them as we use them...

> One request before merge, we need to mark this under breaking changes within the change log that is in the root directory. This will be a breaking changes that is involved with the upgrade from 0.5.* --> 0.6.0.

> For a different branch we also need to make it so that that change log appears on the documentation website before the 0.6.0 release. Given users would need to update their code.

Good point. I'd forgotten about the Changelog and am not too familiar with adjusting its visibility in the docs. I'll have a crack.

SimonBiggs:
With regards to putting the change log within the docs we have two options.
One would be to include the following sphinx module as a dependency with
building our sphinx docs which allows us to also pull in markdown files:

https://github.com/rtfd/recommonmark

Another option would be to convert the change log to reST and then directly
import it.

Of the two options I am leaning towards the former as I am sure there will
be multiple cases where we will want to use markdown within the docs for
some reason.

SimonBiggs:
As a note, when I was first setting up the docs I initially had the readme
as a markdown file and was importing it into the docs. Here is a commit
where that was the case:

https://github.com/pymedphys/pymedphys/tree/69a2a20f86c7c2c4ccdcbc04f49f16c78e137a58

Centrus007:
Shall I merge these in or sort out changelog visibility first?

changelog.md
```diff
@@ -27,7 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - nil
 
 
-## [0.5.1] - 2018-01-05
+## [0.5.1] - 2019-01-05
```
Centrus007: I think this should read 2019. @SimonBiggs, can you confirm?

SimonBiggs:
> Shall I merge these in or sort out changelog visibility first?

Merge this first. Changlog visibility can be a separate branch.

changelog.md
SimonBiggs: Yup 2019. See https://pypi.org/project/pymedphys/#history

SimonBiggs:
As an aside I think we should draw inspiration from how other projects do their changelogs:

https://github.com/angular/angular/blob/master/CHANGELOG.md#breaking-changes

PR #69, opened 2019-02-19
Change changelog headings.

Centrus007:
New headings, largely taken from `angular`, in (my preferred) order:

- Breaking Changes
- New Features
- Bug Fixes
- Code Refactoring
- Performance Improvements 

I'm thinking that we keep all headings there with 'nil' underneath for the **Unreleased** subheading, but delete headings with "nil" once a release is cut. What do you think?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,11 +3,8 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-
+<br/>
 ## [Unreleased]
-
-### Added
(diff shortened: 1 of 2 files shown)


SimonBiggs:
Yup good plan

Centrus007:
It's not really the place for it, but since it was a small change, I also made the following changes to `anonymise_dicom()`:

- Change the `AssertionError` to `ValueError` as per your suggestion.
- Add a couple of tags I've noticed recently that probably should also be anonymised.
- Sort the tags to be anonymised for ease of reading.

PR #81, opened 2019-03-02
[RFR] gamma optimization

Centrus007:
A few minor tweaks that I *believe* made small improvements to gamma calculation. Pretty small fish though.


src/pymedphys/gamma/_level2/gammashell.py
```diff
@@ -399,14 +399,14 @@ def calculate_min_dose_difference(options, distance, to_be_checked,
...
     num_points_in_shell = np.shape(coordinates_at_distance_shell)[1]
 
-    estimated_ram_needed = (np.uint64(num_points_in_shell)
-                            * np.sum(to_be_checked).astype(np.uint64)
-                            * np.uint64(32)
-                            * np.uint64(num_dimensions)
+    estimated_ram_needed = (num_points_in_shell
+                            * np.sum(to_be_checked)
+                            * 32
+                            * num_dimensions
```
SimonBiggs: These `np.uint64` were written in to fix the following bug:

https://gitlab.com/pymedphys/pymedphys/issues/17

I couldn't ever replicate the issues on my machines so I'm not sure how to verify that removing them doesn't reintroduce the bug.

I have a few ideas here though.

Centrus007: I know.

There's no need to explicitly cast all of the multiplied together variables to uint64 though. One cast is sufficient.

Centrus007: Maybe we'll just remove these changes from the merge. I don't think we lose anything remotely noticeable from explicitly casting each of them individually.

I'd mistakenly thought one of the variables was a largish array, but they're all either scalars are single-valued arrays, right?

SimonBiggs:
I suspect the slow part might be the bollean sum that's going on there. And
o can replace that by taking a shape of an object a few lines later. That
should completely fix up that particular slow down. :).

src/pymedphys/gamma/_level2/gammashell.py
Centrus007: Maybe we'll just remove these changes from the merge. I don't think we
 lose anything remotely noticeable from explicitly casting each of them
 individually.

 I'd mistakenly thought one of the variables was a largish array, but
 they're all either scalars are single-valued arrays, right?

Centrus007:
> I suspect the slow part might be the bollean sum that's going on there. And
o can replace that by taking a shape of an object a few lines later. That
should completely fix up that particular slow down. :).

Possibly. I get an 8% improvement in performance by changing the above lines. Not shooting the lights out, but not as small as I'd thought, actually...

You mentioned testing... why do you need to replicate the actual RAM condition? Is it not sufficient to just simulate the values of `to_be_checked` and `num_points_in_shell` to verify no issue with the casting (as I basically did above)?

SimonBiggs:
Wow 8%. Yeah you're right about the testing. If you set all of them to 32
bit does the overflow actually happen or does your numpy install expand to
64 bit?

SimonBiggs:
I'd be happy to merge in, there's going to be heaps of little iterations to
speed this up. Let's grab the 8% and work on it further .

SimonBiggs:
Once you're happy merge this one in :)

Centrus007:
Scrap that, I was being fooled by randomness. I now seem to be getting consistenly slower results with a change in the value of `random_subset`.

I'll adandon those lines.

But in case you're still interested, adding two uint32s does result in an overflow for me:

```python
>>> import numpy as np
>>> a = np.uint32(1e7)
>>> b = np.uint32(1e7)
>>> c=a*b
__main__:1: RuntimeWarning: overflow encountered in ulong_scalars
>>> c
276447232
>>> c.dtype
dtype('uint32')
>>> d = np.uint64(1e7)
>>> e = a*d
>>> e
100000000000000
>>> e.dtype
dtype('uint64')
```

SimonBiggs:
Thanks Matt :)

SimonBiggs:
Use the line profiler tool. It shows you where all that speed is
lost/gained and therefore helps confirm whether or not speed up was due to
random or otherwise.

By the way, all of this work, I imagine we might manage eventually getting
a factor of two improvement... But I doubt we'll be seeing any order of
magnitude jumps.

Centrus007:
> Use the line profiler tool. It shows you where all that speed is
lost/gained and therefore helps confirm whether or not speed up was due to
random or otherwise.

I have been. Mind you, I'm pretty hungover today and basically started looking at this as a way to "relax" and pass the time. In hindsight I don't think my logical faculties are up to scratch for this. I've misread the tool more than once by forgetting that I'm looking at percentages of percentages.

> By the way, all of this work, I imagine we might manage eventually getting
a factor of two improvement... But I doubt we'll be seeing any order of
magnitude jumps.

Yeah, you're probably right. I'll probably leave it (at least for now).

SimonBiggs:
Haha awesome :)

Sounds like it was a great night?

SimonBiggs:
And it is a great way to relax I think :)

I'm having a read of the following by the way:

https://arxiv.org/abs/1809.04430

I know we have a lot of stuff on our plate, but implementing that paper is
looking awfully shiny.

notebooks/gamma/MJ-gamma-testing.ipynb
```diff
@@ -12,8 +12,12 @@
     "import psutil\n",
```
SimonBiggs: <p>Testing a comment using https://app.reviewnb.com/pymedphys/pymedphys/pull/81/files/</p>

 _Reply via <a href='https://app.reviewnb.com/pymedphys/pymedphys/pull/81/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

PR #83, opened 2019-03-02
Add Dummy replacement values to DICOM anonymizer

Centrus007:


src/pymedphys/dicom/_level2/dicom_anonymise.py
```diff
@@ -31,73 +31,85 @@
...
+                                       'CS': "ANON",
+                                       'DA': "20190303",
+                                       'DT': "20190303000900.000000",
+                                       'LO': "Anonymous",
+                                       'PN': "Anonymous",
+                                       'SH': "Anonymous",
+                                       'SQ': ["Anonymous"],
+                                       'ST': "Anonymous",
+                                       'TM': "000900.000000",
+                                       'UI': "12345678"}
```
SimonBiggs: Very neat approach :). Hangover can't have been that debilitating :)

Centrus007: Hahaha, thanks :)

PR #94, opened 2019-03-12
[RFR] Pull profile from arbitrary coords

SimonBiggs:
@JacobMcAloney will provide screenshots of the tests undergone of this new Excel function.


JacobMcAloney:
Test1

Formula bar shows original script used to generate an in-plane profile from a .dcm file from a wedged beam, profile is plot of generated data


Test2

Formula bar shows the new script used to generate an arbitrary profile with an inplane-crossplane-depth coordinate system, profiles are plots of generated data


Test3

The two scripts generated identical profiles

SimonBiggs:
Thanks @JacobMcAloney.

@Centrus007, the key here is that a new function was made that allows for the interpolation of arbitrary profiles from a DICOM file into Excel.

This code suffers from the same coordinate system flaws as the rest of the DICOM dose extractors, but this was quite helpful in extracting diagonal profiles.

@JacobMcAloney could you also provide a screenshot of comparing Monaco's CC algorithm to measurement in Excel of a diagonal profile?

src/pymedphys/dicom/_level2/dicom_dose.py
```diff
@@ -451,6 +455,22 @@ def load_dicom_data(ds, depth_adjust):
     return inplane, crossplane, depth, dose
 
 
+def interpolating_profile_extract(ds, depth_adjust, inplane_ref, crossplane_ref, depth_ref):
```
Centrus007: Slightly odd function name. Ordinarily, the verb should be the first word in the function name. E.g. `extract_interpolated_profile`.
Or is "extract" a noun in this context? Even then, `interpolate` would be preferred over `interpolating`.

src/pymedphys/xlwings/_level1/dicom.py
```diff
@@ -81,3 +82,18 @@ def crossplane_profile(dicom_path, depth_adjust, depth_lookup, averaging_distanc
...
         return np.vstack([crossplane, crossplane_dose]).T
+
+    @xw.func
+    @xw.arg('dicom_path')
+    @xw.arg('depth_adjust')
+    @xw.arg('inplane')
+    @xw.arg('crossplane')
+    @xw.arg('depth')
+    @xw.ret(expand='table')
+    def arbitrary_profile(dicom_path, depth_adjust, inplane, crossplane, depth):
```
Centrus007: Try to prepend the function name with a verb. =)

SimonBiggs: What do you think about the following:

https://stackoverflow.com/a/1991341/3912576

Given there is no program state here being changed just values being returned given inputs is a verb required? I'm not sure?

src/pymedphys/dicom/_level2/dicom_dose.py
SimonBiggs: I changed this to `arbitrary_profile_extract_from_ds`, what do you think? I'm not sold...

SimonBiggs: Thought about it a moment longer, `extract_arbitrary_profile_from_ds` is better.

src/pymedphys/xlwings/_level1/dicom.py
SimonBiggs: Further thought on this point, this function name ends up appearing within Excel. `arbitrary_profile` seems to be closer to the types of names found within Excel than say something like `extract_arbitrary_profile`.

JacobMcAloney:
These changes made (https://github.com/pymedphys/pymedphys/pull/94/commits/5a6d6642f6c436d2fbbe3264cfdafe72942bcf45) allow blank entries in the input resulting in blank output rather than an error.

First picture below shows the function being used in Excel, the second picture shows that blank cells in the input columns resulted in blank outputs of the arrays.

20190314 Profile Change1
20190314 Profile Change2

src/pymedphys/xlwings/_level1/dicom.py
Centrus007: Happy to keep this as `arbitrary profile`. Prepending a verb is not a strong preference of mine, and the argument in the link you provided seems fair. It was more a feeling that most of our other code makes heavy use of words like `extract` (which is just a fancy `get`). Perhaps the answer is to adjust our other code, rather than add `extract` here.

src/pymedphys/dicom/_level2/dicom_dose.py
Centrus007: Or `arbitrary profile from dicom dataset`? Or even `arbitrary profile from dicom dose`? No need to stick to `ds` for function names.

SimonBiggs: Yes, `arbitrary_profile_from_dicom_dose`. I've been getting to the point that I have just wanted to implement our full DICOM object stack that we were referring to. It would make it possible to create tests for these kinds of functions in quite a nice neat way...

Centrus007: Hahaha, I hear you :)

src/pymedphys/xlwings/_level1/dicom.py
SimonBiggs: That makes sense

src/pymedphys/dicom/_level2/dicom_dose.py
SimonBiggs: :)

SimonBiggs:
@Centrus007 if you merge this one then the commit also attached you onto it, baking into the git history recognition of your work on reviewing this one.

Centrus007:
Okay, thanks for the heads up (and the consideration)!

PR #95, opened 2019-03-13
Profile no xarray redo -> merge to master

kingrpaul:
Also, would it be straightforward to create sphinx documentation (HTML or PDF) for something  "in the labs" like this, outside the documentation that is visible from the website?

Thanks in advance, @SimonBiggs, and hope y'all are doing well.


SimonBiggs:
Yup that would be a great idea

kingrpaul:
I would create a new folder in the usual docs location?

SimonBiggs:
Yup, make a whole extra subheading. Make it down the bottom. Make the
heading something like Labs (untested, unreviewed).

Then under Labs should be your name, then under your name you can place
what you would like. We'll see how that goes. See if that works.

PR #96, opened 2019-03-14
[RFR] Variable naming improvements within DICOM header tweaks

SimonBiggs:
Thank @Centrus007 for your feedback :)

I have made those changes here.


src/pymedphys/dicom/_level2/header_tweaks.py
```diff
@@ -128,11 +128,16 @@ def adjust_rel_elec_density_cli(args):
     pydicom.write_file(args.output_file, new_ds)
 
 
-def RED_adjustment_map_from_structure_names(structure_names):
-
+def extract_RED_from_structure_name_regex():
```
Centrus007: I didn't mean to suggest the regex could benefit from refactoring out as a *function*. I simply meant something like the following might help with understanding:

```python
def RED_adjustment_map_from_structure_names(structure_names):
    structure_name_regex = r'^.*RED\s*[=:]\s*(\d+\.?\d*)\s*$'
    structure_name_pattern = re.compile(structure_name_regex)
    adjustment_map = {
        structure: float(structure_name_pattern.match(structure).group(1))
        for structure in structure_names
        if structure_name_pattern.match(structure)
    }
    return adjustment_map

But I'm happy with either approach :)

Centrus007: I didn't mean to suggest the regex could benefit from refactoring out as a *function*. I simply meant something like the following might help with understanding:

```python
def RED_adjustment_map_from_structure_names(structure_names):
    structure_name_containing_RED_regex = r'^.*RED\s*[=:]\s*(\d+\.?\d*)\s*$'
    structure_name_containing_RED_pattern = re.compile(structure_name_containing_RED_regex )
    adjustment_map = {
        structure: float(structure_name_containing_RED_pattern.match(structure).group(1))
        for structure in structure_names
        if structure_name_containing_RED_pattern.match(structure)
    }
    return adjustment_map

But I'm happy with either approach :)

Centrus007:
Line 72 of https://github.com/pymedphys/pymedphys/blob/659b1728e91d69f370b58167e03c90361b8d4b10/src/pymedphys/dicom/_level2/header_tweaks.py has a typo ("stucture" instead of "structure")

src/pymedphys/dicom/_level2/header_tweaks.py
SimonBiggs: That's better :). Made that change.

Centrus007:
So (one of) your solution(s) is to append RED=<value> to your structure names? It's really quite curious that you have to do this and that Mosaiq doesn't export the overridden REDs. Or is it actually your TPS failing to do so? I don't have a full handle on the problem...

SimonBiggs:
> So (one of) your solution(s) is to append RED= to your structure names? It's really quite curious that you have to do this and that Mosaiq doesn't export the overridden REDs. Or is it actually your TPS failing to do so? I don't have a full handle on the problem...

Yeah, Monaco doesn't export REDs within the DICOM file :man_facepalming:

SimonBiggs:
> Line 72 of https://github.com/pymedphys/pymedphys/blob/659b1728e91d69f370b58167e03c90361b8d4b10/src/pymedphys/dicom/_level2/header_tweaks.py has a typo ("stucture" instead of "structure")

Fixed :) thanks :)

PR #98, opened 2019-03-15
[WIP] Large file storage using Zenodo

SimonBiggs:


SimonBiggs:
Considering not using Zenodo in favour of https://github.com/pymedphys/pymedphys/issues/155

That way we can keep using git, and we can upload large files, as much as we want, and to clone we just need to have VFS for git installed...

SimonBiggs:
I'm closing this for now, in favour of watching and waiting the progress of VFS for git.

PR #103, opened 2019-03-16
[RFR] Improve CI testing

SimonBiggs:


SimonBiggs:
@Centrus007 so this has now undergone some serious speed improvements. At the moment I have only got things working on Linux, but they are working for our full test suite, and working much faster.

It's not perfect, but given the speed increase for this current iteration there may be benefit merging this in as is, and then have me work on getting Windows and macOS up and running again in a separate pull request.

SimonBiggs:
Just so you know, to run the full test suite locally once this is merged in tox will need to be installed. 

To run the full test suite simply call `tox`.

The test suite can also be run in parallel bringing the full suite time down to about 25 seconds on my machine. To run in parallel do the following:

```bash
tox -p all
```

Cheers,
Simon

SimonBiggs:
Actually, not ready for review. Thinking further, I want to get this method working for Windows and macOS. I want to make sure that it's possible before we swap to this method.

Centrus007:
> @Centrus007 so this has now undergone some serious speed improvements. At the moment I have only got things working on Linux, but they are working for our full test suite, and working much faster.

Awesome!

> Thinking further, I want to get this method working for Windows and macOS. I want to make sure that it's possible before we swap to this method.

I think that's prudent :)

SimonBiggs:
One thing that would be helpful if you could run this branch and run `tox`
on your machine and let me know if it works for you

SimonBiggs:
I need to also include a pytest-doctest pipeline.

setup.py
```diff
@@ -91,11 +73,11 @@ def run_tests(self):
         'keyring',
         'matplotlib',
         'notebook',
-        'numpy>=1.12',
+        'numpy <1.16, >=1.12',
```
Centrus007: Why < 1.16?

SimonBiggs: Needed until the following bug is resolved:

https://github.com/PyCQA/pylint/issues/2721

SimonBiggs: The alternative, which I am certainly open to, is disabling those error messages from pylint. I figure we can resort to that if the bug takes a long time to resolve.

Centrus007:
I'm getting massive ransomware warnings when I try t use `tox`. A simple Google search for `tox` only adds to my fears...

What am I missing?

SimonBiggs:
Is this the tox you're using:

https://tox.readthedocs.io/en/latest/

?

SimonBiggs:
Also, was this from running pip install tox?

Centrus007:
> Also, was this from running pip install tox?

Yes. Pretty confident there was no typo.

> https://tox.readthedocs.io/en/latest/

If that's what comes when I run `pip install tox`, then yes.

I'll try again tonight.

Centrus007:
Upon reflection, I did run it from within my pymedphys conda env. Now I'm thinking that I shouldn't have...

SimonBiggs:
Yeah, I think have it as a top level pip install. It goes and makes its own
virtual envs during testing.

SimonBiggs:
@Centrus007 so in the end, I just simply couldn't get the tests to work on Windows or mac without using Anaconda. The dependencies are just too hard to install without it.

So, work around, included Anaconda (and the induced slowness), just for mac and windows. But for most tests, if the tests fail, they should fail quite fast on the pip only Linux tests.

SimonBiggs:
Yup, so this is now ready for review @Centrus007 :)

SimonBiggs:
Actually I'm going to need to merge this Matt, we can branch master back to this point if we need to remove it, but I believe it is all okay.

Centrus007:
No worries :)

SimonBiggs:
Haha, after merging this my heart absolutely sank when I saw the fail cross on master :/ Anyway, it was due to the following:

https://github.com/kynan/nbstripout/issues/94

So fixed that now in https://github.com/pymedphys/pymedphys/pull/110

PR #105, opened 2019-03-21
[RFR] included electron factor visualisation plots

SimonBiggs:
Just some plots for the electron factor code


SimonBiggs:
Thanks Matt :)

PR #107, opened 2019-03-25
[RFR] Wildcard for xlwings file lookup + Mephysto importing

SimonBiggs:


SimonBiggs:
I'd really like to think of a way to test these `xlwings` scripts... They need Excel though :/ *sigh*.

SimonBiggs:
@Centrus007 I recognise this is a huge pull request, I don't really have a chance to polish this up. Let me know what you think...

SimonBiggs:
Essentially its purpose was to pull in some old code (to read PTW Mephysto files) and allow Excel to read directly from Mephysto...

SimonBiggs:
Fixes https://github.com/pymedphys/pymedphys/issues/108

src/pymedphys/mephysto/_level1/mccread.py
```diff
@@ -0,0 +1,171 @@
+# Copyright (C) 2017 Cancer Care Associates
```
Centrus007: What does "mc" refer to in "mcread.py"

Centrus007:
**I'd really like to think of a way to test these xlwings scripts... They need Excel though :/ sigh**

Do they? You can't just write to and read from an .xlsx file (using either `xlwings` or something else like `openpyxl`) and check for expected values/structure?

Centrus007:
> I'd really like to think of a way to test these xlwings scripts... They need Excel though :/ sigh.

Why not write the tests anyway and run them locally? Provide evidence in the form of a screenshot for merging.

I don't think we should abandon good testing just because it can't easily done through an Azure pipeline.

src/pymedphys/mephysto/_level1/mccread.py
SimonBiggs: The file extension of PTW Mephysto files `.mcc`

SimonBiggs:
Sorry about this one Matt, I think I'll need to prioritise some other tasks first (such as documentation of the log files code), but will get back to this.

Potentially a neat way to sort out what is needed for these functions is to write docs with screenshots for each function. Doubles as helping the review process as well as getting documentation out...

SimonBiggs:
> Why not write the tests anyway and run them locally? 

What would a written test look like?

I mean we manually user test these things of course, but how does one write an Excel test?

Centrus007:
Shall we merge this in or relocate to a lab?

SimonBiggs:
I'll ask @JacobMcAloney to provide some screenshots of his tests demonstrating usage. Let's see how we go after that.

Also @JacobMcAloney, if the screenshots you pick could be chosen so that in the future they could double as documentation that'd be great.

Cheers,
Simon

JacobMcAloney:
Reference Mephysto file via filepath with wild card (*):

Mephysto_Pull_1


Use _mephysto()_ function to pull data from Mephysto file. This generates multiple columns of data (shown here in yellow), as well as column headings detailing type of profile, and depth of profile for in-plane/cross-plane profiles.

Mephysto_Pull_2


Pulled data plotted in excel (left), visually compared to original Mephysto data (right)

Mephysto_Pull_3


Directly compared data pulled via the _mephysto()_ function and data manually copied from Mephysto:

Mephysto_Pull_4

SimonBiggs:
Thanks @JacobMcAloney. @Centrus007, shall we see if this can go through another review round?

SimonBiggs:
Thanks @Centrus007. Appreciate the review. Sorry about the lack of material on the first pass.

PR #110, opened 2019-03-27
[RFR] pin nbstripout

SimonBiggs:
See https://github.com/kynan/nbstripout/issues/94


SimonBiggs:
Force pulling this nbstripout pin so that master stops failing :) (a package upgraded last night and broke our CI)

PR #114, opened 2019-03-28
[WIP] Begin work on new docs

SimonBiggs:


SimonBiggs:
This is only failing because it doesn't have decode trf included within it yet. Plan to pull trf decode in next week, so that should fix the tests online by the time I need to merge this.

SimonBiggs:
@Centrus007, these docs are starting to look quite nice :). I've put a bit of work in smoothing out the CLI documentation. I've also made it so that we can create example documentation with notebooks, and the notebooks can also be run online.

Currently it can't build without the logfile decode library. But I'd be keen if you'd be able to build it locally and give feedback.

Still a few pages need to be filled out. Will likely do that on Monday. I also need to actually describe the process of setting up the logfile orchestration in full.

SimonBiggs:
Just changed the branch name https://github.com/pymedphys/pymedphys/pull/115

PR #118, opened 2019-04-03
[RFR] bump nbstripout version

SimonBiggs:
nbstripout regression fixed. See:

https://github.com/kynan/nbstripout/issues/94#issuecomment-479241348

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -166,7 +166,7 @@ jobs:
     - bash: "python -m pip install -U pip setuptools --user"
       displayName: Upgrade pip and setuptools
 
-    - bash: "python -m pip install -U "nbstripout>=0.3.5" --user"
+    - bash: python -m pip install -U "nbstripout>=0.3.5" --user
       displayName: Install nbstripout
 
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -166,7 +166,7 @@ jobs:
     - bash: "python -m pip install -U pip setuptools --user"
       displayName: Upgrade pip and setuptools
 
-    - bash: "python -m pip install -U nbstripout==0.3.3 --user"
+    - bash: "python -m pip install -U "nbstripout>=0.3.5" --user"
```
Centrus007: Is this going to work with the added quotaton marks? Shouldn't this be:

```suggestion
    - bash: "python -m pip install -U nbstripout>=0.3.5 --user"
```
?

SimonBiggs: Might a slight change

SimonBiggs: But yes, double quotes there on inside and outside was an issue

Centrus007: > Might a slight change

Sorry?  :p

SimonBiggs: Yup, you were right the double quotes were an issue. I did still need them around the `"nbstripout>=0.3.5"` because `>` is a special bash symbol. But funnily enough I didn't need them around the full item. So I adjusted your suggestion, and then committed :)

PR #123, opened 2019-04-04
[RFR] Initial WL notebooks

SimonBiggs:


SimonBiggs:
@JacobMcAloney this is the branch with my WL notebook in it. The notebook itself can be found at:

https://github.com/pymedphys/pymedphys/blob/2bc34e976e20b43826427ed069474dd945d62a03/notebooks/mviso/07-mviso.ipynb

To get there take the following steps, within your local PyMedPhys directory run:

```bat
git pull
git checkout scratch-mviso
git pull
pip install -e .

jupyter lab
```

Then a browser Window will open up. In the JupyterLab filebrowser navigate to `notebooks/mviso/07-mviso.ipynb`. Change the cell which defines `data_root` to be the new file location, then within the menu press `Run > Run All Cells`.

Cheers,
Simon

SimonBiggs:
Heya @Centrus007,

I know every man and his dog has implemented a winston lutz algorithm. I was reading through the PyLinac implementation and unfortunately from what I can read it centres the fields based on a single 50% pair of points on a single 1D profile through the axis. I believe this method would be susceptible to noise.

I tried to verify this, but struggled to strip the logic down into something I could manage.

In the end I decided to re-implement it. This method finds the field by inverting the image around a suspected centre then taking a sum of square differences across the penumbra width. It then passes this sum of squares to a scipy optimiser.

This overlapping penumbra method appears to be quite robust, for example see the following offset field:

image

So given this implementation, I intend to work this logic into a new module within PyMedPhys and create a corresponding CLI.

Centrus007:
I'm not at all against implementing our own, especially if we can demonstrate improved accuracy and info reporting. WL actually gives a LOT of useful info that can be used to diagnose multiple different issues.

My first thought is that there are surely some very robust algorithms (indeed, functions within existing python modules) that would do a far better job than anything manual we come up with. It's essentially rudimentary image recognition - a LOT of work has been done in this arena.

IIRC, our own in-house Winston-Lutz makes use of a canny filter (at least for the BB). But it's not my baby, I'd have to dive in to look.

Centrus007:
[OpenCV ](https://opencv-python-tutroals.readthedocs.io/en/latest/index.html) is probably own friend here, especially the image processing & feature detection modules...

Centrus007:
FWIW, your example images have excellent contrast and are not very noisy at all!

SimonBiggs:
I just used scipy basinhopping module. It worked very well.

Centrus007:
It's a little difficult for me to follow in the notebooks, even within reviewnb, but it seems to me that this PR has a crapton of lines of code. I imagine fewer than, say, 300 lines making use of existing modules to find both the field centre and ball centre would suffice? Maybe even perform better...

Centrus007:
Sorry, each notebook is demonstrating slightly different implementations?

Centrus007:
Let me review properly, sorry. I'll pull it in

SimonBiggs:
No need to really review the notebooks. I want to tidy this up into python
modules before it's ready for prime time.

SimonBiggs:
I have a bit of a habit, arguably not the best habit, of mapping out my
thought process using version numbers on the notebook. Only 07 is the one
that is valid, all the others are experimenting. 07 works well, but it
certainly needs refactoring. I'll refactor and undergo test driven
development as I transform it into python modules.

Centrus007:
No worries.

I can supply you with some of our own Winston Lutz images, if you like? They might help with ensuring the generlisability of the code?

SimonBiggs:
That would be great actually. Can you supply them in DICOM format?

Centrus007:
Here you go!

[RI.Physics 4.G0C270-12_1_8.zip](https://github.com/pymedphys/pymedphys/files/3061426/RI.Physics.4.G0C270-12_1_8.zip)

SimonBiggs:
Let's do this in a few stages. Let's merge these notebooks in, I'll sort something out regarding proper implementation at a later date.

PR #125, opened 2019-04-04
[RFR] post release work

SimonBiggs:


Centrus007:
I've been wondering about the Changelog a bit. Should Unreleased be at the top? Or is it better to list the latest stable version's changes first?  E.g. see https://docs.scipy.org/doc/numpy/release.html

Centrus007:
One idea: Unreleased could go in a separate file / page in the docs, and get copied across to release notes upon release.

This is not a strong opinion of mine, just a thought.

SimonBiggs:
> I've been wondering about the Changelog a bit. Should Unreleased be at the top? Or is it better to list the latest stable version's changes first? E.g. see docs.scipy.org/doc/numpy/release.html

The way I've done this, this time round. Unreleased won't show up on the stable docs:

http://pymedphys.com/en/stable/getting-started/changelog.html

But on the latest docs they then should appear at the top... I think that makes sense. If you're looking at the "latest" docs, you would likely want to see what has changed since the last stable...

Centrus007:
That does make sense. Nvm

SimonBiggs:
> One idea: Unreleased could go in a separate file / page in the docs, and
get copied across to release notes upon release.

> This is not a strong opinion of mine, just a thought.


I do like this idea

PR #127, opened 2019-04-05
[RFR] 0.7.2 merge into master

SimonBiggs:


setup.py
```diff
@@ -70,10 +70,11 @@ def read(*names, **kwargs):
     license='AGPLv3+',
     install_requires=[
         'attrs',
+        'dataclasses; python_version=="3.6"',
         'keyring',
         'matplotlib',
         'notebook',
-        'numpy <1.16, >=1.12',
+        'numpy < 1.16, >= 1.12',
```
SimonBiggs: @Centrus007 I was hoping to have a different work around for this upper limit. But it was taking too long to solve. So we'll have this upper limit for a bit longer.

SimonBiggs:
This is a `master` + `0.7.x` branch ready for merging back into master

PR #132, opened 2019-04-07
[RFR]: Dicom anonymise file + folder

Centrus007:


notebooks/dcm/anonymise-dicoms-glob-method.ipynb
```diff
@@ -0,0 +1,95 @@
...
+   "source": [
+    "import os\n",
+    "from glob import glob\n",
+    "import pathlib\n",
+    "\n",
+    "import pydicom\n",
+    "\n",
+    "from pymedphys.dicom import anonymise_dicom_dataset"
+   ]
+  },
```
Centrus007: merged in Jacob's changes to pre-empt any conflicts. But we'll merge his changes in first so they're appropriately attributed to him :)

src/pymedphys/dicom/_level2/anonymise.py
```diff
@@ -213,6 +239,36 @@ def anonymise_dicom_dataset(ds, replace_values=True, delete_private_tags=True,
...
+                         delete_private_tags=True, keywords_to_keep=None,
+                         ignore_unknown_tags=False, delete_unknown_tags=False):
+
+    ds = pydicom.dcmread(dicom_filepath)
+
+    if overwrite_file:
+        dicom_anon_filepath = dicom_filepath
+    else:
+        dicom_anon_filepath = pjoin(dirname(dicom_filepath),
+                                    'Anonymised_' + basename(dicom_filepath))
```
Centrus007: Keen for thoughts on this. Currently, if the user doesn't want to replace the file, this function just prepends 'Anonymized' to the existing filename.

One problem with this approach is that the filename itself can have identifying information. Although many DICOM filenames tend to just consist of the SOPInstanceUID preprended with something like 'CT' or 'RP' it's also not so uncommon for DICOM filenames to consist of patient info. E.g. ARIA exports DICOM files containing PatientID by default.

Should we try to handle this, or not? E.g., we could search the filename for instances of, say, PatientID and Patient Name - heck, maybe even search for all of the anonymized DICOM tag values in the filename - and either wipe them or replace them with dummy values. The trade-off as I see it is:

- If we do this, the user can more completely rely on the function's anonymisation, BUT

- It introduces a level of inconvenience to the user, where it'll be harder to see which of the original files a new anonymized file corresponds to. This would especially be annoying for anonymisation within a folder.

Is it enough to simply warn the user that the filename is untouched and that the user should take care to rename themselves? Or maybe even provide both solutions?

Centrus007:
> What if we remove the parameters for "ignore" for both unknown and for private. And we just have the delete versions. As in delete private and delete unknown. Have both delete parameters default to None. Then, if a private or an unknown exist in the dataset and either of these parameters are None then an error is raised saying the user needs to explicitly declare what they want to happen with those tags.

I do think I prefer that approach. :) 

I also think there should be a provision to delete *all* unknown and delete *all* private tags. If this is what the user wants, it could be quite cumbersome for them to be forced to explicitly supply each private/unknown tag that they want to delete.

src/pymedphys/dicom/_level2/anonymise.py
SimonBiggs: I think we should create a family of anonymise CLI functions. One should be of the form:

```
pymedphys dicom anonymise-file input_file output_file
```

Which completely leaves the filename up the user.

Another:

```
pymedphys dicom anonymise-directory input_dir output_dir
```

Which recursively walks through a directory, anonymises the DICOM files and copies them into the new directory with the same names and the same directory tree structure.

I think from there we need to treat the user like a responsible adult. We are anonymising the DICOM header, nothing more. Not the image data, not the file names. I don't even believe a user warning is needed.

What do you think? I think we need to adopt a "assume user is responsible adult" stance, otherwise our interfaces may become very convoluted and the logic within our functions over complicated beyond doing the task at hand.

We could have a third function, not DICOM specific, that searches a range of supported file formats as well as directory and file names, for a string that looks like a name. That function could then be recommended to be run after the DICOM anonymiser within the documentation?

SimonBiggs:
Yup, I think that needs a bit of thought. I can't think of it just now, but
maybe even the concept of unknown and private can be handled together? Not
sure though.

I'm about to get on a flight so I'll lose contact soon :)

Centrus007:
One option:

```python
def anonymise_dicom_dataset(ds, replace_values=True, keywords_to_keep=None,
                            delete_private_tags=False, delete_unknown_tags=False,
                            copy_dataset=True)
```
So an error is thrown if either private or unknown tags are found. The user could then use `keywords_to_keep` in conjuction with `delete_private_tags=True` and/or `delete_unknown_tags=True` to avoid deletion of any private tags that they want to keep?

SimonBiggs:
> One option:
> 
> ```python
> def anonymise_dicom_dataset(ds, replace_values=True, keywords_to_keep=None,
>                             delete_private_tags=False, delete_unknown_tags=False,
>                             copy_dataset=True)
> ```
> 
> So an error is thrown if either private or unknown tags are found. `delete_private_tags` and `delete_unknown_tags` delete all private/unknown tags if set to True, but the user could use these in conjunction with `keywords_to_keep` to avoid deletion of any private/unknown tags that they want to keep?

Except, instead of having the delete tags default to False, have them default to None. That way, if a tag is found and the respective parameter if None, then raise an error saying the user needs to make a choice. Setting delete tags to False, explicitly, will result in all of those respective tags being kept and no error being raised.

Apart from that, yes, I like that API.

SimonBiggs:
Maybe instead of "keywords to keep" something that means "keywords to leave unchanged". That way, it would be used for delete actions, but also anonymisation actions. As in, if I said `keywords_to_leave_unchanged = ['PatientName']` then patient name would not undergo anonymisation.

src/pymedphys/dicom/_level2/anonymise.py
SimonBiggs: > I think we should create a family of anonymise CLI functions. One should be of the form:
> 
> ```
> pymedphys dicom anonymise-file input_file output_file
> ```
> Which completely leaves the filename up the user.
> 
> Another:
> 
> ```
> pymedphys dicom anonymise-directory input_dir output_dir
> ```
> Which recursively walks through a directory, anonymises the DICOM files and copies them into the new directory with the same names and the same directory tree structure.


Or maybe simply just:

```
pymedphys dicom anonymise input_file_dir_or_zip output_file_dir_or_zip
```

I'm not sure...

Centrus007: To be honest, I don't like the recursive dir copy with matching structure. It's unnecessarily slow and, IMHO, introduces confusion. What if you have other non-DICOM files in there? Do you copy them across or not? If so, added time. If not, loss of structure that might be important...

SimonBiggs:
Fair point. The user story I am trying to address is "I have a collection
of DICOM files in a directory tree and I want to anonymise the DICOM files
before sending to a third party." I'm open to not doing it the way you
mention bit the result of the "anonymisation append" method is the user
then needs to delete all of the old dicom files manually.

Maybe there's a better way to meet that user story though...

src/pymedphys/dicom/_level2/anonymise.py
SimonBiggs: Fair point. The user story I am trying to address is "I have a collection of DICOM files in a directory tree and I want to anonymise the DICOM files before sending to a third party."

I'm open to not doing it the way I mention but the result of the "anonymisation name append" method is that the user then needs to copy the directory themselves (if they want to keep the non-anonymised version) and then within the copied directory they need to delete all DICOM files that don't have "anonymised", making sure to not miss one.

SimonBiggs: Also when I say "user story" I mean the following:

https://www.mountaingoatsoftware.com/agile/user-stories

SimonBiggs: > It's unnecessarily slow 

How is it any slower?

SimonBiggs:
>  I'm about to get on a flight so I'll lose contact soon :)

Turns out they wifi on planes now :) so I still get to annoy you :)

src/pymedphys/dicom/_level2/anonymise.py
Centrus007: > to copy the directory themselves (if they want to keep the non-anonymised version) and then within the copied directory they need to delete all DICOM files that don't have "anonymised" in the file name, making sure to not miss one.

The user would copy the dir themselves and then anonymise in-place, I think. So no manual deletion necessary. I admit that the manual copy would be nice to avoid though. But then again, I personally would rarely want to maintain the original dir structure... so a manual copy is almost necessary.

I have noticed in my own tinkering that a manual copy is fantastically fast on a Windows machine. Python's `shutil` and `xcopy` are ridiculously slow by comparison when copying recursively. Maybe I haven't been clever enough - please try yourself! I also just found [this SO answer](https://stackoverflow.com/a/28584857/4665458) that might address the issue.

Centrus007: Another related but separate point: prepending "anonymous" or similar to the filename clearly flag to the user that the file has actually been anonymised. If we instead maintain the original filename, the user is relying on the dir name to flag anonymisation - which falls apart if they ever copy the file out of that dir...

Of course, they could open the DICOM to inspect it for anonymity. But this is inconvenient - DICOM files are binary, after all.

Centrus007:
> That way, if a tag is found and the respective parameter is None, then raise an error saying the user needs to make a choice. Setting delete tags to False, explicitly, will result in all of those respective tags being kept and no error being raised.

Gotcha. But then what's the best way to implement an "all" case?

Centrus007:
> Turns out they wifi on planes now :) so I still get to annoy you :)

Hehehe, I was actually going to miss you!

Centrus007:
> Maybe instead of "keywords to keep" something that means "keywords to leave unchanged". That way, it would be used for delete actions, but also anonymisation actions. As in, if I said keywords_to_leave_unchanged = ['PatientName'] then patient name would not undergo anonymisation.

Yep, that makes sense

src/pymedphys/dicom/_level2/anonymise.py
SimonBiggs: > Python's shutil and xcopy are ridiculously slow by comparison when copying recursively.

I wouldn't imagine using xcopy, instead just output the DICOM files to a the new location. No slower than what is being done to write a new anonymised dicom file with annonymised prepended.

Centrus007: > I wouldn't imagine using xcopy, instead just output the DICOM files to a the new location. No slower than what is being done to write a new anonymised dicom file with annonymised prepended.

So not copying any non-DICOM files?

SimonBiggs: Yeah I suppose not... Hmm...

What about this, have it anonymise inplace, don't append anything, overwrite the old files, then provide a "is anonymised" CLI function if someone wants to check if it has been anonymised.

The "is anonymised" could use pydicom's "read header only" flag so that it would run really quick even for large pixel data files. And it then goes and verifies if a file has been anonymised directly, by reading the header.

Some files, like "BFG cleaner" create a log, and they also have a "this is what I am going to do, is that okay" prompt. (A prompt which can be removed with a "no-interaction" flag).

As in, if it is in place, and overwrites old files, then it could print a list of all subfiles it will overwrite before it does anything.

Then once it has done it, it could write out a log of all files it anonymised into the top directory, a date and time of when it did it, and potentially a list of tags that were found that needed modification within the files. If tags were deleted a lost of tags that were deleted.

And then maybe even a link to the PyMedPhys DICOM anonymisation CLI docs at the top of the log so that the third party who receives the files knows that tool was used to undergo the anonymisation...

SimonBiggs:
> Gotcha. But then what's the best way to implement an "all" case?

What do you mean by that?

SimonBiggs:
> > Maybe instead of "keywords to keep" something that means "keywords to leave unchanged". That way, it would be used for delete actions, but also anonymisation actions. As in, if I said keywords_to_leave_unchanged = ['PatientName'] then patient name would not undergo anonymisation.
> 
> Yep, that makes sense

Maybe a simple parameter name would be just `leave_unchanged`. Then in the docs describe it as a list of DICOM tags provided as strings. I would imagine it would be important to raise an error if a tag that is set to be left unchanged was not found within the DICOM file.

SimonBiggs:
> Hehehe, I was actually going to miss you!

:)

Centrus007:
> What do you mean by that?

What do you think is the best way to implement the case where all private or unknown tags are deleted (i.e., without having the user manually specify each one in a list)?

SimonBiggs:
Have the user set both delete unknown and delete private to True.

Centrus007:
> Have the user set both delete unknown and delete private to True.

Ah I see, so you're going for a None, False, True, case? Not a None, [list of tags], 'all' case?

SimonBiggs:
Yup, that's it. And then a "leave_unchanged" for a partial delete case.

Centrus007:
> Yup, that's it. And then a "leave_unchanged" for a partial delete case.

Perfect.

src/pymedphys/dicom/_level2/anonymise.py
Centrus007: I think an `is_anonymised()` function is worthwhile (e.g. for broad repo checks), but I still really think we ought to simply rename the files if anonymisation is not in place. It really is standard behaviour - every anonymiser I've used does this, including RayStation itself. Pylinac even does it for its log file anonymiser.

- I think your concern at having to delete the originals is solved simply by the user choosing the `in_place` option.

- I think it is far simpler, more expected and more novice-friendly to simply rename the file (with an "anonymous" prepend) for non-in-place anonymisation. It's a very readily visible flag that the file has indeed been anonymised that neither relies on the file remaining in its dir structure, nor requires a CLI run of `is_anonymised()`.

SimonBiggs:
Yup, you've convinced (appeal to authority is a powerful thing ;) )

Centrus007:
A small hiccough I've come across: private tags often don't have keywords - they are instead accessed by their tag number. This is a bit cumbersome for the user to supply and for us to handle...

Shall we just have the option of either deleting all of none of the private tags for now and implement individual tag removal if requested?

SimonBiggs:
Sounds good. I don't think we should default to either though. Still
haven't error initially if any private tags exist...maybe even the error
message for "it has private tags" could print the tags themselves and their
contents out?

Centrus007:
> Sounds good. I don't think we should default to either though. Still
> haven't error initially if any private tags exist...maybe even the error
> message for "it has private tags" could print the tags themselves and their
> contents out?

Why is that? I have almost never used private tags. And if there's an issue the user can simply re-anonymise with `delete_private_tags` set to `False`... Implementing a tristate feels like overkill, and if we default to anything, I'd prefer the option that minimises risk of identifying info being present.

Centrus007:
> Sounds good. I don't think we should default to either though. Still
> haven't error initially if any private tags exist...maybe even the error
> message for "it has private tags" could print the tags themselves and their
> contents out?

> Why is that? I have almost never used private tags. And if there's an issue the user can simply re-anonymise with delete_private_tags=False... Implementing a tristate feels like overkill, and in the boolean case, if we default to anything, I'd prefer the option that minimises risk of identifying info being present.

Actually, thinking about it more, I'm starting to come around. Just be aware that the implication of insisting on such a default will be that the user is forced to run a command like:

```bash
pymedphys dicom anonymise-file --input_filepath <filepath> --delete_private_tags
```
or
```bash
pymedphys dicom anonymise-file --input_filepath <filepath> --keep_private_tags
```

I jsut think it would be nice for the user to be able to run something like 

```bash
pymedphys dicom anonymise-file --input_filepath <filepath>
```
only, if they are happy with the defaults...

SimonBiggs:
One thought is that brachy mostly uses private tags. Therefore for
brachytherapy that will essentially delete the whole plan... If it's in
place that could be quite an issue...

Centrus007:
> One thought is that brachy mostly uses private tags. Therefore for
brachytherapy that will essentially delete the whole plan... If it's in
place that could be quite an issue..

Well, any anonymisation in-place with no backup DICOM is an issue. Protecting private tags doesn't really make any difference here.

SimonBiggs:
More like, prepping a set of files for sending. The code essentially the deletes the contents of the plan file making it unusable. Then the user of the code sends it off. In that scenario I would have preferred an initial error message presenting to the user that these DICOM files have private tags, here are the private tags and their contents presented in the error message, as well as instructions on how to rerun the script to inform the code to either delete them all or keep them all.

When an option comes up to do something implicitly, or explicitly, I'm generally in favour of explicit.

src/pymedphys/dicom/_level2/anonymise.py
```diff
@@ -213,6 +215,61 @@ def anonymise_dicom_dataset(ds, replace_values=True, delete_private_tags=True,
...
 
 
+def anonymise_dicom_file(dicom_filepath, overwrite_file=False, replace_values=True,
+                         keywords_to_leave_unchanged=None, delete_private_tags=True,
+                         delete_unknown_tags=None):
+
+    ds = pydicom.dcmread(dicom_filepath)
+
+    if overwrite_file:
+        dicom_anon_filepath = dicom_filepath
```
SimonBiggs: You mentioned you always wanted to append annonymised at the end? I've come around. Happy for that to be the case, so it seems the logic should be to pass a flag "delete original files" or something of the sort. And the steps would be to always append anonymised to the file name, but delete the old if the flag asks for it.

src/pymedphys/dicom/_level2/anonymise.py
```diff
@@ -213,6 +215,61 @@ def anonymise_dicom_dataset(ds, replace_values=True, delete_private_tags=True,
...
+                                      replace_values=replace_values,
+                                      keywords_to_leave_unchanged=keywords_to_leave_unchanged,
+                                      delete_private_tags=delete_private_tags,
+                                      delete_unknown_tags=delete_unknown_tags,
+                                      copy_dataset=False)
+
+    ds_anon.save_as(dicom_anon_filepath)
+
+
+def anonymise_dicom_files_in_folder(folder, overwrite_files=False, replace_values=True,
```
SimonBiggs: `anonymise_dicom_directory`?

src/pymedphys/dicom/_level2/anonymise.py
```diff
@@ -213,6 +215,61 @@ def anonymise_dicom_dataset(ds, replace_values=True, delete_private_tags=True,
...
+                                    keywords_to_leave_unchanged=None, delete_private_tags=True,
+                                    delete_unknown_tags=None):
+
+    for dicom_filepath in glob(folder + '/**/*.dcm', recursive=True):
```
SimonBiggs: Does this globstring work for top level DICOM files?

.vscode/settings.json
```diff
@@ -13,6 +13,7 @@
     "editor.tabSize": 4
   },
   "editor.rulers": [
-    79
-  ]
+    80,
+    100
```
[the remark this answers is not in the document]
Centrus007: I think this is one PEP convention that is a little ridiculous. Many pythonistas agree. E.g. see [The Best of the Best Practices Guide for Python](https://gist.github.com/sloria/7001839).

Fair enough on 79 vs 80. But I basically figured having a ruler for 79/80 to indicate PEP compliance and a ruler for 100 to indicate a more reasonable max line length would be a better combination.

72 is just flat our ridiculous. It severely limits code readability. Fewer and fewer machines are limited to an 80 char line length and, generally, more (unnecessary) wrapping means more unreadble code. ESPECIALLY for python, for which descriptive variable names are encouraged and indentation is so embedded in the language.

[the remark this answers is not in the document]
Centrus007: Also note the following in PEP:

> "Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters."

src/pymedphys/dicom/_level2/anonymise.py
```diff
@@ -108,93 +110,93 @@
                                        'UI': "12345678"}
 
 
-def anonymise_dicom_dataset(ds, replace_values=True, delete_private_tags=True,
-                            keywords_to_keep=None, ignore_unknown_tags=False,
-                            copy=True):
+def anonymise_dicom_dataset(ds, replace_values=True, keywords_to_leave_unchanged=None,
```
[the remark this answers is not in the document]
Centrus007: Empty lists as default parameters are a bit of a python no-no - they're subject to unexpected behaviour. The python docs suggest `None` for these cases: see [8.6. Function definitions](https://docs.python.org/3/reference/compound_stmts.html#function-definitions) and scroll down a little.

src/pymedphys/dicom/_level2/anonymise.py
Centrus007: I was actually only suggesting that we append anonymised for non-in-place anonymisation. But now I think on it, maybe it is best to do for all cases and just delete the original files...

src/pymedphys/dicom/_level2/anonymise.py
Centrus007: Yes. See example at the bottom of the [glob docs](https://docs.python.org/3/library/glob.html)

src/pymedphys/dicom/_level2/anonymise.py
Centrus007: Sure :)

src/pymedphys/dicom/_level2/anonymise.py
Centrus007: though an `os.path.join()` is probably in order...

.vscode/settings.json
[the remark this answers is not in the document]
SimonBiggs: > "Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters."

Interesting. Never read that bit. I'm certainly willing to be swayed. ... I do often really like having the four side by side screens though... much more than 79 characters and I do lose that flexibility...

src/pymedphys/dicom/_level2/anonymise.py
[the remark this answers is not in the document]
SimonBiggs: Yes! I forgot about that :/

Oops.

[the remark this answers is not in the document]
SimonBiggs: Mutability and all that.

src/pymedphys/dicom/_level2/anonymise.py
SimonBiggs: > I was actually only suggesting that we append anonymised for non-in-place anonymisation. But now I think on it, maybe it is best to do for all cases and just delete the original files...

Sounds good to me :)

src/pymedphys/dicom/_level2/anonymise.py
SimonBiggs: > though an `os.path.join()` is probably in order...

glob can handle forward slashes on Windows just fine. So it's actually a non-issue in this case

src/pymedphys/dicom/_level2/anonymise.py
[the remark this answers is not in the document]
SimonBiggs: That or pass a tuple as default...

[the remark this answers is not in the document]
SimonBiggs: Thinking about this further, the logic on the code is simplified and the default is made relatively clear if the following is used:

```
keywords_to_leave_unchanged=()
```

or

```
keywords_to_leave_unchanged=tuple()
```

It's not a deal breaker for me though if you'd like to stick to `None` then instantiate the mutable list. But in this case, we don't need mutability... So I don't think list is the object for the job.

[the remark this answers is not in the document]
Centrus007: Happy to pass a tuple, I think :)

Centrus007:
> More like, prepping a set of files for sending. The code essentially the deletes the contents of the plan file making it unusable. Then the user of the code sends it off. In that scenario I would have preferred an initial error message presenting to the user that these DICOM files have private tags, here are the private tags and their contents presented in the error message, as well as instructions on how to rerun the script to inform the code to either delete them all or keep them all.
> 
> When an option comes up to do something implicitly, or explicitly, I'm generally in favour of explicit.

You've convinced me, especially since I realised there's a (pretty standard) alternative to the user flow I described above.

If private tags exist, and the user supplies no flag indicating whether to keep or delete private tags, like so:

```bash
> pymedphys dicom anonymise my.dcm
```
the CLI then throws something like
```bash
> pymedphys dicom anonymise my.dcm
WARNING: This DICOM file contains private tags, which may contain identifying information.
Would you like to delete these private tags (Y), keep these private tags (N) or cancel (C)?
>>
```
And then perhaps weave in an instruction that the user can use the `--delete-private-tags` or `keep-private-tags` flag to avoid the warning message?

src/pymedphys/dicom/_level2/anonymise.py
```diff
@@ -108,93 +110,93 @@
...
+                            copy_dataset=True):
     r"""A simple tool to anonymise a DICOM file.
 
     Parameters
     ----------
-    ds
-        The DICOM file to be anonymised. `ds` must represent a valid
-        DICOM file in the form of a `pydicom Dataset` - ordinarily
-        returned by `pydicom.dcmread()`.
+    ds : pydicom.dataset.Dataset
```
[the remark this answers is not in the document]
Centrus007: You really do want to move away from `ds`, don't you? :D

The symmetry *is* there. `ds` is short for `dataset` after all. More than that, a developer should recognise `ds` as referring *specifically to a pydicom dataset`, due to convention. The variable name `dicom dataset` doesn't carry that same *specific* recognition.

It's kind of like "int" being universally recognised as denoting an "integer data type" (as opposed to a variable containing an integer value), but on a far, far smaller scale of course.

.vscode/settings.json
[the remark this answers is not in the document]
Centrus007: Do your "screens" have to be equal in width (i.e. do they all contain text editors)?

I tend to have one text editor next to Cmder, split across one screen so that Cmder gets less than half of the screen and VS Code gets more than half. This is particularly easy in Windows 10, which can handle unequal window snaps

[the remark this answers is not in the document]
Centrus007: 80 chars doesn't need anywhere near a half screen. Assuming you're running with 16:9 and a reasonable font size...

src/pymedphys/dicom/_level2/anonymise.py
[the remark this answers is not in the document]
Centrus007: > [EDIT] This has been out of my mind for a year or so now. I need to run a search over PyMedPhys for "=[" to see if I have made this mistake within the codebase. Potentially pylint picks it up I think... But I'm not certain. It's true with any mutable object, so numpy arrays are also an issue. [/EDIT]

I mentally chuckled to myself when I noticed that "=[" looks like a sad face...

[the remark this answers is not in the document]
SimonBiggs: > I mentally chuckled to myself when I noticed that "=[" looks like a sad face...

Me too :)

.vscode/settings.json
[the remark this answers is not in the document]
SimonBiggs: Two screens at full 1080p:

image

... I really like the 79 character limit as it allows me to do the above ...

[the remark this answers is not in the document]
Centrus007: The side by side diff case is fair. I do squeeze in with 99 chars, but I lose the file preview in the scrollbar. And the sidebar is a little narrow.

image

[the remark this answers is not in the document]
Centrus007: There's still the excessive scrolling point. Not to mention many examples of contiguous text that you really don't want to wrap (like URLs) simply not fitting in 72/79, especially with indenting.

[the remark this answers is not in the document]
Centrus007: As a compromise, how about the following guideline from [BigchainDB](http://docs.bigchaindb.com/projects/contributing/en/latest/cross-project-policies/python-style-guide.html)?

> **Maximum Line Length**
> PEP8 has some maximum line length guidelines, starting with “Limit all lines to a maximum of 79 characters” but “for flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.”
> 
> We discussed this at length, and it seems that the consensus is: try to keep line lengths less than 79/72 characters, unless you have a special situation where longer lines would improve readability. (The basic reason is that 79/72 works for everyone, and BigchainDB is an open source project.) As a hard limit, keep all lines less than 119 characters (which is the width of GitHub code review).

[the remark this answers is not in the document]
Centrus007: I've actually even found 99 to be restrictive. So perhaps 119 to be used in extreme cases only (like URLs), since it matches GitHub? ANd I will endeavour to stick to 72/79 from now on?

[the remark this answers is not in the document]
SimonBiggs: > I've actually even found 99 to be restrictive. So perhaps 119 to be used in extreme cases only (like URLs), since it matches GitHub? And I will endeavour to stick to 72/79 from now on?

Seems like a good compromise. Worthwhile having rulers at 72, 79, and 119.

Centrus007:
@SimonBiggs, another thing that comes to mind:

Would you agree to omitting "dicom" from the function names within the dicom module? I.e.
- `anonymise_dicom_dataset()` --> `anonymise_dataset()`
- `anonymise_dicom_file()` --> `anonymise_file()`
- `anonymise_dicom_directory()` --> `anonymise_directory()`

This is because "dicom" is already implicit in the module, so these constitute [redundant labelling](https://gist.github.com/sloria/7001839#naming).

SimonBiggs:
I'm open to it.

In the end I also suspect our primary API for this one might be object
based, as in `dicom.anonymise()`

tests/test_dicom/test_dicom_anonymise.py
```diff
@@ -1,7 +1,84 @@
-from pprint import pprint
-from pydicom import datadict
-from pymedphys.dicom import anonymise_dicom_dataset, IDENTIFYING_KEYWORDS
-from pymedphys.dicom import BaselineDicomDictionary
+
```
Centrus007: @SimonBiggs,
 Please note the following tests. Happy to add to these soon.

This PR is getting big - shall we aim to merge this in once you're happy and cover the actual interface in another PR?

If you think more testing is required, I understand. Just want to see if we're on the same page on PR scope.

SimonBiggs:
> This PR is getting big - shall we aim to merge this in once you're happy and cover the actual interface in another PR?
>
> If you think more testing is required, I understand. Just want to see if we're on the same page on PR scope.

I'd be happy to do a review with the aim to merge in once the pipelines are passing. I think it is generally a good idea to keep pull requests small and bite sized :).

tests/test_dicom/test_dicom_anonymise.py
```diff
@@ -1,7 +1,84 @@
...
+    anonymous_filepath_moved = pjoin(dirpath_expected_to_pass,
+                                     basename(anonymous_filepath))
+
+    makedirs(dirpath_expected_to_pass, exist_ok=True)
+    replace(anonymous_filepath, anonymous_filepath_moved)
+
+    assert is_anonymised_directory(dirpath_expected_to_pass)
+
+    remove(anonymous_filepath_moved)
+    rmdir(dirpath_expected_to_pass)
```
SimonBiggs: When tox is run in parallel mode across multiple virtual environments (as in testing against py36 and py37 simultaneously) it struggles with tests of this form. Reason being, one test is creating the same files that another is trying to delete etc.

To circumvent that I add a random string to the end of the filenames. See here:

https://github.com/pymedphys/pymedphys/blob/master/tests/test_dicom/test_header_tweaks.py#L40-L44

Another point, is if this code does error for whatever reason, the `remove` and `rmdir` commands are not run. Would you be able to use the `try`/`finally` pattern here to make sure that the directory and files are always removed even if the test fails?

See: https://github.com/pymedphys/pymedphys/blob/master/tests/test_dicom/test_header_tweaks.py#L47-L58

tests/test_dicom/test_dicom_anonymise.py
```diff
@@ -1,7 +1,84 @@
...
+
+    # anonymous_filepath = pjoin(DATA_DIR, "RP.almost_anonymised.dcm")
+
+    dicom_anon_filepath = anonymise_file(dicom_test_filepath,
+                                         delete_private_tags=False)
+
+    assert not is_anonymised_file(dicom_anon_filepath,
+                                  ignore_private_tags=False)
+    assert is_anonymised_file(dicom_anon_filepath,
+                              ignore_private_tags=True)
```
SimonBiggs: Thanks for these tests @Centrus007, really helps reviewing!

tests/test_dicom/test_dicom_anonymise.py
```diff
@@ -1,7 +1,84 @@
...
+    assert not is_anonymised_dataset(ds)
+
+    ds_anon_manual = ds
+    ds_anon_manual.PatientName = "Anonymous"
+
+    ds_anon_func = anonymise_dataset(ds, delete_private_tags=False,
+                                     copy_dataset=True)
+
+    # Patient Name should've been the only non-anonymous value. If not,
+    # test data has been changed
```
SimonBiggs: Let me know what you think. I find when tests like this highly depend on the data they're passed there is a lot of value in having the data be created by the test itself in a transparent way. All able to be read in the same place.

For example, see: https://github.com/pymedphys/pymedphys/blob/master/tests/test_dicom/test_dicom_base.py#L56-L69

Something like:

```python
import uuid
import os

from pymedphys.dicom import (
    DicomBase, is_anonymised_dataset, is_anonymised_directory)

directory_to_use = str(uuid.uuid4())

names_to_use = ['Python^Monte', 'MacDonald^George']

expected_dicom_file = DicomBase.from_dict({
	'Manufacturer': 'PyMedPhys',
	'PatientName': 'Anonymous'
})

filepaths = []
for i, name in enumerate(names_to_use):
	dicom = DicomBase.from_dict({
	    'Manufacturer': 'PyMedPhys',
	    'PatientName': name
	})
	assert not is_anonymised_dataset(dicom.dataset)
	
	filepath = os.path.join(directory_to_use, "{}.dcm".format(i))
    filepaths.append(filepath)

	dicom.to_file(filepath)

    assert not is_anonymised_file(filepath)


assert not is_anonymised_directory(directory_to_use)
anonymise_file(filepaths[0])
assert is_anonymised(filepaths[0])

# Check that one anonymous file doesn't trigger the directory
assert not is_anonymised_directory(directory_to_use)

anonymise_directory(directory_to_use)
assert is_anonymised_directory(directory_to_use)
assert is_anonymised(filepaths[1])

# etc...	
```

tests/test_dicom/test_dicom_anonymise.py
Centrus007: Add the try/finally. But I might handle the parallel issue in another branch, if that's okay? It would actually require a bit of a rework to do properly...

Centrus007: Actually, that latest commit achieves most of the safety you're after, I think. Still not perfect, but very close!

tests/test_dicom/test_dicom_anonymise.py
```diff
@@ -72,13 +73,14 @@ def test_is_anonymised_directory():
...
+        replace(anonymous_filepath, anonymous_filepath_moved)
 
-    assert is_anonymised_directory(dirpath_expected_to_pass)
-
-    remove(anonymous_filepath_moved)
-    rmdir(dirpath_expected_to_pass)
+        assert is_anonymised_directory(dirpath_expected_to_pass)
+    finally:
+        remove_file(anonymous_filepath_moved)
+        remove_dir(dirpath_expected_to_pass)
```
SimonBiggs: Thanks @Centrus007 :)

tests/test_dicom/test_dicom_anonymise.py
```diff
@@ -68,18 +70,21 @@ def test_is_anonymised_directory():
     dirpath_expected_to_fail = DATA_DIR
     assert not is_anonymised_directory(dirpath_expected_to_fail)
 
+    dirpath_expected_to_pass = pjoin(DATA_DIR, 'Anonymous_{}'.format(uuid4()))
```
SimonBiggs: Thanks @Centrus007 :)

tests/test_dicom/test_dicom_anonymise.py
SimonBiggs: > But I might handle the parallel issue in another branch, if that's okay? 

Yup, certainly okay

PR #137, opened 2019-04-07
[RFR] add logo

SimonBiggs:
@Centrus007 what do you think?

https://github.com/pymedphys/pymedphys/blob/logo-iteration/notebooks/logo/logo.svg


SimonBiggs:
Logo turned into favicons for each OS.

[favicon_package_v0.16.zip](https://github.com/pymedphys/pymedphys/files/3051503/favicon_package_v0.16.zip)

SimonBiggs:
@Centrus007 what do you think about the two following, which one do you prefer?

image

image

Centrus007:
Why not match the blue and yellow of Python?

Python logo

Pandas logo

NumPy logo

![PyDicom logo]https://pydicom.github.io/pydicom/stable/_static/pydicom_flat_black.svg)

Centrus007:
Otherwise, very nice!!!

Centrus007:
I don't suppose we can lengthen the tail for a bit more left/right symmetry? :p

SimonBiggs:
Yes, I like it. Those colours will be quite perfect I think. And yup, tail lengthening coming up :)

PR #158, opened 2019-04-16
[RFR] Extract film plane from dicom

SimonBiggs:


SimonBiggs:
Not all changes within the library here are tested unfortunately. At some point (not sure when) I can write more tests for this one if it is thought to need it...

packages/pymedphys_dicom/tests/test_structure.py
```diff
@@ -0,0 +1,99 @@
...
+            [5, 6, 3, 5],
+            [7, 7, 5, 7],
+            [3, 3, 3, 3]
+        ]
+    ]
+)
+
+
+def test_pull_structure():
+    dicom_structure = dicom_dataset_from_dict({
```
Centrus007: I'm still a *little* nervous about relying solely on datasets created from `dicom_dataset_from_dict()` for these purposes, at least until we have built a super-solid level of confidence in that function.

SimonBiggs: Fair point. I'd be very keen to build confidence as much as we can. I personally really like this test format...

SimonBiggs:
Thanks Matt :) if you're happy with it make sure you do the merge so you're on the commit. :)

packages/pymedphys_dicom/tests/test_structure.py
Centrus007: Me too. Confidence is my only concern. Other than that, it is beautiful and flexible.

PR #166, opened 2019-04-17
[RFR] Enforce layer independence, and in light of module layers apply refactor

SimonBiggs:


layers.yml
```diff
@@ -28,22 +28,91 @@ Top Level Modules:
...
-        - xlwings
+        # Module Level 5
         - logfile
+
+        # Module Level 4
         - trf
+
+        # Module Level 3
+        - xlwings
         - msq
```
SimonBiggs: `msq` here depends on `mudensity`. I don't think it should. This then got me to look at the following function:

https://github.com/pymedphys/pymedphys/blob/508d7b788e370a5c7710053d22042ef56456133a/src/pymedphys/msq/_level3/msqfieldcompare.py#L46-L50

It's the only reason `msq` is level 3, and therefore the only reason `trf` is level 4 and `logfile` level 5.

By explicitly seeing the leveling it has highlighted I need to make a small refactor.

PR #172, opened 2019-04-19
RFR: Refactor DICOM docs

Centrus007:
This is just me having a bit of a play with the docs, but I thought I'd add a pull request for easy visibility :)

### DIFF
--- a/src/pymedphys/dicom/__init__.py
+++ b/src/pymedphys/dicom/__init__.py
@@ -23,46 +23,27 @@
 # You should have received a copy of the Apache-2.0 along with this
 # program. If not, see <http://www.apache.org/licenses/LICENSE-2.0>.
 
-
 """A DICOM toolbox. Available functions include:
 
-General
--------
 >>> from pymedphys.dicom import (
+...
+...     # General functions
 ...     anonymise_dataset,
 ...     anonymise_file,
 ...     anonymise_directory,
 ...     is_anonymised_dataset,
 ...     is_anonymised_file,
-...     is_anonymised_directory
-)
-
-
-RT Dose
--------
-Functions related to manipulating or extracting information from DICOM
-RT Dose datasets.
->>> from pymedphys.dicom import (
+...     is_anonymised_directory,
+...
+...     # RT Dose related functions
 ...     extract_iec_patient_xyz,
 ...     extract_iec_fixed_xyz,
-...     extract_dicom_patient_xyz
-)
-
-
-RT Structure
-------------
-Functions related to manipulating or extracting information from DICOM
-RT Structure datasets.
->>> from pymedphys.dicom import (
-...     get_structure_aligned_cube
-)
-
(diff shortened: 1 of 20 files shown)


src/pymedphys/dicom/__init__.py
```diff
@@ -24,22 +24,48 @@
 # program. If not, see <http://www.apache.org/licenses/LICENSE-2.0>.
 
 
-"""A Dicom toolbox built ontop of the pydicom library.
+"""A DICOM toolbox. Available functions include:
 
-Available Functions
--------------------
+General
```
Centrus007: Basically was thinking it would be useful to expand the docs with explanations/examples, and I figured it might help to start by categorising our DICOM functions. Things blew up in complexity from there, so I thought I'd stop and see what your thughts were, @SimonBiggs, @kingrpaul and anyone else :)

src/pymedphys/dicom/__init__.py
```diff
@@ -24,22 +24,48 @@
...
+Functions related to manipulating or extracting information from DICOM
+RT Structure datasets.
+>>> from pymedphys.dicom import (
+...     get_structure_aligned_cube
+)
+
+
+DICOM RT Combinations
+---------------------
+Functions that extracting information from a combination of DICOM
```
SimonBiggs:
```suggestion
Functions that extract information from a combination of DICOM
```

src/pymedphys/dicom/__init__.py
SimonBiggs: Go for it. Take the liberty to rework it :)

Centrus007:
Actually, I think this latest iteration might work a little better... what do you think? Things to note:

- Still have a summary of all functions at the top, but with comments in the code block indicating categories. I think this looks a little neater than my initial approach, which tried to separate both the summary and the API into their own sections. This would've been unnecessary duplication without the ability to easily combine the duplicated sections (since we have to import the module docstring in one chunk, I think).
- API separated into subsections with short descriptions (and maybe examples or background material (I'm thinking coordinate system stuff?), which now have links on the sidebar.

image

Centrus007:
Ready for review, I think. @SimonBiggs :)

SimonBiggs:
I've just made a private build of this branch over at:

https://pymedphys.com/en/try-reorganisation-of-dicom-docs/

SimonBiggs:
I just made a couple of changes. My primary aim was to actually bump the DICOM docs up the list to the top, it is currently the most well documented library interface, so it should be the first one that is clicked on.

In the process I also got slightly carried away and fixed up some of the file indexing. I'll let you review that, and then merge when you're happy.

Centrus007:
> I really, really like that change to the DICOM docs!

I'm glad!

Centrus007:
LGTM. Will merge when tests pass (I updated branch with master first)

Issue #176, opened 2019-04-19
Use a higher res logo for SA Health


SimonBiggs:
@Centrus007 what about if we used the following logo?

https://www.healthdirect.gov.au/partners/sa-health

Centrus007:
Sure. :)

Centrus007:
I'm having a play and the aspect ratio makes pretty formatting a bit of a challenge...

Centrus007:
How's this? Matching width means other logos will be dwarfed by GOSA

![image](https://user-images.githubusercontent.com/19767103/56450654-b5d59c00-6366-11e9-8ef4-7ac1c2247367.png)

Centrus007:
Might be nice to centre the logos based on the text width...

Centrus007:
Actually, the standard sphinx `:align: center` seems to have worked well enough.

Half screen:

![image](https://user-images.githubusercontent.com/19767103/56450728-28df1280-6367-11e9-83be-946014acac97.png)

Full screen:

![image](https://user-images.githubusercontent.com/19767103/56450731-37c5c500-6367-11e9-82df-c7b9f68c0590.png)

SimonBiggs:
Yup, looks good to me.

PR #179, opened 2019-04-20
RFR: Replace GOSA logo.

Centrus007:
Center logos
Add python and rst specific rulers.


Centrus007:
See #176 for screenshots, since logo won't appear until merge, I believe.

PR #181, opened 2019-04-20
[RFR] The BIG Split

SimonBiggs:


SimonBiggs:
It's not ready for merging, still a ways to go. But I am keen for your input on the structure of things...

SimonBiggs:
As a brief description, first make sure you've installed yarn as described in the contributor docs:

http://pymedphys.com/en/stable/developer/contributing.html#git-git-lfs-and-yarn

Then in the root directory run in a command prompt `yarn`. This will step through each of the packages and run `pip install -e .` in each directory to install each package.

SimonBiggs:
The package names certainly aren't locked in by the way. ... and pmpdicom etc is still an option.

Out of interest, _ and - actually map to the same package on pypi. So `pip install pymedphys-dicom` is the same as `pip install pymedphys_dicom`.

SimonBiggs:
So, everything but the docs are back up and running in the new format.

There is a heap of stuff that I made quite WET (as opposed to DRY), just to get an idea of what I need to make things work. I'll need to use some orchestration tools to DRY up all that WET.

So, next up, docs. Will do that another day though.

Feedback is appreciated :).

SimonBiggs:
Instead of changing the docs to match the new API I am going to create wrapper modules within the top pymedphys module. Essentially the aim is to mimic the previous API and once we have decided upon the new API that pymedphys will export we can add deprecation notices.

See https://github.com/pymedphys/pymedphys/issues/184 for more information.

Once the docs are passing with the wrappers in place I propose we merge this. The next step is to then remove the `from level import *` logic. Then I need to make that dependency graph tool. From having that dependency graph tool I shall then use that tool to dynamically maintain each packages setup.py, versions, install_requires, and their package.json. Each of those files will be managed automatically and tests for agreement using integrity tests.

SimonBiggs:
Gee, clever git, merging that one in. No conflict resolution needed, it was a straight fast forward! :)

Centrus007:
Just letting you know I won't get a chance to have a proper look until tomorrow (possibly later tonight, but unlikely).

Looking forward to perusing!

SimonBiggs:
All good :).

Hopefully by then I'll have something that is a drop in replacement for
what it was before, at that point it'll be ready to merge and improve from
there in other pull requests.

I think the changelog will say something like:

In preparation for eventually stabilising the API a large number of
functions that were previously exposed have now been made private. No
function that was within the documentation has been removed. If there is a
function that you were using, and you want it to be exposed again please
let us know via the GitHub issues page [link] and we will be more than
happy to expose it also. Bear in mind however that the entire API that is
currently exposed will likely change before a 1.0.0 release.

... That last point "the entire API is likely to change before a 1.0.0
release" is something that we should have written on the front page.

SimonBiggs:
Documentation is now parity. Before merge I need to set up doctests within tox. We haven't been running doctests for a while. We have some docstrings which are not in the docs proper, but they need to also be accurate. So, I'll set up doctest with tox, run that, then fix any issues with that.

After that, we have a drop in replacement for every API we have provided within the docs, and any docstring within the library. All other functions have been made inaccessible.

SimonBiggs:
@Centrus007 alright, the tests pass. All APIs that appear within docstrings or the docs have been temporarily exposed via the top level `pymedphys` package.

Any part of the code that did not have testing could very well be broken... but I guess that teaches us to test. If we find something broken, we should probably write tests for it before fixing it.

My aim has been to rework the entire internals without actually changing how any of it behaves (except reducing the scope of the API to only that which is documented somewhere). Let me know if you find something that has changed.

Cheers :)

Simon

SimonBiggs:
@Centrus007 ready for review :).

SimonBiggs:
In saying what I said, @Centrus007, if you did want to do the DICOM object stuff first, as long as you don't make any new files potentially `git` will just handle it:

https://github.com/pymedphys/pymedphys/pull/181/commits/100f3dfe9447caa727337ada27a27bda20e6cb15

Up to you, see what you reckon :).

Centrus007:
I can't get the docs to build...:

image

Centrus007:
> In saying what I said, @Centrus007, if you did want to do the DICOM object stuff first, as long as you don't make any new files potentially git will just handle it:

Nah, let's get this in.

SimonBiggs:
Need to run `yarn` in top level directory. That will install all the
packages.

SimonBiggs:
See https://github.com/pymedphys/pymedphys/pull/181#issuecomment-485069039

SimonBiggs:
Working well enough?

Centrus007:
> Working well enough?

Perfectly. I'm sorry, I should've read more thoroughly.

azure-pipelines.yml
```diff
@@ -121,11 +134,11 @@ jobs:
...
-        mac-py36:
-          image_name: macos-10.13
-          test_title: macOS py36
-          toxenv: py36
-          python: 3.6
+        # mac-py36:
+        #   image_name: macos-10.13
+        #   test_title: macOS py36
+        #   toxenv: py36
+        #   python: 3.6
```
Centrus007: Why was this done? Anticipating removing support for python 3.6?

docs/user/library/dicom.rst
```diff
@@ -6,7 +6,7 @@ DICOM Module
 Summary
 *******
 
-.. automodule:: pymedphys.dicom
+.. automodule:: pymedphys_dicom.dicom
```
Centrus007: Lots of redundant labelling...

packages/pymedphys_dicom/src/pymedphys_dicom/_install_requires.py
```diff
@@ -0,0 +1,8 @@
+install_requires = [
+    'pymedphys_analysis',
+    'pymedphys_utilities',
+    'numpy < 1.16, >= 1.12',
```
Centrus007: How important is it that we keep these pins consistent across packages? If so, can we somehow apply these pins "DRYly"? Or perhaps include a test that every pinned external dependency has the same pins across all subpackages?

packages/pymedphys_labs/setup.py
```diff
@@ -0,0 +1,61 @@
...
+    globs = {}
+    execfile(filepath_in_package, globs)
+    variable_value = globs[variable]
+
+    return variable_value
+
+
+version = get_variable_from_file('_version.py', '__version__')
+install_requires = get_variable_from_file(
+    '_install_requires.py', 'install_requires')
```
Centrus007: Or have you already done this and I'm failing in my review logic?

setup.py
```diff
@@ -18,58 +18,67 @@ def read(*names, **kwargs):
...
+    package_dir={'': package_dir},
     include_package_data=True,
     package_data={'pymedphys': []},
     entry_points={
         'console_scripts': [
             'pymedphys=pymedphys.cli.main:pymedphys_cli'
         ],
     },
-    license='AGPLv3+',
+    license='AGPL-3.0-or-later',
```
Centrus007: Does the '+' in the old version signify "or later" or "with added terms"?

src/pymedphys/dicom.py
```diff
@@ -0,0 +1,8 @@
+from pymedphys_dicom.dicom import (
```
Centrus007: Nice to see removal of `import *`

Centrus007:
See my suggested wording changes [here](https://github.com/pymedphys/pymedphys/pull/181/commits/2e7bdad919ee6a7e80ce756c95e3ba17cd1509f7).

SimonBiggs:
> How important is it that we keep these pins consistent across packages?
If so, can we somehow apply these pins "DRYly"? Or perhaps include a test
that every pinned external dependency has the same pins across all
subpackages?

It's very important to be DRY for this and a range of other things. The aim
is to have these dependency pins templated and then propegated throughout
via a template. It will likely happen before every commit using a commit
hook. Or at a point similar to that...

SimonBiggs:
My plan was to build a dependency tree from the files themselves, have a
reference set of pins in one location, then using the dependency tree
propegate the dependencies to each library with pins that match the
reference.

azure-pipelines.yml
SimonBiggs: I actually just didn't want to deal with the `dataclasses` dependency at this point. We call pull `dataclasses` back in as a dependency if python==3.6 if we like. Or we can remove `dataclasses` from the code base if we don't need them. We never did actually use them in the end.

Either options will allow us to support Python 3.6 again...

docs/user/library/dicom.rst
SimonBiggs: That module likely won't be called dicom anymore. I envision there being a swathe of modules under dicom such as:

* pydicom_wrapper
* anonymise
* create
* rtstruct
* rtplan
* collection
* etc...

Just move it like that for now so that it was minimal changes for the split.

packages/pymedphys_dicom/src/pymedphys_dicom/_install_requires.py
SimonBiggs: It's very important to be DRY for this and a range of other things. The aim
is to have these dependency pins templated and then propegated throughout
via a template. It will likely happen before every commit using a commit
hook. Or at a point similar to that...

My plan was to build a dependency tree from the files themselves, have a
reference set of pins in one location, then using the dependency tree
propegate the dependencies to each library with pins that match the
reference.

packages/pymedphys_labs/setup.py
SimonBiggs: This is certainly prepping to make it DRY. But it is not there yet.

SimonBiggs: I figured we could have a period of time where it wasn't DRY, as it will take a bit of time to sort out the tools to DRY it all up. Bit of a chicken and the egg, need to have it in this format to know what tools are needed to DRY things.

setup.py
SimonBiggs: In symbolises "or later". This isn't a license change, just that there has been a consensus that the + was not sufficiently explicit within the community.

src/pymedphys/dicom.py
SimonBiggs: Yup, and we need to remove that also from all the levelling. With this new format each module at each level depth is going to be explicit about what API it is exposing. That will make versioning much clearer, allow us to keep a stable API, while still changing lots of things. ... them OOP folk and their encapsulation :). (only psudo encapsulation, because people can still dive in past the API, but we won't provide guarantees for that sort of thing).

Centrus007:
Okay, I think I'm abreast of it all! Pretty happy that things are all working fine.

I'd like to rework [`file-structure.rst`](https://github.com/pymedphys/pymedphys/blob/the-big-split/docs/developer/file-structure.rst) to reflect these changes. It will appear quite convoluted to a newcomer, I think. But we can probably address that in a separate PR :)

> My plan was to build a dependency tree from the files themselves, have a reference set of pins in one location, then using the dependency tree propegate the dependencies to each library with pins that match the reference.

I think that's a good plan.

---

I'm loving the documentation parity. I think the file structure works well and the subpackages are neatly arranged. It's just beautifully put together, really.

I haven't settled on any glaring issues with the subpackage names. Minor points:

- I'm not *immediately* convinced on the need for `wrappers` (or at least, that wrappers should be collected together like that, regardless of what they're "wrapping").
- I'd probably prefer a different word to `numerics`.
- I think `electronfactors` ought not to belong in a subpackage called `shapes`. I wouldn't look at shapes and think "I'm going to find output factor algorithms in there!" But maybe that's just me...
- Not super keen on `sql`. Maybe `databases`? Dunno why... probably being silly.

README.rst
```diff
@@ -57,6 +57,18 @@ can access the Documentation `here <https://pymedphys.com>`__.
 .. _`conda-forge`: https://anaconda.org/conda-forge/pymedphys
 
 
+PyMedPhys is currently within the ``beta`` stage of its lifecycle. It will
+stay in this stage until the version number leaves ``0.x.x`` and enters
+``1.x.x``. While PyMedPhys is in ``beta`` stage, **no API is guaranteed to be
+stable from one release to the next.** In fact, it is very likely that the
+entire API will change multiple times before a ``1.0.0`` release. In practice,
+this means that upgrading ``pymedphys`` to a new version will possibly break
+any code that was using the old version of pymedphys. We try to abreast of
```
SimonBiggs:
```suggestion
any code that was using the old version of pymedphys. We try to be abreast of
```

Centrus007:
> Any part of the code that did not have testing could very well be broken... but I guess that teaches us to test. If we find something broken, we should probably write tests for it before fixing it.

Agreed - not the worst way to kinda retrospectively 'force' ourselves to do it. :)

packages/pymedphys_labs/setup.py
Centrus007: Fair enough.

SimonBiggs:
> I'm loving the documentation parity. I think the file structure works well and the subpackages are neatly arranged. It's just beautifully put together, really.

Thanks Matt :)

SimonBiggs:
> I haven't settled on any glaring issues with the subpackage names. Minor points:
> 
> * I'm not _immediately_ convinced on the need for `wrappers` (or at least, that wrappers should be collected together like that, regardless of what they're "wrapping").
> * I'd probably prefer a different word to `numerics`.
> * I think `electronfactors` ought not to belong in a subpackage called `shapes`. I wouldn't look at `shapes` and think "I'm going to find output factor algorithms in there!" But maybe that's just me...
> * Not super keen on `sql`. Maybe `databases`? Dunno why... probably being silly.

I agree with `electronfactors`, the only dependency that was pulling it out of something like `pymedphys-analysis` was `shapely`. There might be other packages that need `shapely`, (like `logo`), so thought that was a reasonable grouping name. But I certainly wasn't convinced.

With regard to `pymedphys-numerics` I originally called it `base`, but then I thought `dicom` will also end up becoming a bit of a base also... It's defining feature is only depending on `numpy`, but once again also wasn't convinced on the name.

By the time I got to `pymedphys-wrappers` I was getting lazy :). Essentially their defining features was "they depend on a lot of packages combined and it would blow the external dependency tree out if they were included lower down the levels." ... they probably should be split into two separate packages and be named more sensibly.

Yup, I agree with `sql`, I like `pymedphys-databases`.

SimonBiggs:
Thanks Matt, you do the merge so that you're name is also on the commit. :).

SimonBiggs:
@Centrus007 merged commit 7bee41a into master  6 minutes ago

PR #190, opened 2019-04-22
[RFR] Fix rtd post split

SimonBiggs:
Woo hoo, got readthedocs to work:

https://readthedocs.org/projects/pymedphys/builds/8946448/

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,30 +3,11 @@
   "python.linting.mypyEnabled": true,
   "files.trimTrailingWhitespace": true,
   "editor.formatOnSave": true,
-  "editor.detectIndentation": false,
+  "editor.detectIndentation": true,
   "editor.insertSpaces": true,
-  "editor.tabSize": 2,
-  "[python]": {
-    "editor.tabSize": 4,
-    "editor.rulers": [
-      72,
-      79,
-      119
-    ]
-  },
-  "[yaml]": {
-    "editor.tabSize": 4
-  },
-  "[rst]": {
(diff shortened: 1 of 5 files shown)


.vscode/settings.json
```diff
@@ -3,30 +3,11 @@
...
-    "editor.rulers": [
-      79,
-      119
-    ]
-  }
+  "editor.rulers": [
+    72,
+    79,
+    119
+  ]
```
Centrus007: Was this intentionally committed?

SimonBiggs:
Ah yes, figured we could discuss it, but I think we can just have those
rulers everywhere.

And we I think letting VSCode automatically detect the indentation works
well enough...

SimonBiggs:
Here's the link to the docs now building on rtd:

https://pymedphys.com/en/fix-rtd-post-split/

.vscode/settings.json
SimonBiggs: Ah yes, figured we could discuss it, but I think we can just have those
rulers everywhere.

And we I think letting VSCode automatically detect the indentation works
well enough...

Centrus007: I personally found the 72 line a little annoying for `.md` and `.rst`...

Centrus007: Would something like:

```js
  "editor.rulers": [
    79,
    119
  ],
"[python]": {
    "editor.rulers": [
      72,
      79,
      119
    ]
  },
```

be alright?

SimonBiggs:
Yup certainly

PR #193, opened 2019-04-23
RFR: Add batch install files. Combine 'pylint' setup into 'formatting'. Improve nbstripout install status check.

### LINKED ISSUE #192: Developer setup guidelines no longer work after the big split (#171)
This little snippet no longer works:

```bash
λ conda install pymedphys --only-deps
...
λ pip install -e .
Obtaining file:///C:/Users/mjenni03/PythonProjects/pymedphys
Collecting pymedphys_analysis (from pymedphys==0.8.0.dev0)
  Could not find a version that satisfies the requirement pymedphys_analysis (from pymedphys==0.8.0.dev0) (from versions: )
No matching distribution found for pymedphys_analysis (from pymedphys==0.8.0.dev0)
```

Should we replace `pip install -e .` with `yarn`?

Also, I was thinking it might be nice to write up a simple batch script for users to run? I think that roughly the entire second half of the process could be batched...

Centrus007:
Resolves #192


scripts/developer/developer-install-with-env-and-extras.bat
```diff
@@ -0,0 +1,13 @@
...
+conda create -y --name pmp python=3.7 &&^
+conda activate pmp &&^
+conda install -y nbstripout &&^
+nbstripout --install &&^
+nbstripout --is-installed &&^
+conda install -y pymedphys --only-deps &&^
+yarn &&^
+pip install -e .[docs] &&^
+pip install -e .[testing] &&^
+pip install -e .[formatting]
```
SimonBiggs: What do you think about making these into [npm scripts](https://docs.npmjs.com/misc/scripts) that run with yarn in a cross-platform way?

setup.py
```diff
@@ -94,14 +94,16 @@ def read(*names, **kwargs):
             'deepdiff',
             'pytest',
             'pytest-cov',
+            'tox',
```
SimonBiggs: hmmm... this is going to be a bit odd here, `tox` itself uses this to install its testing requirements. `tox` probably shouldn't be installing `tox` everytime `tox` is called...

package.json
```diff
@@ -1,11 +1,10 @@
...
   "scripts": {
-    "install": "lerna bootstrap",
-    "postinstall": "pip install -e ."
-  },
-  "devDependencies": {
-    "lerna": "^3.13.3"
+    "install": "pip install -r requirements-development.txt",
+    "nbstripout:install": "conda install -y nbstripout; nbstripout --install",
+    "conda:install-deps": "conda install -y pymedphys --only-deps",
+    "conda:create-env": "conda create -y --name pmp python=3.7; conda activate pmp; yarn conda:installdeps; yarn nbstripout:install; yarn"
```
SimonBiggs: Here is an example of how npm scripts are made.

I haven't tested these ones, just did them quickly, but for example running the following should achieve similar to one of the bat files:

```bash
yarn conda:create-env
```

requirements-development.txt
```diff
@@ -0,0 +1,6 @@
+-r requirements-editable.txt
+-r requirements-docs.txt
+-r requirements-testing.txt
+-r requirements-linting.txt
+
+-e .
```
SimonBiggs: What do you think about leaning on requirement files? The job is then to automatically create these files as needed...

setup.py
Centrus007: We can remove. I only considered the manual use case (which I've only ever done once).

requirements-development.txt
Centrus007: Happily :)

package.json
Centrus007: Sure! :) Just never done it before. I assume I can call an npm scripts from a command prompt or Cmder?

SimonBiggs: yup :)

Just fixed the scripts above (had to get yarn installed at work and test it). Now `yarn conda:create-env` will create the conda environment, and `yarn conda:delete-env` will delete it.

SimonBiggs:
Okay, that's all from me for now. Would you be able to see what you think about a few of those changes?

package.json
```diff
@@ -1,11 +1,10 @@
 {
   "private": true,
   "scripts": {
-    "install": "lerna bootstrap",
-    "postinstall": "pip install -e ."
-  },
-  "devDependencies": {
-    "lerna": "^3.13.3"
+    "install": "pip install -r requirements-development.txt",
```
SimonBiggs: Now `yarn` simply runs `pip install -r requirements-development.txt`.

setup.py
SimonBiggs:
```suggestion

```

docs/developer/contributing.rst
```diff
@@ -197,30 +200,28 @@ contain private and/or sensitive information.
...
-    conda create --name pmp python=3.7 # Optional.
-    conda activate pmp # Optional
-    conda install nbstripout
-    nbstripout --install
-    nbstripout --status
+    λ conda create --name pmp python=3.7 # Optional.
+    λ conda activate pmp # Optional
+    λ conda install nbstripout
+    λ nbstripout --install
+    λ nbstripout --is-installed && echo Success!
```
SimonBiggs: I think we should remove the `λ` here, what do you think?

docs/developer/contributing.rst
```diff
@@ -231,8 +232,9 @@ conda. With cmder open in the pymedphys directory, run:
 .. code:: bash
 
     conda install pymedphys --only-deps
```
SimonBiggs: This line here will also break after we make our release. `pymedphys` will depend on our custom packages only. Maybe what we could do is bundle this into yarn so that all we need to do is run `yarn` and it handles the rest.

docs/developer/contributing.rst
SimonBiggs: I see what you're doing now. Making sure success! is different from the commands.

Centrus007: I actually put it in deliberately to distinguish between commands typed by the user (text after a lambda) and outputs from commands (text not after a lambda).

I do think if we remove the lambda, we need some way of distinguishing between the two text types. We don't want people typing "Success!" and wondering why it's not a recognized command...

Centrus007: > I see what you're doing now. Making sure success! is different from the commands.
Yep! haha. Commented at the same time

Centrus007:
I was hoping for a single script that would:

- remove env 'pmp' if it exists (i.e. a cleanout), continue if it doesn't, or continue once it is successfully removed
- create env
- install everything (including docs, testing, linting, etc.)

all in one single command.

Then I realised variations would be required for those not using an env and perhaps also for those not needing the docs, linting, testing packages (if those people exist?). Hence the four batch files.

Is there a single command in your npm stuff that does this?

package.json
```diff
@@ -1,11 +1,31 @@
...
+  "cross-os": {
+    "dependencies": {
+      "win32": "yarn dependencies:darwin:win32",
+      "darwin": "yarn dependencies:darwin:win32",
+      "linux": "yarn dependencies:linux"
+    }
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "nbstripout --is-installed"
```
SimonBiggs: So once a user runs `yarn bootstrap` they will have a pre-commit hook installed which forcibly checks if `nbstripout` is installed appropriately. ... potentially that means we can remove bits from the contributor docs. What do you think?

docs/developer/contributing.rst
```diff
@@ -230,9 +231,9 @@ conda. With cmder open in the pymedphys directory, run:
 
 .. code:: bash
 
-    conda install pymedphys --only-deps
-    pip install -e .
+    yarn bootstrap
```
SimonBiggs: See the `bootstrap` command within `package.json`, this now handles dependencies, nbstripout, and a development install, as well as adding a hook so that before every commit it is confirmed that nbstripout is installed.

package.json
SimonBiggs: image

docs/developer/contributing.rst
SimonBiggs: Potentially we can remove this section of the docs however... I've collapsed this installation into `yarn bootstrap` combined with a pre-commit hook that checks that `nbstripout` is installed. Let me know what you think.

package.json
Centrus007: Cool! Yeah, I we could omit *nbstripout*, I think.

Once we're comfortable with the commands, I'd almost introduce the batch file and suggest running that, but retain the existing recommendations (adapted to whatever changes we make) for understanding?

SimonBiggs:
> I was hoping for a single script that would:
> 
> * remove env 'pmp' if it exists (i.e. a cleanout), continue if it doesn't, or continue once it is successfully removed
> * create env
> * install everything (including docs, testing, linting, etc.)
> 
> all in one single command.
> 
> Then I realised variations would be required for those not using an env and perhaps also for those not needing the docs, linting, testing packages (if those people exist?). Hence the four batch files.
> 
> Is there a single command in your npm stuff that does this?

There is now :)

```
yarn conda:reinstall
```

SimonBiggs:
So the process should be, if they want an environment based install then they should run `yarn conda:create`, if they want a non-environment install then instead run `yarn bootstrap`.

... I am leaning towards recommending just one way. How about, instead of making it optional, we simply say "run `yarn conda:create`".

Centrus007:
> I am leaning towards recommending just one way. How about, instead of making it optional, we simply say "run yarn conda:create".

Well, I'd love that. But only if you're sure...

SimonBiggs:
Yup, it's our job to provide a recommended setup, and there should only be one recommended setup. Otherwise we are asking newcomers to make decisions they're not yet equipped to make.

SimonBiggs:
Alright, time for me to drive home :) Hopefully this branch passes now :)

docs/developer/contributing.rst
Centrus007: Agree. Let's remove it.

SimonBiggs:
> Well, I'd love that. But only if you're sure...

I'm sure, environments are a good idea. We can make them easier to use if need be.

docs/developer/contributing.rst
SimonBiggs: Maybe instead of removing, moving it to another doc where we describe what is happening under the hood when `yarn conda:create` is run?

package.json
SimonBiggs: > but retain the existing recommendations (adapted to whatever changes we make) for understanding?

Potentially retain them in a separate document that is designed to describe "what magic just happened" when they run `yarn conda:create`. We can then link to that document right before asking them to ran `yarn conda:create` so those who want to do it manually, or learn what is about to happen to their machine can do so...

package.json
```diff
@@ -1,11 +1,34 @@
...
   "private": true,
   "scripts": {
-    "install": "lerna bootstrap",
-    "postinstall": "pip install -e ."
+    "bootstrap": "yarn && yarn conda:forge && yarn dependencies && yarn nbstripout:install && yarn install:dev",
+    "install:dev": "pip install -r requirements-development.txt",
+    "install:prod": "pip install -r requirements.txt",
+    "nbstripout:install": "pip install nbstripout && nbstripout --install",
+    "dependencies": "cross-os dependencies",
+    "dependencies:darwin:win32": "conda install -y --file requirements-dependencies-only.txt",
```
SimonBiggs: A python 3.7 flag here might fix your issue... Or maybe one within the requirements dependencies .txt file?

SimonBiggs: Nope, it can't sit in the requirements file... we could write the python version required in here too... Again... Hmmmm, this WET first DRY later thing is bugging me ...

PR #195, opened 2019-04-23
[RFR] Reworking package names

SimonBiggs:


SimonBiggs:
@Centrus007 this is ready for review :) (hopefully it passes online)

SimonBiggs:
I changed `wrappers` to `toolbox`. I figured we might want a `toolbox` package anyway. What's in it however can be up for question at some point.

package.json
```diff
@@ -1,7 +1,9 @@
 {
+  "name": "pymedphys",
   "private": true,
   "scripts": {
     "bootstrap": "yarn && yarn conda:forge && yarn dependencies && yarn nbstripout:install && yarn install:dev",
+    "lerna:bootstrap": "lerna bootstrap && pip install -e .",
```
SimonBiggs: I added this back in, just so that we can use `lerna` if need be. We will need to zero in on one tool that works well, but I found `lerna` made the installation quite effortless.

So essentially `yarn lerna:boostrap` will do what just `yarn` used to do before I attempted to swap over to requirement files.

SimonBiggs: @Centrus007

SimonBiggs: @Centrus007

PR #221, opened 2019-04-29
RFR: DICOM coordinates refactor

Centrus007:
- Combine coordinate extraction functions into single function: `xyz_axes_from_dataset()`
   - rename to "axes" as this is the best word I can think of for the particular arrays that the function returns. Calling them coordinates is a little confusing, since I think it is natural to assume that "coordinates" would instead refer to a grid of coordinate triplets that matches the size of, and whose voxel indices correspond to, the orignal dose grid.
- Move coordinate related functions out of `dose` and into a new `coords` file in level 1 of the `dicom` module.
  - Turns out the coordinate functions also work perfectly well for DICOM images (esp 'CT').
- Remove deprecated functions.
  - `load_xyz_from_dicom()` has been sitting around awaiting execution for a while now. Let's just rip the bandaid off? I replaced all uses of it with the new `xyz_from_dataset()`. All tests pass, so...?


Centrus007:
Finally got the tests to pass. I was quite impressed by the docs test suite actually - it actually ran through your gamma example in the docs to ensure it ran without error. This brought to my attention an error in my changes that was not picked up in the standard test suite...

changelog.md
```diff
@@ -49,6 +49,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   `mudensity` module.
 - `compare_mosaiq_fields` moved from the `msq` module into the `plancompare`
   module.
+- The DICOM coordinate extraction functions - `extract_dicom_patient_xyz()`,
+  `extract_iec_patient_xyz()` and `extract_iec_fixed_xyz()` - have been
+  combined into a single function called `xyz_from_dataset()`. The x, y, z axes
+  can still be returned in either the DICOM, IEC FIXED or IEX PATIENT
+  coordinate systems by passing `coord_system='D'` (`F` or `P`) to the
```
SimonBiggs: So that people reading this API in the wild know what the code is doing I think these parameters values should be as descriptive as is normally expected from function names. See:

https://docs.scipy.org/doc/scipy-0.18.1/reference/optimize.html

Parameter names such as coord_system='dicom', 'iec_fixed', 'iec_patient' makes sense...

packages/pymedphys_analysis/src/pymedphys_analysis/gamma/_level1/gammafilter.py
```diff
@@ -44,17 +46,17 @@ def calculate_pass_rate(gamma_array):
...
 
-def gamma_filter_brute_force(coords_reference, dose_reference,
-                             coords_evaluation, dose_evaluation,
+def gamma_filter_brute_force(axes_reference, dose_reference,
+                             axes_evaluation, dose_evaluation,
                              distance_mm_threshold, dose_threshold,
                              lower_dose_cutoff=0, **kwargs):
 
-    xx_ref, yy_ref, zz_ref = np.meshgrid(*coords_reference, indexing='ij')
+    coords = coords_from_xyz_axes(axes_reference)
```
SimonBiggs: I don't believe I ever wrote tests for gamma filter. Did I? I'm just doing this on my mobile... Is gamma filter still accurate with the axis labels being placed on different array dimensions?

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level1/coords.py
```diff
@@ -0,0 +1,195 @@
...
+
+    Returns
+    -------
+    coords : ndarray
+        An array containing three grids consisting of the `x`, 'y` and
+        `z` coordinates of the corresponding grid (e.g. DICOM dataset's
+        pixel array) from which the original axes were extracted.
+    """
+    ZZ, YY, XX = np.meshgrid(
+        xyz_axes[2], xyz_axes[1], xyz_axes[0], indexing='ij')
```
SimonBiggs: I seem to remember this order here is important as in it decides which axis corresponds to which coordinate value. What's the responding for the reverse order? Not against it, just interested why the choice.

changelog.md
SimonBiggs: Just had a look at the function and I see you do have the option for more descriptive naming :)

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level1/coords.py
```diff
@@ -0,0 +1,195 @@
...
+        are:
+
+        'DICOM' or 'D':
+            Return axes in the DICOM coordinate system.
+
+        'PATIENT', 'IEC PATIENT' or 'P':
+            Return axes in the IEC patient coordinate system.
+
+        'FIXED', 'IEC FIXED' or 'F':
+            Return axes in the IEC fixed coordinate system.
```
SimonBiggs: Not opposed here, but reasoning for all caps?  Maybe there's value in running an "upper" on user input so that lower case works also? Not sure. Up to you.

Nevermind, I missed it, you did use an upper :).

changelog.md
Centrus007: Yeah, I was just a little lazy with the changelog. I'll update to match the docstring better :)

packages/pymedphys_analysis/src/pymedphys_analysis/gamma/_level1/gammafilter.py
Centrus007: I should've been more careful here. Let's remove that change for now.

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level1/coords.py
Centrus007: Hehe. Would you prefer lowercase in the docstring?

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level1/coords.py
Centrus007: This is the order required for the following to be true:

The coordinate value at `coords[axis, i, j, k]` corresponds to the dose value at `dose[i, j, k]`.

When you read `grid = ds.pixel_array`, the index order is `grid[slice, row, column]`, which corresponds to `grid[z, y, x]` in the DICOM coordinate system.

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level1/coords.py
SimonBiggs: :) I think if they're acronyms they can get away with all caps, such as IEC and DICOM but everything else should probably be lower within the docstring.

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level1/coords.py
Centrus007: I think this was all relatively subconscious, but in your older functions (where you were extracting PDDs, etc.) you were working in the IEC fixed coordinate system. In the IEC fixed coordinate system, `grid[slice, row, column]` corresponds to `grid[y, z, x]`. You also seemed to prefer the order `grid[row, column, slice]` (corresponding to  `grid[y, x, z]`), hence the transpose [here](https://github.com/pymedphys/pymedphys/blob/4d200898f678d2306c49b517ab9d8700270063cc/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/dose.py#L75).

Finally, when you were working with `coords` (now called "axes") and `dose` together, you tended to reorder the coords tuple to `(y, x, z)` to match your adjusted dose grid. E.g. [here](https://github.com/pymedphys/pymedphys/blob/4d200898f678d2306c49b517ab9d8700270063cc/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/dose.py#L425).

Centrus007: Later we agreed to keep the dose grid unaltered (things got messy by both transposing the grid, AND reordered coords, plus other aesthetic reasons besides). So we wrote some coords extraction functions for different coordinate systems to handle cases where they might be more intuitive (like water tank applications).

SimonBiggs:
> * Combine coordinate extraction functions into single function: `xyz_axes_from_dataset()`
>   
>   * rename to "axes" as this is the best word I can think of for the particular arrays that the function returns. Calling them coordinates is a little confusing, since I think it is natural to assume that "coordinates" would instead refer to a grid of coordinate triplets that matches the size of, and whose voxel indices correspond to, the orignal dose grid.

Yup, good choice there I think.

> * Move coordinate related functions out of `dose` and into a new `coords` file in level 1 of the `dicom` module.
>   
>   * Turns out the coordinate functions also work perfectly well for DICOM images (esp 'CT').

Yup, great move.

> * Remove deprecated functions.
>  
>   * `load_xyz_from_dicom()` has been sitting around awaiting execution for a while now. Let's just rip the bandaid off? I replaced all uses of it with the new `xyz_from_dataset()`. All tests pass, so...?

This worries me a little. It's my fault, it's a piece of code that was used to mass compare profiles between DICOM files and water tank scans. It'd be nice if that code would continue to work. But really, what I need to do is pull that code into the repo and have some tests for it (even something really basic) so that it is clear what it needs to achieve.

Let me do that, I might get a chance tonight...

SimonBiggs:
> > * Remove deprecated functions.
> >   
> >   * `load_xyz_from_dicom()` has been sitting around awaiting execution for a while now. Let's just rip the bandaid off? I replaced all uses of it with the new `xyz_from_dataset()`. All tests pass, so...?
> 
> This worries me a little. It's my fault, it's a piece of code that was used to mass compare profiles between DICOM files and water tank scans. It'd be nice if that code would continue to work. But really, what I need to do is pull that code into the repo and have some tests for it (even something really basic) so that it is clear what it needs to achieve.
> 
> Let me do that, I might get a chance tonight...

So I've had a bit more of a chance to think about it. Let's rip them off. Like a band-aid. If I need to use the old functions I can pin PyMedPhys to < 0.8.0 for that project. But in reality I will likely spend the time when it is needed again to clean the code up and submit it to PyMedPhys using the new functions / dicom object methods.

packages/pymedphys_analysis/src/pymedphys_analysis/gamma/_level1/gammafilter.py
SimonBiggs: Kk, so this probably the only change I need and then I'm all good for this to merge. :)

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level1/coords.py
Centrus007: `y, x, z` corresponds to `i, j, k` index order, which is preferred for linear algebra. Consequently, it appears to have made it into standard programming practice. It's just that for our application, axis order is probably more important than index order (when dealing with coordinates, physicists are going to much prefer xyz, ijk is unlikely to enter their minds). Moreover, I'd say medical physicists are probably more used to a kji index order *anyway*, since we're used to CT/volumetric applications where we typically *select slice first*.

SimonBiggs: I'm all okay with this :). It'll just take me a bit to get used to I think.

Centrus007:
> I was quite impressed by the docs test suite actually - it actually ran through your gamma example in the docs to ensure it ran without error. This brought to my attention an error in my changes that was not picked up in the standard test suite...

Just to elaborate on this, I rally, really think this is great. Once reason I've been reluctant to build too much in the way of examples is that, while we're still making massive API changes, it'd be hard to ensure example code stays correct (case in point: #192). The more examples we put in notebooks that are run by the test suite, the more likely we'll catch obsolescence in the docs caused by API changes.

Centrus007:
> This worries me a little. It's my fault, it's a piece of code that was used to mass compare profiles between DICOM files and water tank scans. It'd be nice if that code would continue to work. But really, what I need to do is pull that code into the repo and have some tests for it (even something really basic) so that it is clear what it needs to achieve.

I don't think it matters too much for `load_xyz_from_dicom()`. A straight switch to `xyz_from_dataset()` would be fine - there were no convention changes between those two functions. `xyz_from_dataset()` is simply a generalised version of `load_xyz_from_dicom()` that handles all patient orientations (and now multiple coordinate systems).

As for `load_dose_from_dicom()`, fair enough. The reshape is the only issue. We don't yet have a replacement for it though, so I didn't propose removing it.

SimonBiggs:
> > I was quite impressed by the docs test suite actually - it actually ran through your gamma example in the docs to ensure it ran without error. This brought to my attention an error in my changes that was not picked up in the standard test suite...
> 
> Just to elaborate on this, I really, really think this is great. Once reason I've been reluctant to build too much in the way of examples is that, while we're still making massive API changes, it'd be hard to ensure example code stays correct (case in point: #192). The more examples we put in notebooks that are run by the test suite, the more likely we'll catch obsolescence in the docs caused by API changes.

Yup, makes sense. It is very nice for users to have a self contained piece of code, which can be run online if needed, and is guaranteed to work with the documented version of PyMedPhys since the docs won't build if it has an error in it.

SimonBiggs:
> > This worries me a little. It's my fault, it's a piece of code that was used to mass compare profiles between DICOM files and water tank scans. It'd be nice if that code would continue to work. But really, what I need to do is pull that code into the repo and have some tests for it (even something really basic) so that it is clear what it needs to achieve.
> 
> I don't think it matters too much for `load_xyz_from_dicom()`. A straight switch to `xyz_from_dataset()` would be fine - there were no convention changes between those two functions. `xyz_from_dataset()` is simply a generalised version of `load_xyz_from_dicom()` that handles all patient orientations (and now multiple coordinate systems).
> 
> As for `load_dose_from_dicom()`, fair enough. The reshape is the only issue. We don't yet have a replacement for it though, so I didn't propose removing it.

Yup, sounds good to me :).

SimonBiggs:
Are you happy for me to merge?

Centrus007:
> Are you happy for me to merge?

Please do!

Centrus007:
> Thanks Matt for this. Awesome :).

No worries :D

SimonBiggs:
Heya @fbockelmann,

Here is where the changes to `load_xyz_from_dicom` were made.

Also see the documentation of the change at:

https://pymedphys.com/getting-started/changelog.html#breaking-changes

image

PR #222, opened 2019-04-29
[RFR] Dicom anonymisation CLI (FINALLY)

Centrus007:


Centrus007:
Seems to work well! Will add testing though.

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
```diff
@@ -247,7 +248,7 @@ def anonymise_file(
         This ensures that the filename contains no identifying
         information. If set to `False`, `anonymise_file()` simply
         appends "_Anonymised" to the original DICOM filename. Defaults
-        to True.
+        to `True`.
```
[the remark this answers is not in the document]
Centrus007: > How often do DICOM files have identifying information in their file name? Part of me thinks this is overstepping the scope of the function. But it's not something I'll hold back a merge over.

Varian's preferred format is:
`"<Modality>.<PatientID>.<optional extras - like plan name>.dcm"`

So all the time, if you count `PatientID`...

> Then again, that file name format is quite neat. You've convinced me :)... And you did without even speaking :).

Hahaha, I just mimicked RaySearch and Philips' (and apparently everyone except Varian's) convention :)

> For a single file (as opposed to a directory) the API for these sorts of things usually are input file, output file.

Sure. Where in particular do you want that reflected though? Just in the CLI functions? Because I really don't like variable names like `input_file` in the backend - it's not clear that it's a filepath, file handler/pointer, etc. It's fine in the CLI, since the user kinda *has to* supply a filepath...

> Seeing that written, wouldn't it be really nice of we did have a top level, plain, Dicom object that was able to appropriately mixin the methods of various Dicom types depending on what the type was of the DICOM file stored within the file

It certainly would :)

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
```diff
@@ -392,20 +393,62 @@ def anonymise_directory(
             remove_file(dicom_filepath)
 
 
-# def anonymise_files_cli(args):
+def anonymise_directory_cli(args):
+    if args.delete_unknown_tags:
```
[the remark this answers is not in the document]
Centrus007: I think that is a far better idea. :)

packages/pymedphys_dicom/tests/test_dicom_anonymise.py
```diff
@@ -19,33 +25,102 @@
...
+
+
+def test_anonymise_and_is_anonymised_dataset():
+
+    # Create dict with one instance of every identifying keyword and
+    # run basic anonymisation tests
+    non_anon_dict = dict.fromkeys(IDENTIFYING_KEYWORDS)
+
+    for key in non_anon_dict:
+        non_anon_dict[key] = _get_non_anonymous_replacement_value(key)
```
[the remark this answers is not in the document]
Centrus007: :)

packages/pymedphys_dicom/tests/test_dicom_anonymise.py
```diff
@@ -19,33 +25,102 @@
...
+    assert is_anonymised_dataset(ds_anon_blank)
+
+    # Test handling of unknown tags by removing PatientName from
+    # baseline dict
+    patient_name_tag = tag_for_keyword('PatientName')
+    patient_name = BaselineDicomDictionary.pop(patient_name_tag)
+
+    with pytest.raises(ValueError) as e_info:
+        anonymise_dataset(ds_non_anon)
+    assert str(e_info).count("At least one of the non-private tags "
```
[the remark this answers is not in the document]
Centrus007: Probably. I do like to be as narrow in error checking as possible though. Some *very* quirky things can happen (as we've seen in `pydicom`) that, IMHO, make it advisable to be absolutely sure to catch only the error you're intending to catch...

packages/pymedphys_dicom/tests/test_dicom_anonymise.py
```diff
@@ -19,33 +25,102 @@
...
+
+    # Test blank anonymisation
+    assert not is_anonymised_dataset(ds_non_anon)  # Sanity check
+    ds_anon_blank = anonymise_dataset(ds_non_anon, replace_values=False)
+    assert is_anonymised_dataset(ds_anon_blank)
+
+    # Test handling of unknown tags by removing PatientName from
+    # baseline dict
+    patient_name_tag = tag_for_keyword('PatientName')
+    patient_name = BaselineDicomDictionary.pop(patient_name_tag)
```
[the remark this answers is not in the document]
Centrus007: > How is this working here? ... I guess that does work doesn't it... It makes my skin crawl a little editing an imported dict like that... It makes me think we should use a tuple instead of a dict for the BaselineDicom... But maybe we are all "responsible adults" and letting people do this is just fine... ... It'll only be used when trying to break things...

Well, I guess we'll want an easy way to update the dict, which at the moment is a straight copy of `pydicom`'s dict, IIRC? I don't think it's *too* much of a risk, but I'm happy to consider making it immutable :)

> Maybe do the edit inside a try, finally pattern? So if this one fails we don't get really confused if we see other tests in other modules failing one day?

Yes, good idea.

packages/pymedphys_dicom/tests/test_dicom_anonymise.py
```diff
@@ -19,33 +25,102 @@
...
+    assert str(e_info).count("At least one of the non-private tags "
+                             "within your DICOM file is not within "
+                             "PyMedPhys's copy of the DICOM dictionary.")
+
+    ds_anon_delete_unknown = anonymise_dataset(ds_non_anon,
+                                               delete_unknown_tags=True)
+    assert is_anonymised_dataset(ds_anon_delete_unknown)
+    with pytest.raises(AttributeError) as e_info:
+        ds_anon_delete_unknown.PatientName
+    assert str(e_info).count("'Dataset' object has no attribute "
```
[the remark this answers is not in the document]
Centrus007: See my other related comment

packages/pymedphys_dicom/tests/test_dicom_anonymise.py
```diff
@@ -19,33 +25,102 @@
...
+    ds_anon_ignore_unknown = anonymise_dataset(ds_non_anon,
+                                               delete_unknown_tags=False)
+    assert is_anonymised_dataset(ds_anon_ignore_unknown)
+    assert patient_name_tag in ds_anon_ignore_unknown
+
+    patient_name = BaselineDicomDictionary.setdefault(patient_name_tag,
+                                                      patient_name)
+
+    # Test copy_dataset=False:
+    ds_anon_uncopied = anonymise_dataset(ds_non_anon, copy_dataset=False)
```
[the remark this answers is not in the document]
Centrus007: I agree. Maybe a True/False indicating success? Bit superflous though...

src/pymedphys/cli/dicom.py
```diff
@@ -94,3 +98,64 @@ def dicom_structure_name_RED_adjust(dicom_subparsers):
     parser.add_argument('output_file', type=str, help='output_file')
 
     parser.set_defaults(func=adjust_RED_by_structure_name_cli)
+
+
+def dicom_anonymise_directory(dicom_subparsers):
+    parser = dicom_subparsers.add_parser(
+        'anonymise-directory',
```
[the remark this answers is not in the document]
Centrus007: Agreed.

packages/pymedphys_dicom/tests/test_dicom_anonymise.py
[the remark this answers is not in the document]
SimonBiggs: Fair enough. If we find our tests begin to get brittle (as in they're giving off false negatives too often) we should revisit this practice I think.

packages/pymedphys_dicom/tests/test_dicom_anonymise.py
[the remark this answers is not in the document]
SimonBiggs: Yeah, I'd rather it return nothing, that way if someone tries to pass it something an error is raised? Is that how it works if there is no return? ... Or does it just return None... I can't remember.

SimonBiggs:
This is quite nice :) Thanks Matt.

packages/pymedphys_dicom/tests/test_dicom_anonymise.py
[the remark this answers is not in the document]
SimonBiggs: I think this can fall under the "we are all responsible adults" clause:

https://docs.python-guide.org/writing/style/#we-are-all-responsible-users

It can be left as is. It's not our job to know better than the user's of the code... If a piece of code wants to overwrite that library (maybe to write a test? Who knows?) We shouldn't stop them, no matter how much it makes my skin crawl.

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
```diff
@@ -404,14 +404,29 @@ def anonymise_directory_cli(args):
...
+            dicom_filepath=args.input_path,
+            delete_original_file=args.delete_original_files,
+            anonymise_filename=not args.preserve_filenames,
+            replace_values=not args.clear_values,
+            keywords_to_leave_unchanged=keywords_to_leave_unchanged,
+            delete_private_tags=not args.keep_private_tags,
+            delete_unknown_tags=handle_unknown_tags)
+
+    elif isdir(args.input_path):
+        anonymise_directory(
```
SimonBiggs: Very nice :)

src/pymedphys/cli/dicom.py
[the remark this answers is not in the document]
Centrus007: > Potentially here there is value on having an optional parameter output which can be the directory name in which to place the new files in, or the new name of the anonymised file.

I'll worry about those features in another pull request, I think.

src/pymedphys/cli/dicom.py
```diff
@@ -81,16 +84,81 @@ def dicom_adjust_rel_elec_density(dicom_subparsers):
...
+              "supplied, all DICOM files within the directory and its "
+              "subdirectories will be anonymised"))
+
+    parser.add_argument(
+        '-d', '--delete_original_files',
+        action='store_true',
+        help=("Use this flag to delete the original, non-anonymised "
+              "files in the processed directory. The original files "
+              "will only be deleted if anonymisation completes "
+              "successfully."))
```
SimonBiggs: I think it's worth considering how to fail gracefully. For example if the directory is half done then there are a heap of files left in the directory already anonymised, and a heap not anonymised. Rerunning the code in its current form I believe will then try and anonymise everything, including those that have been anonymised already.

Maybe this isn't a problem... But maybe it's worth considering.

src/pymedphys/cli/dicom.py
```diff
@@ -37,14 +39,15 @@ def dicom_cli(subparsers):
         help='A toolbox for the manipulation of DICOM files.')
     dicom_subparsers = dicom_parser.add_subparsers(dest='dicom')
 
-    dicom_adjust_machine_name(dicom_subparsers)
-    dicom_adjust_rel_elec_density(dicom_subparsers)
-    dicom_structure_name_RED_adjust(dicom_subparsers)
+    anonymise(dicom_subparsers)
```
SimonBiggs: Yup, this certainly be the first API seen.

Does it get crowded within the DICOM CLI docs? Should it get it's own page? What do you think?

src/pymedphys/cli/dicom.py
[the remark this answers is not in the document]
SimonBiggs: Yup, sounds good to me.

packages/pymedphys_dicom/tests/test_dicom_anonymise.py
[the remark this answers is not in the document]
Centrus007: All python "procedures" (functions that return nothing) return `None`. See 7.1. Expression statements [here](https://docs.python.org/3/reference/simple_stmts.html?highlight=return%20none)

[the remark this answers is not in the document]
SimonBiggs: Yeah, after I wrote that, I was pretty sure it was None. I think we should simply return nothing, it symbolises to the user that they should just use the variable they passed to the function. Might remind the user in a way that they asked it to be done with no copy / inplace.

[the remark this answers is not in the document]
SimonBiggs: You don't have to explicitly return None, we just don't write a return statement.

src/pymedphys/cli/dicom.py
Centrus007: Hard to say. I do think the formatting of the CLI docs is pretty horrible. The heading sizes are all off. And it'd be nice to space out the sub commands and for them to have links in the side bar...

SimonBiggs: We do have the option here to customise sphinx argparse. We can, for example, fork the project, and make the customisations we need.

SimonBiggs:
Let me know when you want this reviewed for merge, it seems to be getting pretty close...

src/pymedphys/cli/dicom.py
Centrus007: See latest commit :)

Centrus007:
> Let me know when you want this reviewed for merge, it seems to be getting pretty close...

Code wise, yes. Still pretty untested though ... :/

SimonBiggs:
> > Let me know when you want this reviewed for merge, it seems to be getting pretty close...
> 
> Code wise, yes. Still pretty untested though ... :/

Fair point yes.

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
```diff
@@ -215,7 +215,10 @@ def anonymise_dataset(
 
     ds_anon = _anonymise_tags(ds_anon, keywords_to_anonymise, replace_values)
 
-    return ds_anon
+    if copy_dataset:
+        return ds_anon
+    else:
+        return None
```
SimonBiggs: I don't think this return None is needed. I think the function should just be left to finish...

SimonBiggs:
Once we've got this sufficiently tested how about we fast track a webapp for this one?

SimonBiggs:
There's value in putting the DICOM files that are created all within a scratch directory and including the scratch directory within a git ignore in the directory above. That way, these testing artefacts won't ever be commited. Testing artefacts should always be gitignored..

Centrus007:
> There's value in putting the DICOM files that are created all within a scratch directory and including the scratch directory within a git ignore in the directory above. That way, these testing artefacts won't ever be commited. Testing artefacts should always be gitignored..

Yeah, that would've been a good idea. In this case, I was in quite a rush (Lauren was waiting at the car), so I literally committed and pushed and immediately noticed I'd committed unwanted files... :(

SimonBiggs:
Hehe :) it's okay :) Hope you didn't make Lauren late :)

Centrus007:
So I think this is probably ready... admittedly I haven't actually built tests for the CLI, but everything underlying it is pretty solidly tested now. Not completely perfect, but pretty good, I think.

What do you think?

Centrus007:
That latest force push cleaned out the dcm files. Pretty well getting the hang of that, I think!

SimonBiggs:
Git is a seriously amazing tool :)

SimonBiggs:
> So I think this is probably ready... admittedly I haven't actually built tests for the CLI, but everything underlying it is pretty solidly tested now. Not completely perfect, but pretty good, I think.
> 
> What do you think?

I think we do need some tests of the CLI, but I understand if you'd like to merge this in and leave that for another day.

Centrus007:
> I think we do need some tests of the CLI, but I understand if you'd like to merge this in and leave that for another day.

Yeah, on second thought, me too. On it!

Centrus007:
How about now? :)

src/pymedphys/cli/dicom.py
Centrus007: I think it's worthwhile. One day, hehe.

Centrus007:
> Once we've got this sufficiently tested how about we fast track a webapp for this one? I'd be happy to put in place the framework.

Yes, please!

packages/pymedphys_dicom/tests/test_anonymise.py
```diff
@@ -257,6 +257,66 @@ def test_anonymise_cli():
...
+    assert not is_anonymised_file(test_filepath)
+    assert not exists(test_anon_filepath)
+    anon_file_keep_command = ('pymedphys dicom anonymise'.split()
+                              + [test_filepath]
+                              + '-k PatientName'.split())
+    try:
+        subprocess.check_call(anon_file_keep_command)
+        assert not is_anonymised_file(test_anon_filepath)
+        ds = pydicom.dcmread(test_anon_filepath)
+        ds.PatientName = "Anonymous"
```
SimonBiggs: Not for this pull request, but I think there would be value in learning how to use https://pydicom.github.io/pydicom/stable/ref_guide.html#pydicom.filereader.read_partial with the `stop_when` parameter. I think once we learn how to use that we could as a whole speed up tests that load DICOM files.

packages/pymedphys_dicom/tests/test_anonymise.py
```diff
@@ -257,6 +257,66 @@ def test_anonymise_cli():
...
+        subprocess.check_call(anon_file_private_command)
+        assert not is_anonymised_file(test_anon_filepath)
+        assert is_anonymised_file(test_anon_filepath,
+                                  ignore_private_tags=True)
+        assert exists(test_filepath)
+    finally:
+        remove_file(test_anon_filepath)
+
+    # TODO: File anonymisation - unknown tag handling
+    # # Calling a subprocess reloads BaselineDicomDictionary...
```
SimonBiggs: Makes sense.

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
```diff
@@ -109,12 +110,24 @@
...
                                        'PN': "Anonymous",
                                        'SH': "Anonymous",
-                                       'SQ': ["Anonymous"],
+                                       'SQ': [Dataset()],
                                        'ST': "Anonymous",
                                        'TM': "000900.000000",
                                        'UI': "12345678"}
 
 
+def anonymised_dicom_filepath(filepath, ds=None, preserve_original=False):
```
SimonBiggs: What's this `ds` for here?

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
```diff
@@ -109,12 +110,24 @@
...
+    if preserve_original:
+        basename_anon = "{}_Anonymised.dcm".format('.'.join(
+            basename(filepath).split('.')[:-1]))
+    else:
+        mode_prefix = DICOM_SOP_CLASS_NAMES_MODE_PREFIXES[ds.SOPClassUID.name]
```
SimonBiggs: Will `ds.SOPClassUID` fail here if no `ds` is passed?

SimonBiggs: Ahh, I see what's going on here. I think this is clearly two functions squished awkwardly into one. For example, one of them would be:

```python
def create_filename_from_dataset(dataset):
    mode_prefix = DICOM_SOP_CLASS_NAMES_MODE_PREFIXES[ds.SOPClassUID.name]
    filename = "{}.{}".format(mode_prefix, ds.SOPInstanceUID)

	return filename
```

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
```diff
@@ -373,39 +375,111 @@ def anonymise_directory(
...
+                keywords_to_leave_unchanged=keywords_to_leave_unchanged,
+                delete_private_tags=delete_private_tags,
+                delete_unknown_tags=delete_unknown_tags)
+        except Exception as e:
+            failing_filepaths.append(dicom_filepath)
+            errors.append(e)
 
     # Separate loop provides the ability to raise Exceptions from the
     # unsuccessful deletion of the original DICOM files while preventing
     # these Exceptions from interrupting the batch anonymisation.
```
SimonBiggs: This is quite a neat way to handle it.

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
```diff
@@ -373,39 +375,111 @@ def anonymise_directory(
...
     if delete_original_files:
         for dicom_filepath in dicom_filepaths:
-            remove_file(dicom_filepath)
+            if not dicom_filepath in failing_filepaths:
+                remove_file(dicom_filepath)
```
SimonBiggs: I suspect we potentially shouldn't delete anything if there are any errors.

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
SimonBiggs: If there are any errors though I do wonder if it would be better simply to "fail fast" and not go any further. Let users sort out what's wrong before running over the rest of the directory...

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
```diff
@@ -373,39 +375,111 @@ def anonymise_directory(
...
 
+    for i, filepath in enumerate(failing_filepaths):
+        raise Exception("Could not complete anonymisation of {}; a {} "
+                        "error was raised. For more details, try "
+                        "anonymising {} individually.\n"
+                        .format(filepath, type(errors[i]), filepath))
```
SimonBiggs: I think here only the first filepath will be raised. Once the Exception is raised for the first issue it won't go any further, rendering the for loop pointless.

Does the following work:

```python
try:
   # for loop with exceptions goes here
except:
   raise
```

src/pymedphys/cli/dicom.py
```diff
@@ -81,16 +84,81 @@ def dicom_adjust_rel_elec_density(dicom_subparsers):
...
+        action='store_true',
+        help=("Use this flag to simply clear the values of all of the "
+              "identifying elements in the anonymised DICOM files, "
+              "as opposed to replacing them with 'dummy' values."))
+
+    parser.add_argument(
+        '-k', '--keywords_to_leave_unchanged',
+        metavar='KEYWORD',
+        type=str,
+        nargs='*',
```
SimonBiggs: Yup, nice :).

packages/pymedphys_dicom/tests/test_anonymise.py
```diff
@@ -17,75 +28,308 @@
...
+                                           ignore_private_tags)
+        else:
+            assert not is_anonymised_dataset(ds, ignore_private_tags)
+            assert not is_anonymised_file(temp_filepath,
+                                          ignore_private_tags)
+            assert not is_anonymised_directory(temp_dirpath,
+                                               ignore_private_tags)
+    finally:
+        remove_file(temp_filepath)
+        remove_dir(temp_dirpath)
```
SimonBiggs: Very nice :)

SimonBiggs:
> > Once we've got this sufficiently tested how about we fast track a webapp for this one? I'd be happy to put in place the framework.
> 
> Yes, please!

Alright, I think there is value in building the web apps into the documentation... I'm not totally sure, but I do think so... Hmmm... further thought.

Anyway, my steps are the following:

* Implement the ability to login to the docs with GitHub, will use [Firebase to facilitate this](https://firebase.google.com/docs/auth/web/github-auth).
  * Have a list of GitHub users that are marked as developers
  * Make it so that advertising shows, but isn't recorded against the viewable impressions tally if there is a developer.
* Implement viewable impressions on the advertising spot as well as the sponsor page
* Implement click through tracking
* Implement user location classification
* Implement pulling advertising images, click url, and prices paid from a database (once again using Firebase)
  * Have these ads be displayed on the website by user location, and prioritised by who is paying the most -- need to sort out the algorithm here, potentially included a little bit of randomisation so that 1c extra advertising spend doesn't result in only those ads being seen.
* Implement the ability for advertisers to log in, create an advert, and have it be uploaded to the database.
  * Advert would consist of an image for the advertising spot, the company's logo for the sponsors page, a url for the advert, and a url for the company's logo.
* Implement payment with stripe.
* Display to advertisers a report of click through rates, impressions, location of users who have seen the ads, as well as the number of impressions seen by developers that were not counted to the tally.
* When sponsors are going to pay they can see what others have paid, and have it be clear that those who pay the most have their advertising prioritised over those who haven't paid as much.

... then after that, let's build a web app using pyodide, blueprintjs and react :)

packages/pymedphys_dicom/tests/test_anonymise.py
Centrus007: Yep, sounds good :)

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
Centrus007: Wait, did I misinterpret what you were saying [here](https://github.com/pymedphys/pymedphys/pull/222#discussion_r279619915)? I thought you preferred this approach and so implemented it that way.

Happy to fail fast :)

SimonBiggs: Yeah let's just fail fast.

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
Centrus007: See commit c9767ad :)

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
Centrus007: Is this what you meant? It *is* what I had originally...

It fails fast in the sense the if an Error is throw during `anonymise_file()`, the program should simply exit (and thus the file deletion loop not executed).

Centrus007:
@SimonBiggs, you've left your approval, but would you like to merge once happy? You've done quite a bit of review...

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
```diff
@@ -604,14 +601,13 @@ def _filter_identifying_keywords(keywords_to_leave_unchanged):
...
-    return keywords_filtered
 
 
 def get_anonymous_replacement_value(keyword):
-    """Get an appropriate dummy anonymisation value for a DICOM element
-    based on its value representation (VR)
+    """Get an appropriate dummy anonymisation value for a DICOM element based
+    on its value representation (VR)"""
+    vr = BASELINE_KEYWORD_VR_DICT[keyword]
+    return VR_ANONYMOUS_REPLACEMENT_VALUE_DICT[vr]
```
SimonBiggs: erghh, sorry. Some weird linter went to town and broke things.

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
```diff
@@ -302,12 +302,12 @@ def anonymise_file(
...
-                                            dirpath=dirname(dicom_filepath))
-        dicom_anon_filepath = label_dicom_filepath_as_anonymised(path)
+        filepath_used = create_filename_from_dataset(
+            ds, dirpath=dirname(dicom_filepath))
     else:
-        dicom_anon_filepath = label_dicom_filepath_as_anonymised(
-            dicom_filepath)
+        filepath_used = dicom_filepath
+
+    dicom_anon_filepath = label_dicom_filepath_as_anonymised(filepath_used)
```
SimonBiggs: I'm very sorry, all I wanted to do was change just the above. Something broke and I committed bad auto linting :/

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
SimonBiggs: Upgrading the Python extensions in VSCode fixed the problem.

SimonBiggs:
Figured it wasn't too much of an issue making that small change. So I merged. Thank Matt :)

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
Centrus007: All good :)

packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py
SimonBiggs: Yeah, sorry about that. I think a solution to the problem that was running through my head was to keep the output directory separate to the input directory. Meaning a half run program that still fails fast wouldn't have the outputs mixed in with the inputs, so the code could be repeated and it wouldn't try to use outputs as inputs for the next iteration.

I think this would be solved by using an implementation of output directory however. So we can leave that for another branch.

Sorry for sending you all over the place with that comment.

Centrus007: All good.

We could ignore files with the "_Anonymous" label?

SimonBiggs: Interesting solution ... It worries me though, if we start to ignore files like that. If I run anonymise on a file, no matter what its name I do expect it to run anonymisation over it. If we start having certain cases where that doesn't happen it starts to get quite confusing.

I think for my personal use I'll follow the separation of concerns pattern by keeping input directory and output directory apart from each other. We can sort that out at some point.

PR #226, opened 2019-04-30
[RFR] Netlify docs

SimonBiggs:
Here are the docs built with netlify, using yarn:

https://staging.pymedphys.com

This means, once we remove readthedocs we can pour all of our install resources into using yarn, and make that be "the one way" instead of needing requirements files.

I'll likely just be merging this one in so that I can keep iterating, attempting to make multiple version deploys.


SimonBiggs:
Sorry for the force pull on this one. Nature of using branches to build docs, I kinda need it merged to keep working on it ...

Centrus007:
All good :)

SimonBiggs:
@Centrus007 by the way, the documentation preview built automatically for every pull request is awesome :). You just need to click "details" next to the following item:

image

Centrus007:
Yep, i noticed that! Very cool :D

PR #252, opened 2019-05-05
Split docs and app out

SimonBiggs:


SimonBiggs:
@Centrus007 so this is working :), and it is awesome. We need to strip out some dependencies so that we can begin using these tools... :)

Centrus007:
How do I test it?

SimonBiggs:
Certainly not worth a merge. It's still completely experimental, but go to https://deploy-preview-252--pymedphys.netlify.com/ the press `Ctrl + Shift + I` and you will see the output of the following bit of code:

```python
import pydicom

dataset = pydicom.Dataset()
dataset.PatientName = 'MacDonald^George'
print(dataset)
```

SimonBiggs:
Magical :)

Centrus007:
Very, very cool!

SimonBiggs:
@Centrus007 so I realised we can't actually have the pypi version be the version being used in the online apps. What if we want to update some of the python code and have that new python code show up in the pull request preview for example?

Anyway, so I thought about it further, and anyway, I have our packages loading from the repo itself. So, if you now go to https://deploy-preview-252--pymedphys.netlify.com/ and open up the console you should see the result of the following code printed:

```python
from pymedphys_dicom.dicom import DicomBase

dicom = DicomBase.from_dict({
    'Manufacturer': 'PyMedPhys',
    'PatientName': 'Python^Monte'
})

dicom.anonymise(inplace=True)

print(dicom.dataset)
```

It will take a while to come up because I have told it to download numpy, pandas, and matplotlib as well as every package in the library (not all of which are needed). But for now, I just wanted to pull everything in.

SimonBiggs:
I'm pulling out the code for merging, so if you want to see `pymedphys_dicom` running you'll need to use the following link:

https://5cdfa03c0af00b0009d5b51e--pymedphys.netlify.com/

SimonBiggs:
Going to force merge this so that I can iterate on new docs format

PR #255, opened 2019-05-05
RFR: Try darker grey navbar.

Centrus007:


Centrus007:
I think we could do with an improved color scheme. A bit more contrast between the navbar and body for sure. But not suuuuper sold on this particular navbar colour either...

SimonBiggs:
I'd say a touch lighter, and make the version picker a touch darker so that there's also contrast there...

Centrus007:
So, cheekily, I actually made it a touch *darker* to try to match the grey background to the right and bottom of the body. Maybe I shouldn't.

Played with the version number as well. Still not great to be honest.

SimonBiggs:
Yeah... I don't like it... sorry...

Centrus007:
> Yeah... I don't like it... sorry...

Fair enough.

What I'm trying to achieve is contrast between the navbar and the body. I think this is important if the navbar scrolls. It looks like the original design mimics `pydicom`, but I honestly think it's a poor example of design. It looks very unnatural to me for a section of a page that isn't visibly separated from the rest of the page to scroll independently of the rest of the page. And the very slight colour change you *did* put in at my request is barely perceptible... (just having a look at `pydicom` again, and it also has a very, very slight colour difference. But IMHO, it's really not enough...)

Moreover, searching through examples of documentation, `pydicom`'s design is very rare. Almost every example I can find has a clearly differently coloured sidebar.

That said, I'm really not a big fan of what I've tried on this PR so far. I started tinkering with grey since the menu's colour scheme is in shades of grey. But I reckon it might be worth exploring a different colour scheme entirely!

Centrus007:
Just for reference:

- [Django](https://docs.djangoproject.com/en/2.2/) has a white body with a grey sidebar (on right)
- JupyterLab(https://jupyterlab.readthedocs.io/en/stable/) has a white body with dark grey sidebar on left
- [Numpy](https://docs.scipy.org/doc/numpy-1.14.2/reference/generated/numpy.array_equal.html), Scipy and matplotlib avoid independently scrollable sidebars entirely...
- [TensorFlow](https://www.tensorflow.org/api_docs/python/tf/assign_sub) has a border between the two - which is admittedly something else to try!

SimonBiggs:
Tensorflow is using material design elevation to distinguish the elements. I really like material design. https://material.io/design/environment/elevation.html#elevation-in-material-design

If you like that we can pull in the material hi library into the docs (https://material-ui.com) and use their design concepts.

Centrus007:
How about that latest one? Reverted back to original but with a subtle right border?

Centrus007:
Also note that the existing design introduces a border when accessing submenus, which looks a little out of place with the no border main menu:

image



And with the newly introduced right border:

image

Centrus007:
> Tensorflow is using material design elevation to distinguish the elements. I really like material design. https://material.io/design/environment/elevation.html#elevation-in-material-design

> If you like that we can pull in the material ui library into the docs (https://material-ui.com) and use their design concepts.

I could get around it! But a quick fix for the short term would be nice as well :)

SimonBiggs:
Beautiful :)

packages/pymedphys_sphinxtheme/src/sass/_theme_badge.sass
```diff
@@ -92,8 +92,8 @@
 
 
 .rst-versions .rst-current-version
-  background-color: #eaeaea
-  color: #55a5d9
+  background-color: #dddddd
+  color: #0077CC
```
SimonBiggs: What does this change?

SimonBiggs: I'm okay with this version colouring, but was leaving this change in there intentional?

Centrus007: It was intentional for the colour change, but I meant to revert it for the border option. I'll revert it :)

Issue #253, opened 2019-05-05
Strip out heavy dependencies


SimonBiggs:
Offending packages can either be removed, refactored, or bumped up into the toolbox for now.

Dependencies that we need to get rid of:

* Pillow
* psutil
* and others...

File formats needs to not depend on databases.

SimonBiggs:
Eh, do this on an as needs basis.

Issue #223, opened 2019-04-30
Pyodide + DICOM anonymise


SimonBiggs:
I propose that one of the first web apps we build is for the DICOM anonymiser.

It would use pyodide, so it would be completely run user side. Therefore the identifying DICOMs won't be sent to the server. They will be processed client side.

Note that the following pull request is almost complete. We've also implemented the split. So all systems are almost go.

https://github.com/iodide-project/pyodide/pull/147

SimonBiggs:
The pyodide pip install method has merged :) ... I want to take it for a spin ...

SimonBiggs:
We need to cut a release of our subpackages with their new dependencies.

SimonBiggs:
Done

PR #256, opened 2019-05-05
[RFR] Remove heavy dependencies from dicom and fileformats

SimonBiggs:
@kingrpaul to be able to achieve the following:

https://github.com/pymedphys/pymedphys/issues/223

I needed to remove the Pillow dependency. Pillow is quite a heavy dependency, so I moved narrow_png to labs. See the following issue:

https://github.com/pymedphys/pymedphys/issues/253

Hope this is okay...

`pymedphys_dicom` depends on `pymedphys_analysis`, which is why psutil and Pillow needed stripping out.


package.json
```diff
@@ -8,16 +8,17 @@
     "docker:internal": "yarn install:prod",
     "git:clean": "git clean -fxd",
     "git:undo": "git reset --soft HEAD~1",
-    "install:dev": "yarn lerna run pip:install:dev && pip install -e .",
-    "install:prod": "yarn lerna run pip:install:prod && pip install .",
+    "install:dev": "yarn lerna run pip:install:dev --concurrency 1 && pip install -e .",
+    "install:prod": "yarn lerna run pip:install:prod --concurrency 1 && pip install .",
```
SimonBiggs: I believe this fixes issues with running `yarn bootrap:dev:all`

SimonBiggs:
Let's merge, I'll need to depend things on this in the near future...

SimonBiggs:
Thanks @kingrpaul :)

kingrpaul:
> 
> 
> Thanks @kingrpaul :)

You're welcome @SimonBiggs. Though having some difficulties, was pretty much in the process of moving most (all?) of mine into "labs". Difficulties may pass when you finish the changes you are working on. Expaning but also putting a GUI in front of things so that I can use / share more easily, but trying to make sure will integrate with pymedphys-GUI when that is settled.

SimonBiggs:
You've confused me a little there. Can you explain that again

kingrpaul:
Sorry, I used too much short-hand. 

I am already moving most or all of my stuff into the labs. So, it is not a problem for me that you find it necessary to move something into the labs. In addition to expanding the functions, I am setting up a user interface, to make these things more useful to me. But, I hope to find that they will integrate easily with the pymedphys user interface, when that is developed. I am having some challenges in moving things to the lab. But, I know that you are making major structural changes, so I hope that my difficulties will lessen after those changes are complete. 

I think I missed part of the discussion, though. So, I am not sure what is being sought in making the structural changed that you are making. 

Hoping that was less cryptic.

SimonBiggs:
Yup, that makes heaps more sense :). The main reasoning for splitting the
back end up into subpackages is that in the rare case where environments
cannot pull in two many dependencies, it is possible to still use portions
of pymedphys without having to install all of the dependencies.

This is particularly important for pyodide, where the options of which
dependencies can be used are currently quite small.

The greater majority of users of pymedphys will be none the wiser as the
main package, pymedphys, still exports the same interface.

On that note, if you're finding it difficult to move packages to the labs I
could do that for you if you would like?

Issue #260, opened 2019-05-07
Make gamma compatible with pyodide


SimonBiggs:
We need to repackage a bit further. Turns out although scipy works, it is humungous and doesn't get cached.

https://github.com/iodide-project/pyodide/issues/240

Until that's fixed it is not feasible for us to build any webapps that have scipy as a dependency.

...so, we need to do a bit more splitting and refactoring so that the stuff we really want (dicom and trf reading) has no dependency on scipy. Should be doable, just need to sort things out a bit.

We should do this before we forge a subpackage release of 0.2.0.

Centrus007:
Gamma would be nice too!

SimonBiggs:
We would have to swap in our own linear interpolation for that...

SimonBiggs:
... I'm not keen on that level of restructuring just yet...

SimonBiggs:
... I think we need to decouple dicom from analyse. If we need to we can make two dicom packages, one that depends on analyse and the other that doesn't...

SimonBiggs:
I could just copy "cube geometry" into dicom. It's only purpose at the moment is to extract dose planes from DICOM files... So it possibly better fits there anyway...

https://github.com/pymedphys/pymedphys/blob/4d3302df15353dc3a4c6ac3af1c8550aa4d8304c/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level1/structure.py#L39

SimonBiggs:
This needs to go too:

https://github.com/pymedphys/pymedphys/blob/4d3302df15353dc3a4c6ac3af1c8550aa4d8304c/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/dose.py#L37

SimonBiggs:
And that's it! Gee, that should be easy enough.

Centrus007:
> We would have to swap in our own linear interpolation for that...

For now, could we simply extract the relevant code from `scipy` and insert somewhere in `pymedphys` (with the appropriate acknowledgements, of course)? If so, that'd be pretty easy...

SimonBiggs:
Yes certainly, as long as it's mostly standalone within scipy... I think that interpolation function is reltively standalone ...

That could solve https://github.com/pymedphys/pymedphys/issues/260#issuecomment-490050260 also... Yeah, I think that's a good idea ...

SimonBiggs:
[...]

PR #261, opened 2019-05-07
[RFR] Remove scipy from dicom and fileformats

SimonBiggs:


kingrpaul:
@SimonBiggs: if you would, very briefly, why remove Scipy?

SimonBiggs:
Oh sorry, bit completely removing it. Just removing it from some of the packages. Namely dicom and fileformats.

It's not needed there, and we want to be able to use those packages within an online webapp.

See https://github.com/pymedphys/pymedphys/issues/260 for more info

azure-pipelines.yml
```diff
@@ -8,11 +8,11 @@ jobs:
...
-        # Pylint-Subpackages:
-        #   image_name: ubuntu-16.04
-        #   test_title: Pylint Subpackages
-        #   toxenv: analysis,dicom,fileformats,labs,logfiles,coordsandscales,electronfactors,databases,utilities,workshops,toolbox,xlwings
-        #   python: 3.7
+        Test-Subpackages:
+          image_name: ubuntu-16.04
+          test_title: Test Subpackages
+          toxenv: analysis,dicom,fileformats,labs,logfiles,coordsandscales,electronfactors,databases,utilities,workshops,toolbox,xlwings
+          python: 3.7
```
SimonBiggs: We really do need this... especially if I am refactoring packages. It would really suck to release standalone packages which won't work in a standalone environment ...

packages/pymedphys_analysis/src/pymedphys_analysis/gamma/_level1/gammafilter.py
```diff
@@ -33,7 +34,7 @@
 
 import numpy as np
 
-from pymedphys.dicom import coords_from_xyz_axes
+from pymedphys_dicom.dicom import coords_from_xyz_axes
```
SimonBiggs: To keep the subpackages independent of the main package we need to only import other subpackages.

Also, I need to get that layer linter asap, because the moment pymedphys_dicom was used within pymedphys_analysis we had a circular dependency...

I have just refactored things so that the analysis depends on dicom, but prior to this pull request it was the other way round...

SimonBiggs:
Erghh... the subpackage cli tests fail... because pymedphys isn't installed ... hmm...

SimonBiggs:
@Centrus007 yayy! The branch passes. :sweat_smile: 

That took a while. Anyway, for now we are ignoring CLI tests when we are just testing subpackages by themselves.

This is ready for review and merge... and then we release subpackages-0.2.0 :).

SimonBiggs:
I didn't exactly pick it up. I wrote the new pipelines, and the new pipelines picked it up.

And I do agree with the "how to" ... I was just thinking last night we should have a document describing how to begin contributing to the labs. As in, how to take some of your in house code, make your own named labs directory, create a pull request, etc...

SimonBiggs:
Thanks for the review @Centrus007 :)

SimonBiggs:
There is a great place to build an ecosystem of a labs packages. For new contributors we need to focus on making contributing to the labs well documented, and sufficiently easy to begin. Those who are enthusiastic and willing to put in the time will then learn how to begin moving some of their work from labs into the main packages. I think that's a reasonable stepping stone...

package.json
```diff
@@ -10,6 +10,7 @@
     "git:undo": "git reset --soft HEAD~1",
     "install:dev": "yarn lerna run pip:install:dev --concurrency 1 && pip install -e .",
     "install:prod": "yarn lerna run pip:install:prod --concurrency 1 && pip install .",
+    "install:subpackage": "yarn lerna run pip:install:test -- --scope=$SUBPACKAGE --include-filtered-dependencies --concurrency 1",
```
SimonBiggs: @Centrus007 this command is really cool by the way! :). So glad we decided to hone in on using lerna with yarn. This takes a package defined by scope, and then runs a given command in that package and the packages that package depends upon. And the order it runs them in is dependency order, starting at the bottom of the tree.

So if scope is defined as pymedphys_analysis it will run the command in the following packages on the following order:

* utilities
* coordsandscales
* dicom
* analysis

So, this is the perfect way to just install a subpackages and its dependencies only.

PR #265, opened 2019-05-09
[RFR] Mere aggregation

SimonBiggs:
Included section about mere aggregation within FAQ


SimonBiggs:
Erghh, I stuffed that branching up

PR #266, opened 2019-05-09
RFR: Perform CLI tests in temp dirpath

### LINKED ISSUE #262: Test Anonymise CLI can't cope with parallel running
The following test:

https://github.com/pymedphys/pymedphys/blob/86fce19771b71b1008a2ab7ffdb24224b19fa78f/packages/pymedphys_dicom/tests/test_anonymise.py#L246-L250

Intermittently fails when tests are being run in parallel:

```
_________________________________________________________________ test_anonymise_cli _________________________________________________________________

    def test_anonymise_cli():
    
        # Basic file anonymisation
        assert not is_anonymised_file(test_filepath)
>       assert not exists(test_anon_filepath)
E       AssertionError: assert not True
E        +  where True = exists('/home/simon/git/pymedphys/packages/pymedphys_dicom/tests/data/anonymise/RP.1.2.246.352.71.5.53598612033.430805.20190416135558_Anonymised.dcm')

packages/pymedphys_dicom/tests/test_anonymise.py:250: AssertionError
```

Centrus007:
Fixes #262 

I should've noticed this ages ago, since I've used them quite a bit for RAH projects, I'd just forgotten because it'd been a while since I set them up.

We ought to use `pytest.fixtures` to run pre- and post-testing code. E.g. see [`tmp_path `](https://docs.pytest.org/en/latest/tmpdir.html). It's perfect for making temp dirs and ensuring their deletion after testing, regardless of the testing outcome. Much cleaner than a million `try/finally`s, I think.

Maybe for another PR though - I'm going to be flat out with other things until next Tuesday... :(

### DIFF
--- a/packages/pymedphys_dicom/tests/test_anonymise.py
+++ b/packages/pymedphys_dicom/tests/test_anonymise.py
@@ -33,7 +33,6 @@ DATA_DIR = pjoin(HERE, 'data', 'anonymise')
 test_filepath = pjoin(DATA_DIR, "RP.almost_anonymised.dcm")
 test_anon_basename = \
     "RP.1.2.246.352.71.5.53598612033.430805.20190416135558_Anonymised.dcm"
-test_anon_filepath = pjoin(DATA_DIR, test_anon_basename)
 file_meta = read_file_meta_info(test_filepath)
 temp_dirpath = pjoin(DATA_DIR, 'temp_{}'.format(uuid4()))
 temp_filepath = pjoin(temp_dirpath, "test.dcm")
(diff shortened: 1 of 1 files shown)


SimonBiggs:
> We ought to use `pytest.fixtures` to run pre- and post-testing code.

Yup, we certainly should. Good idea. Thanks @Centrus007

Centrus007:
I think that single commit did the trick :)

SimonBiggs:
The default tmp dir fixture looks awesome:

https://docs.pytest.org/en/latest/tmpdir.html

Centrus007:
> The default tmp dir fixture looks awesome:

> https://docs.pytest.org/en/latest/tmpdir.html

It does look pretty good, but still might not serve our purposes - it creates a unique path at each *invocation*, but I'm afraid this might not hold for parallel cases. It might. But even if it doesn't, we can use the `uuid()` in our own custom fixtures easily enough :)

PR #278, opened 2019-05-11
[RFR] Monomanage

SimonBiggs:


SimonBiggs:
So, what I'm planning on producing is something like the following for the packages, then for each package have a graph of the dependencies for each module, then for each module in each package have a graph of dependencies of each of the files:

https://github.com/pymedphys/pymedphys/pull/278/files?file-filters%5B%5D=.svg#diff-4aacaa814b40b0864a491dd1f057913b

image

There will also be a test at each of these layers that there are no cyclic dependencies. All these graphs will be placed within the docs, each node on the graph will be hyperlinked and will take you to either the corresponding dependency graph for that node, or in the case of a node representing a file, it'll take you to the GitHub link for the source code of that file.

SimonBiggs:
Alright, so the module graphs are complete. See https://github.com/pymedphys/pymedphys/pull/278/commits/88e3f2a4b60a9880985eaa9d0ce1799a4b1d8b47?file-filters%5B%5D=.svg#diff-acf60b6786568843664f2b789c53abf0

SimonBiggs:
Heya @Centrus007, this one is ready for review and merge. Next up I need to make graphs for the dependencies relationships of the files within modules themselves. 

But I'd rather do that in a fresh pull request.

SimonBiggs:
Here are all the current graphs by the way:

https://github.com/pymedphys/pymedphys/tree/88e3f2a4b60a9880985eaa9d0ce1799a4b1d8b47/docs/graphs

SimonBiggs:
Here's an exceptionally pretty dependency tree:

pymedphys_dicom dicom

:)

Anyway, all these graphs are built. Some of the code seriously smells... But it works...

The pipeline is now testing that there are no cycles, and that the most up to date dependencies have been included.

I think a job for another day is to make it into a commit hook so that anytime a new import is undergone it automatically propagates the new dependency structure.

SimonBiggs:
> I think a job for another day is to make it into a commit hook so that anytime a new import is undergone it automatically propagates the new dependency structure.

All good, just implemented the hook. Essentially whenever you commit, if there was a change in the internal dependency structure of any package then all dependencies get updated and propegated to `_install_requires.py` as well as the graphs for the docs.

Centrus007:
Fantastic ideas and great job!

SimonBiggs:
> Fantastic ideas and great job!

Thanks Matt :)

PR #283, opened 2019-05-13
[RFR] Hook improvements

SimonBiggs:
Various improvements to the new dependency commit hook

PR #284, opened 2019-05-14
[RFR] 0.8.x

SimonBiggs:
I have made a set of release candidates. We should install these and play with them a bit. See if everything works. Then I think we should forge a release (that way we can begin playing with the files within pyodide...)


SimonBiggs:
This depends was based on top of https://github.com/pymedphys/pymedphys/pull/283

SimonBiggs:
Actually, I think it's worth making a full 0.8.0 release. We won't be releasing it on conda for a while, but getting a full release out in on pypi will be worthwhile.

SimonBiggs:
Let's just merge this in within https://github.com/pymedphys/pymedphys/pull/285

Centrus007:
I'm having trouble following, sorry. Do you still want to release 0.8.0? Should this be closed?

SimonBiggs:
@Centrus007 all good. No need to merge this though.

PR #290, opened 2019-05-15
[MRG] Extend CT

SimonBiggs:
We were umming and ahhing about whether or not this sort of thing should be put within PyMedPhys...

Your thoughts would be appreciated


SimonBiggs:
@msobolewski just requesting your review due to your involvement.

packages/pymedphys_dicom/tests/test_extend_ct.py
```diff
@@ -0,0 +1,75 @@
...
+            'ImagePositionPatient': ['0.0', '0.0', str(float(slice_location))]
+        })
+        for i, (uuid, slice_location)
+        in enumerate(zip(uuids, slice_locations))
+    ]
+
+    return convert_datasets_to_deque(initial_datasets)
+
+
+def test_extend_datasets():
```
SimonBiggs: @Centrus007 @msobolewski this is the primary test function making sure it is working as expected.

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
```diff
@@ -0,0 +1,176 @@
...
+    new_filepaths = [
+        os.path.join(output_dir, filename)
+        for filename in new_filenames
+    ]
+
+    for dicom_dataset, filepath in zip(dicom_datasets, new_filepaths):
+        dicom_dataset.save_as(filepath)
+
+
+def extend_datasets(dicom_datasets, index_to_copy, number_of_slices, uids=None):
```
SimonBiggs: @Centrus007 @msobolewski here is the function called by the test.

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
```diff
@@ -0,0 +1,180 @@
...
+import random
+from collections import deque
+
+import pydicom
+from ..constants.uuid import PYMEDPHYS_ROOT_UID
+
+
+def extend(input_dir, output_dir, index_to_copy, number_of_slices):
+    filepaths = glob(os.path.join(input_dir, '*'))
+    dicom_datasets = load_dicom_into_deque(filepaths)
```
Centrus007: `dicom` in `dicom_datasets` is superfluous. We're in the `dicom` package...

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
```diff
@@ -0,0 +1,180 @@
...
+    filenames = [
+        os.path.basename(filepath)
+        for filepath in filepaths
+    ]
+
+    common_prefix = os.path.commonprefix(filenames)
+
+    number_of_digits = len(str(len(dicom_datasets)))
+    new_filenames = [
+        "{}{}.dcm".format(
```
Centrus007: Not `CT.SOPInstanceUID`? That's how both our old Philips and new Toshiba CT scanners name CT files...

It's actually pretty standard across all DICOM modalities, not just CT.

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
```diff
@@ -0,0 +1,180 @@
...
+        for filepath in filepaths
+    ]
+
+    dicom_datasets = convert_datasets_to_deque(dicom_datasets_initial_read)
+
+    return dicom_datasets
+
+
+def convert_datasets_to_deque(datasets):
+    dicom_datasets = deque()
```
Centrus007: A little confusing. `datasets` is a list but `dicom_datasets` is a `deque`.

Maybe replace `dicom_datasets` with `dataset_deque`?

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
```diff
@@ -0,0 +1,180 @@
...
+
+def instance_number(dicom_dataset):
+    return dicom_dataset.InstanceNumber
+
+
+def slice_location(dicom_dataset):
+    return float(dicom_dataset.SliceLocation)
+
+
+def copy_slices_and_append(dicom_datasets, index_to_copy, number_of_slices):
```
Centrus007: Adds to confusion. Should `dicom_datasets` be a `deque`?

packages/pymedphys_dicom/tests/test_extend_ct.py
Centrus007: Not to revive an old debate, but I'd really like to see these functions tested on actual DICOM CT files. At least once...

Centrus007: So I guess a test of `ct.extend()`...

Centrus007:
Perhaps a little ambitious, but I was hoping to see a `DicomCTSeries` or similar object here :p

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
```diff
@@ -0,0 +1,180 @@
...
+    if uids is None:
+        uids = generate_uids(len(dicom_datasets))
+
+    for dicom_dataset, uid in zip(dicom_datasets, uids):
+        dicom_dataset.SOPInstanceUID = uid
+
+
+def generate_new_slice_locations(dicom_datasets, index_to_copy, number_of_slices):
+    if index_to_copy == 0:
+        slice_diff = dicom_datasets[0].SliceLocation - \
```
Centrus007: One almost insignificant point:

It might pay to use `ds.ReconstructionProtocolElementSequence.SpacingBetweenSlices` **if it exists** (it's an optional parameter).

This accounts for cases with missing slices. Not as rare as one would hope...

Centrus007: Or get minimum separation across all slices?

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
SimonBiggs: See https://stackoverflow.com/a/8793247/3912576

SimonBiggs: All for changing the name, but I'm not keen on using hungarian typing...

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
SimonBiggs: Key is, it doesn't have to be a deque. As long as the provided append method exists on the object.

SimonBiggs: If it quacks like a duck, it can go in there:

https://dev.to/alvarocavalcanti/practical-python---duck-typing-to-the-rescue-46e6

SimonBiggs: (A list will work for appending slices to the end)

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
SimonBiggs: Yup, very fair points. Not insignificant at all.

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
SimonBiggs: I'd certainly be happy to change to that. This was just based on how Monaco exports the DICOM. More than happy to use a different (better) standard :).

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
SimonBiggs: Maybe when it's on a DICOM object...

I'd agree if I called teh function `dicom_ct_extend` which would be superfluous. But the variable name doesn't necessarily have to be DICOM...

packages/pymedphys_dicom/tests/test_extend_ct.py
SimonBiggs: Fair point, we have a beautiful anonimiser, which I can use to drop some files in :)

SimonBiggs:
> Perhaps a little ambitious, but I was hoping to see a `DicomCTSeries` or similar object here :p

:) slow and steady wins the race :)

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
SimonBiggs: So I changed it to `<Whatever prefix was common before><SOPInstanceUID>.dcm`

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
SimonBiggs: Could you provide me a DICOM file that has the `SpacingBetweenSlices` element on it?

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
Centrus007: Really? We have quite a number of examples in `pymedphys` where we *have* done this. Particularly for dictionaries...

Not to mention that any time we use a variable name like `dataset` for an instance of class `Dataset`, we are basically doing the same thing as my suggestion above. We have dozens of examples of `ds_<description>`. Would you say the `ds` should be removed? Surely not...

The SO post makes sense for simple data types where type is either (clearly) implicit or you want to maintain type flexibility (e.g. you might be happy for an `int` to be cast as a `float` and so `age_int` would be stupid). But I really don't think this holds for more complex abstract data types, especially where you want to maintain the specific type (otherwise functionality will break). I would argue this is the case for your use of `deque` here...

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
Centrus007: Fair enough for this specific function, but does that apply to all of your `deque` functions?

Centrus007: If so, ignore all my comments regarding renaming `datasets` (except perhaps the still arguably superfluous `dicom` hehe). If all uses of `deque` are truly perfectly compatible with `list`, I'm happy :)

SimonBiggs: A deque is essentially a list, just stored in memory in a way optimised for appending and removing items from either side...

SimonBiggs: List doesn't have an 'appendleft' method, but that will raise an error if one tries to append left with a list.

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
Centrus007: I don't think that works... what if part of the common prefix includes part of the SOPInstanceUID. I think it always will, actually...

Centrus007: > This was just based on how Monaco exports the DICOM.

Ah. Sorry, I thought only Varian were annoying enough to break convention...

> More than happy to use a different (better) standard

Not sure it's better, to be honest. Often the alphabetical order of conventionally named CT slices named doesn't match slice order. Maybe your way is better after all.

I only raised it because we adopted that convention in the anonymiser. I.e. a file basename of:

`<2-char-modality>.<SOPInstanceUID>.dcm`

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
Centrus007: The `dicom` module is pretty analogous to a dicom class. And given our extension use of `dataset`/`ds`, dataset is effectively our "object" of that class (for now)...

I see the value of "explicit is better than implicit", so if you're firm on this, I'll let it go.

I guess I'm just starting to cringe at our growing lack of consistency in our variable names for instances (or collections of instances) of `pydicom`'s `Dataset` . We now have lots of `dicom_dataset`, `dataset` AND `ds`.

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
Centrus007: Fair enough. :)

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
SimonBiggs: Ahh okay, I do like it printing in slice order. Alright, I might revert it...

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
SimonBiggs: Yeah... I see the cringe. ... In this case a believe it helps the readability of this code... Okay we stick with this here?

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
Centrus007: No worries :)

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
Centrus007: Yep :)

SimonBiggs: Kk :) thanks :)

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
SimonBiggs: I'll leave this one unresolved so I remember to revert :)

packages/pymedphys_dicom/src/pymedphys_dicom/ct/extend.py
Centrus007: I don't actually have one. I only thought of it from my own research running into this issue for a script that force aligns a CT to the z-axis

Incidentally, I was planning to add this to `pymedphys`. The use case is that we often come across diagnostic CT scans from scanners that seem to have a very slight skew between the imaging and scanner coordinate systems (and RayStation insists upon perfect z-axis alignment - for now at least).

Centrus007: I did try to find one on the internet to no avail. I'll keep looking.

SimonBiggs: Kk, thanks @Centrus007

packages/pymedphys_dicom/tests/test_extend_ct.py
SimonBiggs: > Not to revive an old debate, but I'd really like to see these functions tested on actual DICOM CT files. At least once...

Don't want to leave this pull request hanging for this need if that's okay?

Will assume this is a prefixed `Nit:` comment.

SimonBiggs:
This one is ready for merge

packages/pymedphys_dicom/tests/test_extend_ct.py
Matthew-Jennings: So I've run into issues with the `dataset_from_dict` before. It's possible to create datasets that are valid within `pydicom` but then won't write to file.

If this is a function that is likely to result in DICOM file writes, I would really, really prefer you add file-based testing. Moreover, it really is quite a quick and easy fix.

SimonBiggs:
Might be best for a call regarding what you're after.

packages/pymedphys_dicom/tests/test_extend_ct.py
SimonBiggs: These tests here are primarily to test the appropriate header conversions. By design `dicom_dataset_from_dict()` in it's basic form will likely never write to file (because it needs all the extra required header info). However, these tests cover the key items of logic that these functions are undergoing.

Requiring file io for our tests when the logic has been sufficiently tested will cause our testing time to blow out. I am actually hoping in the future we can reduce the amount of file io our tests are doing.

I believe we can learn a lot from how `pydicom` itself is tested:

https://github.com/pydicom/pydicom/blob/master/pydicom/tests/test_dataset.py

It covers a lot of important cases while still having the test suite run exceptionally fast.

packages/pymedphys_dicom/src/pymedphys_dicom/shim.py
```diff
@@ -0,0 +1,14 @@
...
+    def __getattribute__(self, name):
+        raise self.import_error
+
+
+try:
+    import pydicom
+
+    pydicom.config.enforce_valid_values = True
+except ImportError as e:
+    pydicom = Shim(e)
```
SimonBiggs: Once we remove all of the sub packages I thought the above method would be a good way to make packages optional.

packages/pymedphys_dicom/tests/test_extend_ct.py
Matthew-Jennings: Reiterating (and prettifying) some of our phone conversation here for reference :):

I'm concerned about the fact that `pydicom` permits edits to a dataset's header that would make the dataset "invalid". I.e., that the edited dataset would not successfully write to file or, worse, write but not be DICOM compliant. I'd noticed this when writing tests for DICOM anonymisation. I also noticed that this was less often the case if editing a dataset that was originally read from a file, and reasoned that there might be additional internal validity checks in `pydicom` when using datasets read from a file (e.g. mediated through the `FileDataset` class rather than `Dataset`).

For this reason, I've tended to prefer testing on datasets actually read from files (instances of `FileDataset`). Indeed, I've even tested for successful write of a DICOM file post-edits and re-read the file to check for correct edits (e.g. see [here](https://github.com/pymedphys/pymedphys/blob/4ed85f587f01e7827178b3121b1ef1d7fa27bddb/packages/pymedphys_dicom/tests/test_anonymise.py#L77) and [here](https://github.com/pymedphys/pymedphys/blob/4ed85f587f01e7827178b3121b1ef1d7fa27bddb/packages/pymedphys_dicom/tests/test_anonymise.py#L81)). I'd taken the philosophy that any function that returns an edited DICOM dataset and that will likely be written to file really ought to be tested for successful file write .

@SimonBiggs argued that file I/O is computationally expensive and in many cases probably testing overkill. I agree, especially with the former point. I then recalled that I (thought I'd) seen a `pydicom` function that explicitly checked for DICOM validity. We agreed that this would provide a suitable compromise: it avoids the risk of tests not catching invalid edits because the tests themselves work with looser constraints while also avoiding the computational code of file read/writes.

We then discovered `pydicom.config.enforce_valid_values = True`, which appears to be a configuration option that, when switched "on", adds validity checks for all DICOM dataset editing operations. If it works as we believe it does, we tentatively decided that this should simply be switched on for all testing, and possibly also for the library itself.

Still TBD:
- what exactly does `pydicom.config.enforce_valid_values = True` enforce?
- if suitable, substantially reduce DICOM file I/O in testing suite.

SimonBiggs:
Beautiful summary :)

packages/pymedphys_dicom/src/pymedphys_dicom/shim.py
Matthew-Jennings: Yep, lovely :)

I'm not sure I understand the `Shim` class? Why not:

```python
try:
    import pydicom

    pydicom.config.enforce_valid_values = True
except ImportError:
    raise
```

SimonBiggs: Essentially it makes it so the import error is raised when pydicom is actually attempted to be used instead of at the initial import time. It will need a bit of playing with to work out how best to implement it. But I imagined having a shim package which also has numpy and all other external packages. Then we just always import external packages via shim everywhere...

I think it'll be one of those things that will improve as we implement it I believe...

packages/pymedphys_dicom/tests/test_extend_ct.py
SimonBiggs: Beautiful summary :)

PR #292, opened 2019-05-15
[RFR] Remove dep tree hook

SimonBiggs:
Further thought on this. I think commit hooks should be reserved for things that absolutely must not be committed. CI (pipelines) can cover the rest.


SimonBiggs:
So in the future, other privacy or security focused hooks would be valid. And maybe something that checks for a maximum file size... but not much else.

SimonBiggs:
:)

PR #296, opened 2019-05-16
[WIP] Update file-structure docs for new structure.

Matthew-Jennings:


Centrus007:
Umm, not sure why this si failing... :/

SimonBiggs:
That was an intermittent network error from downloading a Zenodo file:

image

docs/developer/file-structure.rst
```diff
@@ -148,18 +118,16 @@ The PyMedPhys Source Code
...
-dependency hierarchy of code within modules. See diagram below:
+main ``pymedphys`` package, located in
+``pymedphys/packages/pymedphys/src/pymedphys/``. However, no *library* code
+actually exists within this directory. Instead, all library code is arranged
+into PyMedPhys' various subpackages, such as ``pymedphys_analysis`` and
+``pymedphys_dicom``. From there, each subpackage contains a directory named
+``src/<package_name>/``. Within each ``src/<package_name>/``, code is further
+arranged into categories, such as ``gamma`` and ``mudensity``. These correspond
+to Python modules. Finally, code within these categorised directories is
+organised into *levels*. Levels define the dependency hierarchy of code within
```
SimonBiggs: Levels are no longer being defined or exposed in any way within the file structure. Instead, the levelling is achieved apart from the file structure in the graphviz graphs....

docs/developer/file-structure.rst
```diff
@@ -225,15 +193,39 @@ dependency hierarchy of code within modules. See diagram below:
...
-the suite of automated tests for that particular subpackage. Any code present
-in ``src/<subpackage>/`` should be covered by tests in this directory.
-Automated testing is essential for effective `continuous integration`_, which
-is a core development philosophy of PyMedPhys. If you would like to make
-meaningful contributions to PyMedPhys - and become a much better developer as a
-result - it pays to get very familiar with automated testing and the code
-within these directories.
+If you look within the main ``pymedphys`` package, you won't see this same
+structure. Code is separated into modules, just like the subpackages, but there
+is no levelling, nor is there any actual library code. The modules within
```
SimonBiggs: No levelling visible in the file structure within the sub packages either

docs/developer/file-structure.rst
```diff
@@ -225,15 +193,39 @@ dependency hierarchy of code within modules. See diagram below:
...
+
+```python
+from pymedphys.dicom import anonymise_dataset
+```
+
+in their python file's list of imports. To the user, it appears that
+``anonymise_dataset`` is just a function inside of a module called ``dicom``
+that is contained in the main ``pymedphys`` package. However, under the hood,
+``anonymise_dataset`` is actually a function in a file called ``anonymise.py``
+in level 2 of the ``pymedphys_dicom`` subpackage, or:
```
SimonBiggs: No longer level 2

docs/developer/file-structure.rst
```diff
@@ -225,15 +193,39 @@ dependency hierarchy of code within modules. See diagram below:
...
+``pymedphys/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/_level2/anonymise.py``
```
SimonBiggs: No longer at this path.

Centrus007:
I should've paid more attention, sorry.

Wow, big changes!

Centrus007:
> The main comments here are based on the fact that when the dependency tree graphs came the "level" directories left.
> 
> ...the code dependency structure / levels is no longer visible via the directory layout. In fact, I don't think the discussion of dependencies has a place in the discussion about file structure any longer. ... It could however move over to the graphs page?
> 
> I hope this vastly simplifies where to put new code etc while not losing any of the benefits of having well levelled dependency structure. The visibility of the dependency structure that was provided for the levels is now achieved through the dependency graphs, and I believe it is now achieved in a way that is more effective that previously...
> 
> ...my hope is it is now less complicated, while simultaneously being more robustly levelled...

Fair enough. I suspect you're right on all counts :)

SimonBiggs:
> Wow, big changes!

Yup :). And I think it is now quite a bit simpler, while achieving the transparent goal of levelling even better... Which is a big win :)

PR #305, opened 2019-05-18
[RFR] React app

SimonBiggs:


SimonBiggs:
So, no effort has been made to make it look pretty. ... but it is functional!

There is no user feedback (like loading bars and the like) which is really sorely needed. But, you can select a trf file from your computer, and it decodes it, returning two csv files, one for the header, and one for the table.

See https://deploy-preview-305--app-pymedphys.netlify.com/

SimonBiggs:
This is ready for review. Now that the app website is separate, and no one is currently using it, I believe this is okay to merge into master.

SimonBiggs:
> It's a start! :)

A start it is :)

JavaScript and Python together feels a little bit like squeezing a square peg into a round hole
...

... Just now, I just realised a neat solution rxjs and rxpy... That will solve it I think... Brain wave. Awesome :).

The issue that is being solved is async programming. JavaScript is asynchronous forcibly... Trying to time appropriately the python code, and the results, is not currently neat... But with the Rx libraries I think we might get somewhere :). Awesome.

PR #309, opened 2019-05-20
[RFR] trf2dcm

SimonBiggs:


SimonBiggs:
There's a lot of very smelly code in here...

I've been rushing... and accruing technical debt.

I'm okay for this to merge, but I do need to write tests for the trf to delivery data, and delivery data to dicom.

Not for today though.

SimonBiggs:
@Centrus007 let's close this one. The code is crapola. I had forgotten just how specific and hardcoded the Jupyter notebook was that produced the abstract figures.

... Yesterday I then went, alright, short on time, let's see if I just copy this notebook code across into the library... it should work right? ... Nope. Really nope.

PR #310, opened 2019-05-20
[RFR] Interconversion between Logfiles and DICOM, as well as MU Density on both

SimonBiggs:
@Centrus007, so, this likely won't work with your DICOM files. But, for the purpose of demonstrating the procedure, and showing that it is possible this is a good start.

Now I am just going to be iterating on the app a little, so feel free to review and merge this when you get the chance.


SimonBiggs:
Due to being super keen to be able to link this I think I am going to force merge...

PR #343, opened 2019-05-23
[RFR] Fix Docs

SimonBiggs:


SimonBiggs:
Fails now appear to be because the npm package server is having issues.

SimonBiggs:
@Centrus007 npm server is no longer having issues, so these tests are passing now.

SimonBiggs:
@Centrus007 this one is worth a merge also....

PR #346, opened 2019-05-23
[RFR] Make multi fraction groups work, add `from_logfile()` method, implement some @Centrus007 suggestions

SimonBiggs:
@Centrus007 I'm making this pull request merge into the other pull request so that the new diff is clearly visible.


packages/pymedphys_deliverydata/src/pymedphys_deliverydata/object/exportable.py
```diff
@@ -106,7 +111,8 @@ def metersets(self, gantry_angles, gantry_tolerance):
         return get_metersets_from_delivery_data(
             self.mask_by_gantry(gantry_angles, gantry_tolerance))
 
-    def mudensity(self, gantry_angles=None, gantry_tolerance=None, grid_resolution=1):
+    def mudensity(self, gantry_angles=None, gantry_tolerance=None,
+                  grid_resolution=1, output_always_list=False):
```
SimonBiggs: @Centrus007 is this an okay compromise?

PR #355, opened 2019-05-25
[MRG] Prep for ICCR conference

SimonBiggs:
@Centrus007 so I haven't yet done any work to get the tests to pass / propagate the internal API changes to the rest of the code...

...but, I think this is far, far, far neater... and I think (now) you might approve...

It's worth having a look at in its current form, reviewing, making any naming/API changes you can think of, before I do the work of propagating the new internal API.

Also, I think we need three tags:

* [WIP] -- don't bother even looking at this pull request
* [RFR] -- worth reviewing, but don't merge
* [MRG] -- ready for merge when reviews and tests pass


SimonBiggs:
NOTE TO SELF: I was over zealous in my deletion. Need to restore testing data.

SimonBiggs:
NOTE TO SELF: remove the gantry angle told detection logic. Just set it to 3 by default and instead have a check that no defined gantry angles a less than twice the tolerance apart.

Centrus007:
Reviewing on my phone - looking absolutely gorgeous, old boy!

Centrus007:
I like the third PR tag of MRG idea. Let's roll with that

packages/pymedphys/src/pymedphys/deliverydata.py
```diff
@@ -1,4 +1,25 @@
...
+
+try:
+    from pymedphys_mudensity.deliverydata import DeliveryDataMuDensity
+    to_be_combined += [DeliveryDataMuDensity]
+except ImportError:
+    pass
+
+
+class DeliveryData(*to_be_combined[::-1]):  # type: ignore
+    pass
```
SimonBiggs: I should move this onto the root import. I think that should be fine for this package... We'll see.

packages/pymedphys_base/src/pymedphys_base/deliverydata/exportable.py
```diff
@@ -0,0 +1,226 @@
...
+
+# You should have received a copy of the Apache-2.0 along with this
+# program. If not, see <http://www.apache.org/licenses/LICENSE-2.0>.
+
+
+from collections import namedtuple
+import functools
+from typing import Union, Tuple
+
+import numpy as np
```
SimonBiggs: There might be value in refactoring numpy out of this package... Given it's going to be right at the bottom of the dependency tree... But then again, maybe there's no harm in always requiring numpy, especially now that numpy works just fine with pyodide... Not sure if there is an environment we want to target that doesn't have numpy...

Will we ever want the opportunity to have packages out of pymedphys be on hardware using something like https://micropython.org/? ... I think we would like to not be locked out of the option...

...okay, maybe it is best at some point to refactor numpy out of here.

Centrus007: For our user base, I would say that `numpy` has a level of ubiquity on par with the standard library. 
But if you want to cater to microcontroller programming (which is probably strictly in scope, but by a very fine line), I won't object to it.

SimonBiggs: > For our user base, I would say that `numpy` has a level of ubiquity on par with the standard library.
> But if you want to cater to microcontroller programming (which is probably strictly in scope, but by a very fine line), I won't object to it.

Let's not cater to it for now. On the day we need it (if we do), we can progressively refactor out the code that is usable into a separate package. Much the same as what we did for pyodide.

Centrus007: :+1:

SimonBiggs:
@Centrus007 this is ready for merge once tests pass :)

packages/pymedphys/src/pymedphys/deliverydata.py
```diff
@@ -1,4 +1,25 @@
...
+to_be_combined = [DeliveryDataBase]
+
+try:
+    from pymedphys_dicom.deliverydata import DeliveryDataDicom
+    to_be_combined += [DeliveryDataDicom]
+except ImportError:
+    pass
+
+try:
+    from pymedphys_fileformats.deliverydata import DeliveryDataLogfile
```
Centrus007: Perhaps not something to worry about yet, but would this need further stratification due to multiple log file vendors and major versions? E.g.
- `ElektaLog`
- `VarianTrajectoryLog`
- `VarianDynalog`

SimonBiggs: Very possibly. This is part of our internal API however, so we can just change that when it is needed.

packages/pymedphys_databases/src/pymedphys_databases/deliverydata/core.py
```diff
@@ -0,0 +1,7 @@
+from pymedphys_base.deliverydata import DeliveryData as DeliveryDataBase
+from pymedphys_fileformats.deliverydata import DeliveryDataLogfile
+from pymedphys_mudensity.deliverydata import DeliveryDataMuDensity
+
+
+class DeliveryDataDatabases(DeliveryDataMuDensity, DeliveryDataLogfile, DeliveryDataBase):  # type: ignore
```
SimonBiggs: There are quite a few methods that belong on here. ... But that's not a job for this pull request.

SimonBiggs: ... the name is awkward also ... Will fix that at some point.

SimonBiggs:
Yayy passes :)... and the tree is kinda neat:

image

packages/pymedphys/src/pymedphys/deliverydata.py
Centrus007: No worries

app/src/python/user-scripts/split-fraction-groups.py
```diff
@@ -16,10 +16,16 @@
     print("Loading DICOM file")
     dicom_dataset = pydicom.dcmread(filepath, force=True)
```
Centrus007: What do you think about wrapping `pydicom.dcmread()` into a modality specific DICOM reader function? Something like:

```python

MODALITIES = ["RTPLAN", "RTDOSE", "CT", ...]

def read_dicom_by_modality(filepath, modality):
    if not modality in MODALITIES:
        raise ValueError("The supplied modality is not supported") 
    
    ds = pydicom.dcmread(filepath)

    if not ds.Modality == modality:
        raise AttributeError("The supplied DICOM file's modality is not {modality}")

    return ds

Comment (Centrus007) [reply app/src/python/user-scripts/split-fraction-groups.py]: It would help mitigate the need for `try/except`s like the one in line 19, since the DICOM Standard itself enforces required tags (e.g. all valid RTPLANs have a `FractionGroupSequence`)

Comment (Centrus007) [reply app/src/python/user-scripts/split-fraction-groups.py]: (Not necessarily in this PR)

Comment (SimonBiggs) [reply app/src/python/user-scripts/split-fraction-groups.py]: That seems a little bit like look before you leap...

**Comment by SimonBiggs** [comment]
@Centrus007 So, I kept working in this branch. It's no longer ready for merge. But reviewing is still worthwhile.

File: packages/pymedphys_dicom/src/pymedphys_dicom/dicom/dose.py [hunk 6]
```diff
@@ -42,6 +42,15 @@
 # pylint: disable=C0103
 
 
+def zyx_and_dose_from_dicom(dicom_filepath):
+    ds = pydicom.dcmread(dicom_filepath, force=True)
```
SimonBiggs: @Centrus007 the new coordinate functions happened to break `gamma_dicom`. It was unfortunate the the testing file had x and y coords of the same length

SimonBiggs:
This here, is a very cool button:

image

app/src/python/tests/test_user_scripts.py
```diff
@@ -33,8 +33,8 @@ def test_gamma_calc(temp_workingdir):
     runpy.run_path(scriptpath)
 
 
-# def test_mu_density_diff(temp_workingdir):
-#     os.chdir(temp_workingdir)
-#     scriptpath = os.path.join(USER_SCRIPTS, 'mu-density-diff.py')
+def test_mu_density_diff(temp_workingdir):
+    os.chdir(temp_workingdir)
```
SimonBiggs: Now running a "test", where two of the user-scripts are run to see if they don't raise an error.

app/src/python/user-scripts/gamma.py
```diff
@@ -30,17 +33,107 @@
...
 
 plt.title("Gamma Percent Pass: {0:.2f} %".format(pass_ratio*100))
 
-plt.show()
+plt.savefig(os.path.join(output_dir, "gamma_histogram.png"))
+
+
+# (z_ref, y_ref, x_ref), dose_reference = zyx_and_dose_from_dataset(reference)
+# (z_eval, y_eval, x_eval), dose_evaluation = zyx_and_dose_from_dataset(evaluation)
+
```
SimonBiggs: I set about plotting all the slices, but without using `random_subset`, it's just way too slow in the browser.

app/src/python/user-scripts/options.tsx
```diff
@@ -12,20 +13,24 @@ export interface IUserScript {
 
 export const USER_SCRIPTS: IUserScript[] = [
     {
-        name: "trf2csv", code: trf2csv,
-        description: "Convert files that end in `.trf` into header and table `.csv` files"
+        name: "Anonymise DICOM", code: anonymiseDicom,
+        description: "Anonymises DICOM headers"
     },
```
SimonBiggs: I do think anonymise should be up top

changelog.md
```diff
@@ -66,6 +66,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - DICOM: `'d'` or `'DICOM'`
   - IEC fixed: `'f'`, `'fixed'` or `'IEC fixed'`
   - IEC patient: `'p'`, `'patient'` or `'IEC patient'`
+- `gamma_dicom` now take datasets as opposed to filenames
```
SimonBiggs: Made a change to the user API

docs/user/examples/gamma_from_dicom.ipynb
```diff
@@ -17,45 +17,25 @@
    "source": [
     "import os\n",
     "import warnings\n",
-    "import urllib.request\n",
+    "from glob import glob\n",
+    "import zipfile\n",
     "\n",
```
SimonBiggs: Updated the calc_gamma example notebook

SimonBiggs:
Let's get this one in. It works, if all else fails this could be used as it is in here on the day.

Centrus007:
Did you say the `gamma` wasn't working? If so, shall we chop it off into another branch?

SimonBiggs:
> Did you say the `gamma` wasn't working? If so, shall we chop it off into another branch?

It certainly is working, it just only outputs a histogram at the moment. :)

app/src/python/user-scripts/gamma.py
Centrus007: I see - the idea being you don't want to use `random_subset` for plotting?

Could use `gamma_filter` instead (for demo purposes at least)?

SimonBiggs: Fair point. Could use filter, as long as I only present pass or fail pixels...

SimonBiggs: Won't change that in time for the workshop however, I mean we could make a second script there that uses filter instead of shell...

SimonBiggs: Depending on whether or not you had the chance, you might be able to try give it a spin? Make a second Gamma DICOM script named "Gamma DICOM filter method"...

app/src/python/user-scripts/split-fraction-groups.py
Centrus007: Sort of, but not quite (according to my interpretation of the LBYL vs. EAFP debate). Let me explain:

If you have a function that fails due to an expected error (and failures are low probability), it is more efficient to have:

```python
ds = pydicom.dcmread(filepath)
try:
    rt_plan_specific_function(ds)
except not_an_rt_plan:
    handle_error()
```

vs.:

```python
ds = pydicom.dcmread(filepath)
if is_rt_plan:
    rt_plan_specific_function(ds)
else:
    handle_error()
```

because in the latter case, you are executing "checking code" (a conditional statement) in all scenarios, but in the former, extra code is only executed in (low probability) failure scenarios.

That's the essence of preferring EAFP over LBYL.

However, if you now consider:

```python
ds_plan = load_rt_plan(filepath)
rt_plan_specific_function(ds_plan)
```

vs.

```python
ds = pydicom.dcmread(filepath)
try:
    rt_plan_specific_function(ds)
except not_an_rt_plan:
    handle_error()
```

The former executes extra code, but only *once*, with built in safety and the luxury of being able to avoid the plethora of `try`/`except`s that would be required for *any* RT Plan specific function executed on a generic `ds`. To me, the trade off is worth it in fewer lines of code and greater security.

Also note that if we implement an RT Plan *object* we are essentially doing this anyway...

SimonBiggs: Sure, so is it applicable here? My interpretation of one of the benefits of EAFP is that it partially goes hand in hand with duck typing. It explicitly allows more flexibility. If for example I wanted to write a test where I just tested for this case, I could create just a fraction group and nothing else. I provide to the function only what it needs, and nothing more. And the test raises an exception only if it doesn't have what it needs, and nothing more. So as long as something sufficiently quacks like a duck it is allowed to enter into the function. ... Even if it was a completely different object, even a namedtuple should be allowed in, as long as it has the fraction group attribute.

There are other huge benefits such as avoiding race conditions. But that's not really relevant in this case.

SimonBiggs: I do think though, if we create an RT Plan object, it certainly should have contract with the user. "You supply me something that has these things, and I'll make you an RT plan object."

... I'd probably prefer that an RT plan constructor only required non-duplicate information. I would argue that if fraction group, and all the rest are there, then that information needs to be passed to create the RTPlan object, and to require the other information is to require duplicate information. If conflicting information is provided however such as it looks and quacks like an RTPlan object, but claims to be a dose object, we should raise an error on instantiation. So, in that point, I agree with implementing something "like" this for an RTPlan object. ... Key point is though one should be able to make the object by leaving rendundent information out if they choose. But certainly if redundant information is provided it should raise an error if it is internally inconsistent.

Another piece of rendundent information is CW or CCW etc. The DICOM standard is full of duplicate data.

SimonBiggs: The dicom object itself should always output the full spec however... So if it is created by leaving some items out, the output should fill those in via inference.

Centrus007: The `try`/`except` would be moved from the Fraction Group code block to the DICOM reading code block. The potential advantage is if you do other things that are RT Plan specific with that DICOM dataset, you avoid more `try`/`except` blocks. You also avoid issues with *forgetting* `try`/`except` blocks.

I might be wrong, but I don't think duck typing really affects the EAFP vs. LBYL debate. You can code your `if`/`else` with the same scope as your `try`/`except`. I think LBYL is perfectly compatible with duck typing. But your duck typing point does apply to the broader question of whether or not to implement the function, and there I think we have a difference in coding philosophy, which I think amounts to:

- You prefer to keep code as general as possible (in an effort to stay DRY).
- I prefer to keep code more tightly scoped so that all edge cases are handled.

The latter might be more pythonic, but that doesn't necessarily make it the right philosophy for *all* scenarios for PyMedPhys. In my opinion, insisting on keeping functions as general as possible when they are clearly designed for a specific object introduces many unknown unknowns, and a heck of a lot more work/careful thought to ensure known edge cases are handled.

Anyway, let's either discuss in a separate issue, or in person tomorrow :)

PR #373, opened 2019-06-04
[MRG] 0.8.x to be merged

SimonBiggs:


Centrus007:
Why the failures?

SimonBiggs:
> Why the failures?

Looks like something changed in the notebook dependency. It was part of an old attempt at a jupyter based GUI. It wasn't needed, so stripped that out and re released.

PR #378, opened 2019-06-05
[WIP] Pedro Martinez Labs Work

peterg1t:
Hi Simon,

Here is my initial commit were I just created a folder with my name under pymedphys/packages/pymedphys_labs/src/pymedphys_labs/ and added a copy of the qc-jaws script. I still need to clean the code a little bit but it is mostly complete. Maybe it is not the right place to place it but it works to start getting more familiar with pymedphys.

Cheers,
Pedro


packages/pymedphys_labs/src/pymedphys_labs/pedromartinez/qc-jaws.py
```diff
@@ -0,0 +1,1464 @@
...
+    # print(dataset.pixel_array)
+    # # exit(0)
+    # # plot the image using matplotlib
+    # plt.imshow(dataset.pixel_array, cmap=plt.cm.bone)
+    # plt.xlabel('X Distance from reference point')
+    # plt.ylabel('Y Distance from reference point')
+    # plt.show(block=True)
+
+
+try:
```
SimonBiggs: From here down needs to be wrapped in a function called `main()` and then at the very bottom of the code you'll need the following:

```python
if __name__ == "__main__":
    main()
```

See an example within https://github.com/pymedphys/pymedphys/blob/master/scripts/test-watch.py

This is so that when tools load up your Python file just to look inside they're not forced into running your script.

packages/pymedphys_labs/src/pymedphys_labs/pedromartinez/qc-jaws.py
```diff
@@ -0,0 +1,1464 @@
...
+#   generates a pdf file with a summary of all the results.
+#
+#   Example usage: python qc-jaws "/folder/"
+#
+#   Author: Pedro Martinez
+#   pedro.enrique.83@gmail.com
+#   5877000722
+#   Date:2019-04-09
+#
+###########################################################################################
```
SimonBiggs: It's worth at this point seeing whether or not you have a preference of licensing. If you wish to use the same license as the rest of PyMedPhys, put the following header on your file:

```python
# Copyright (C) 2019 Pedro Martinez

# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU Affero General Public License as published
# by the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version (the "AGPL-3.0+").

# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
# GNU Affero General Public License and the additional terms for more
# details.

# You should have received a copy of the GNU Affero General Public License
# along with this program. If not, see <http://www.gnu.org/licenses/>.

# ADDITIONAL TERMS are also included as allowed by Section 7 of the GNU
# Affero General Public License. These additional terms are Sections 1, 5,
# 6, 7, 8, and 9 from the Apache License, Version 2.0 (the "Apache-2.0")
# where all references to the definition "License" are instead defined to
# mean the AGPL-3.0+.

# You should have received a copy of the Apache-2.0 along with this
# program. If not, see <http://www.apache.org/licenses/LICENSE-2.0>.
```

But, you are free to license your code with any license that is compatible with the AGPL https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses

SimonBiggs: For more info on licensing see https://docs.pymedphys.com/getting-started/licensing.html

SimonBiggs:
Also, once you've finished with this pull request feel free to create another pull request to add yourself to the team over at https://github.com/pymedphys/pymedphys/blob/master/README.rst

peterg1t:
Hi Simon,
I fixed the same branch and recommitted the changes. Hope this is ok, I didn't wanted to create a new branch but I can do it ifit is better to track the changes. I added the license and the def main() to the code and also added my affiliation. If you have some insights or guide on how to make a test that would be awesome.  I'm still working in the documentation.
Cheers
Pedro

SimonBiggs:
All in this branch is perfect.

SimonBiggs:
So to work out where to go next, the first failing pipeline I look at is "pylint-all". There's a huge amount of output in that test but only three lines matter:

Screenshot_20190607-064756

See lines 1130-1132. The CI takes a fresh install of PyMedPhys, using the dependencies it has been told about, and tries to see if all the code makes sense.

The reason for those first two lines is it hasn't been told to install tdqm and fitz. That last error is actually a bug not related to us, so let's ignore that one for now.

To get those two packages to be installed I have a command that looks through all the code and updates the dependencies. To run it, run `yarn tree` within the top level pymedphys directory. You'll need to have yarn installed for that command to work. See https://docs.pymedphys.com/developer/contributing.html#install-contributor-system-dependencies on how to get that if you haven't already.

Once you've run yarn tree, commit, and then push again and let's see where it's up to.

SimonBiggs:
To begin with the set of tests there's a few approaches. I find as programmers, naturally as we go, we manually test our code by running parts of it, seeing the results etc, "do they make sense". The best tests actually come by hooking into that process, instead of testing manually, spending a bit more time to get those manual tests we are undergoing and repeating programmatically defined. So a good first test is to remember back to what was something you were repeatedly doing on the side to check it was working appropriately while you were writing the code?

Another way to approach it is to initially focus on the most pertinent lines of code. In fact, to make the code easier to read there might be significant value in splitting the code up into a few files. For example there are the core functions that actually achieve what you have written, there are output display creation functions, and there are UI functions guiding the user through running the script. Having a file "core", another "reporting", and another "cli" might be quite benefitial. 

From there, we would just focus on testing core. And from there, for the first test to write I would ask, what lines of code/pieces of logic are the most important. Potentially soon those lines of code into their own function, then write a test that imports that function provides it some inputs, and asserts that the output agrees with what it should be given that input.

If you chose simple enough inputs and outputs such that the reader of the code can recognise that that output should occur that really helps.

peterg1t:
Hi Simon,

I got yarn tree running but ran into this issue (see enclosed). I'm not sure what it means. I did run just yarn on the top level pymedphys directory and it seemed to resolve all the dependencies (please see the second pic). A dependency manager is something that to be honest I have never used. 

I can remove both the tqdm dependency (it is mainly a visual thing anyway to measure the progress of the script) and the fitz one from the script. I was using fitz to append the hospital logo into the pdf output but it is not necessary, so it can go away too. As per your suggestion I will split the code placing the main functions in separate files within the same folder. I'll get back at you with the main components to test.

image

image

Cheers,
Pedro

SimonBiggs:
Ahh, you'll need to run `yarn bootstrap` first as per those contributor
docs.

SimonBiggs:
Those dependencies are fine to have. Tqdm is especially nice. I don't mind
depending on it at all.

peterg1t:
Ran successfully the bootstrap!, however still yarn tree is giving me another problem.

image

SimonBiggs:
Are you running Python 3.7?

peterg1t:
Yep, I upgraded to 3.7. I can setup an environment for a previous version if needed.

SimonBiggs:
I just ran `yarn tree` myself and have pushed the change. We can try get to the bottom of what was happening on your machine at a later point I think.

You'll just need to do a `git pull` now in your branch to get my additional commit.

peterg1t:
Done git pull :)! Now that I think of it I have python3.7 setup to run under command python3. python2.7 is still coexisting in my linux box and it is what it comes when I just type python. Maybe this is the cause of the issue. I'll see if I changing the path does the trick.

packages/pymedphys_labs/src/pymedphys_labs/pedromartinez/qc-jaws.py
```diff
@@ -0,0 +1,1072 @@
...
+    for i in range(dim_len):
+        if N % 2 == 0:
+            a, b = i - (N - 1) // 2, i + (N - 1) // 2 + 2
+        else:
+            a, b = i - (N - 1) // 2, i + (N - 1) // 2 + 1
+
+        # cap indices to min and max indices
+        a = max(0, a)
+        b = min(dim_len, b)
+        out[i] = np.mean(x[a:b])
```
SimonBiggs: Let's add the following comment to this line to ignore that pylint error until the respective bug fix comes from numpy.

```suggestion
        out[i] = np.mean(x[a:b])  # pylint: disable=unsupported-assignment-operation
```

SimonBiggs: If you're happy with that change just press the following button :)

image

SimonBiggs:
Hmm, and maybe we can remove the `fitz` dependency. It doesn't appear to be so squeaky clean on the inside:

image

peterg1t:
Yeah I can definitely remove that!

SimonBiggs:
(`yarn tree` will be needed once a dependency is no longer needed at all by a package)

SimonBiggs:
I'm going to put some work into moving the yarn tree side of things around a bit. Make this process a bit tidier.

README.rst
```diff
@@ -157,6 +157,8 @@ along with their affiliation:
 
 .. _`University of Calgary`: http://www.ucalgary.ca/
 
+.. _`Tom Baker Cancer Centre`: https://www.ahs.ca/tbcc
```
SimonBiggs: The documentation testing produced the following error:

image

Meaning, you had referenced "Tom Baker Cancer Centre" within your name, but didn't actually provide a URL for it. Is this URL okay?

peterg1t: Hi Simon, 
Sorry for the late reply. It's that time of the month when I have to do some monthly QA on the linacs and finish really late. That URL works great but if it is a problem we can remove it and just keep the u of c one. I have been trying to run yarn as superuser on my linux box but that didn't do the trick. I tried "yarn run tree" but it also gave me the same result as "yarn tree".

SimonBiggs: No need to be sorry. :). The URL isn't a problem at all. It was just erroring out because it wasn't included. You can include both logos if you would like also. I think we'll need to squish down the space the logos are taking up soon by shrinking them down and putting multiple on the same line. But that's a good thing.

PR #395, opened 2019-06-11
[MRG] Expose efacts

SimonBiggs:


packages/pymedphys/src/pymedphys/electronfactors.py
```diff
@@ -1,5 +1,9 @@
 from pymedphys_electronfactors.electronfactors import (
```
Centrus007: Nice to see you’re starting to fully embrace my import formatting style!

If I were being super picky, I’d say you should alphabetise as well... ;)

SimonBiggs:
Hehe :)

PR #402, opened 2019-06-25
[MRG] Pinnacle Export Tool

pchlap:
Imported current version of the Pinnacle to DICOM export tool. In the end I decided to put it straight into its own package, since I wanted to expose it on the CLI from the start, it didn't seem appropriate to have it in labs.

The next steps are to add some documentation on how to use, as well as add some more automated tests to allow for contributions moving forward. But I'm curious to hear what you think so far!


SimonBiggs:
@Matthew-Jennings I've requested your review here to see if you have any comments

SimonBiggs:
@darcymason not sure how keen you might be to review PyMedPhys code, up to you. But this is certainly up your alley if you wanted to give it a squiz. No pressure though :).

SimonBiggs:
Also, @pchlap make sure to add your name, institution, and logo to the front page. Put yourself third on the list.

notebooks/pinnacle/ExportSample.ipynb
```diff
@@ -0,0 +1,235 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "import tarfile\n",
+    "import shutil\n",
+    "\n",
+    "from pymedphys_pinnacle import Pinnacle"
```
SimonBiggs: Expose this object through the top level `pymedphys` package and then get users to simply run `from PyMedPhys import Pinnacle`

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/exporttool/pinnacle.py
```diff
@@ -0,0 +1,158 @@
...
+from .rtstruct import convert_struct
+from .rtplan import convert_plan
+from .rtdose import convert_dose
+from .image import convert_image
+
+
+# This class holds all information relating to the Pinnacle Dataset
+# Can be used to export various elements of the dataset, Images,
+# RTSTRUCT, RTPLAN or RTDOSE.
+class Pinnacle:
```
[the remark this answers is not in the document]
SimonBiggs: I think this might be better named `PinnacleExport` so as not to clash with other pinnacle tools that will be exported from PyMedPhys.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/exporttool/pinnacle.py
```diff
@@ -0,0 +1,158 @@
...
+            im_info = im.get_image_info()[0]
+            im_suid = im_info['SeriesUID']
+            if len(series_uid) > 0 and im_suid == series_uid:
+                convert_image(im, export_path)
+                break
+
+        if image:
+            convert_image(image, export_path)
+
+    def print_images(self):
```
SimonBiggs: Would this be better as `log_images`?

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/exporttool/pinnacle.py
```diff
@@ -0,0 +1,158 @@
...
+            self.get_patient_info()
+
+            self.images = []
+            for image in self.patient_info['ImageSetList']:
+                pi = PinnacleImage(self, self.path, image)
+                self.images.append(pi)
+
+        return self.images
+
+    def export_struct(self, plan, export_path="."):
```
SimonBiggs: If we change the class to `PinnacleExport` I imagine the `export` could be removed from here.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/exporttool/pinnacle.py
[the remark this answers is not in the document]
SimonBiggs: Also, could this object be default return pydicom datasets instead of going straight to disk? That means one doesn't need to go on and off the disk if they want to programmatically use the DICOM files themselves.

In the future I would imagine just a single function `pinnacle_dicom_export` that accepts a path and returns a `DicomCollection` object that contains all of the DICOM objects contained at that pinnacle location.

What do you think?

Certainly doesn't have to be like that for this pull request though. But worth thinking about the API

pchlap:
Thanks @SimonBiggs for the great comments. Really very useful. I'll work my way through them today and I'll let you know how I proceed.

SimonBiggs:
I had an "uhuh" moment with the API. What if, instead of importing pinnacle the user does the following for example:

```python
from pymedphys import DicomImage, DicomStructure

images = DicomImage.from_pinnacle('a_path')
structures = DicomStructure.from_pinnacle('a_path')
```

This would be building on the following skeleton objects:

https://github.com/pymedphys/pymedphys/blob/master/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/collection.py#L155-L169

You may feel free to add methods to those DICOM objects, they are free pickings.

SimonBiggs:
> Thanks @SimonBiggs for the great comments. Really very useful. I'll work my way through them today and I'll let you know how I proceed.

My pleasure :) great to be coding with you :)!

SimonBiggs:
To make the API be through the DICOM objects you would need to invoke a `@classmethod` decorator. Much like the following:

https://github.com/pymedphys/pymedphys/blob/master/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/collection.py#L76-L83

pchlap:
> I had an "uhuh" moment with the API. What if, instead of importing pinnacle, the user does the following:
> 
> ```python
> from pymedphys import DicomImage, DicomStructure
> 
> dcm_image = DicomImage.from_pinnacle('a_path')
> dcm_struct = DicomStructure.from_pinnacle('a_path')
> ```
> 
> If they wanted to save the dicom files they could then do the following:
> 
> ```python
> dcm_image.to_file('a_file_path')  # already implemented
> ```
> 
> This would be building on the following skeleton objects:
> 
> https://github.com/pymedphys/pymedphys/blob/master/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/collection.py#L155-L169
> 
> You may feel free to add methods to those DICOM objects, they are free pickings.
> 
> @pchlap what do you think? That to me does feel like quite a natural API.

I do think that would feel like a natural API. However the main reason I currently run everything through that "Pinnacle" object is because parsing some of the raw files can be time consuming. So using that object instance I can just parse the files once in there and then use it to export the different modalities as needed. By simply passing the raw data path to DicomImage, DicomStructure, etc ... it would have to reparse these raw files each time.

SimonBiggs:
> I do think that would feel like a natural API. However the main reason I currently run everything through that "Pinnacle" object is because parsing some of the raw files can be time consuming. So using that object instance I can just parse the files once in there and then use it to export the different modalities as needed. By simply passing the raw data path to DicomImage, DicomStructure, etc ... it would have to reparse these raw files each time.

There might be a way around that using the LRU cache (https://docs.python.org/3/library/functools.html#functools.lru_cache). If the parse function receives an immutable string, and returns an immutable object, and is decorated by `@lru_cache` then if that particular string has been provided to that function before it won't rerun the calculation, instead it will just output the previous result.

pchlap:
> > I do think that would feel like a natural API. However the main reason I currently run everything through that "Pinnacle" object is because parsing some of the raw files can be time consuming. So using that object instance I can just parse the files once in there and then use it to export the different modalities as needed. By simply passing the raw data path to DicomImage, DicomStructure, etc ... it would have to reparse these raw files each time.
> 
> There might be a way around that using the LRU cache (https://docs.python.org/3/library/functools.html#functools.lru_cache). If the parse function receives an immutable string (the file contents), returns an immutable object (the parsed object, maybe as a `named_tuple`), and has no side effects, it can be decorated by `@lru_cache`. Then if that particular string has been provided to that function before it won't rerun the calculation, instead it will just output the previous result.
> 
> And then maybe the process will be something like the following:
> 
> ```python
> from pymedphys import DicomStructure
> 
> with open('a_path', 'r') as a_file:
>     contents = a_file.read()  # or however is best to load the raw file data into memory
> 
> # internally this will call the lru cached parse function
> dcm_struct = DicomStructure.from_pinnacle(contents)  
> ```

That's very cool. I think something along those lines will work. But I think we'll have to leave that for a future version.

Matthew-Jennings:
Thanks so much, @pchlap! Awesome to have such a quality contribution!

I've had a cursory look through and I'd echo most of Simon's comments. I'll look through more thoroughly when I can - though many comments I'd have need not necessarily be addressed in this PR.

I think as a minimum I'd like to see end-to-end tests before a merge. Then if we iterate on structure, names, etc., we can do so with much more confidence.

I'd also like to try to stick to some minimum level of documentation  for anything exposed in the library. There are a few comments you already have that could be converted to docstrings to help with this :)

SimonBiggs:
> That's very cool. I think something along those lines will work. But I think we'll have to leave that for a future version.

Yup, I'd be happy to leave that for a future version.

Matthew-Jennings:
It's not as clear to me that a DICOM object mediated `from_pinnacle()` method approach is the best way forward. Maybe it is, but at least one complication arises from the fact that a Pinnacle .tar file may contain a variety of DICOM modality combinations that the user likely doesn't necessarily understand in advance. This might be mediated through `DicomCollection`, but even then the user will often only want to save a subset of the DICOMs present in the archive.

Examples include:
- Pinnacle .tar file don't have to be patient scoped
- .tar files can contain multiple CTs, MRs, structure sets, plans and doses (of different types) for the same and different patients.

Nothing insurmountable, but I think the above does kinda imply mediation through DICOM may not be the best design...

SimonBiggs:
> It's not as clear to me that a DICOM object mediated `from_pinnacle()` method approach is the best way forward. Maybe it is, but at least one complication arises from the fact that a Pinnacle .tar file may contain a variety of DICOM modality combinations that the user likely doesn't necessarily understand in advance. This might be mediated through `DicomCollection`, but even then the user will often only want to save a subset of the DICOMs present in the archive.
> 
> Examples include:
> 
> * Pinnacle .tar file don't have to be patient scoped
> * .tar files can contain multiple CTs, MRs, structure sets, plans and doses (of different types) for the same and different patients.
> 
> Nothing insurmountable, but I think the above does kinda imply mediation through DICOM may not be the best design...

Makes sense. I think what will really help is solve this for other TPSs also and see if a natural interface falls out...

SimonBiggs:
@Matthew-Jennings @pchlap further thought makes me think something like a `Planning` object might be pertinent. It's API might look something like:

```python
from pymedphys import Planning

plan_data = Planning.from_pinnacle(something)
dcm_struct = plan_data.to_dicom_struct()

other_plan_data = Planning.from_monaco(something)
dcm_dose = other_plan_data.to_dicom_dose()
```

Matthew-Jennings:
So I do think an eventual implementation of this ought to be able to cater to any Pinnacle supported .tar file arrangement. This would mean parsing the .tar file into an arbitrary number of patients, each with an arbitrary number of DICOM-mappable and non-DICOM mappable data structures. Here a `Planning` object might be useful to house non-DICOM mappable data alongside DICOM, but note that it still doesn't alone solve the .tar file flexibility problem. Mind you, I think an object that maps to the tar file is unnecessary.

Phil's current implementation appears patient-scoped, which is common for Pinnacle users, I believe. But I'm nearly certain that this not forced, nor universal. Generalizing would be nice, but certainly not for this PR. And Phil, please note that I'm not taking anything away from your work! It's a lovely, useful and welcome addition! You should see how much we've iterated - and plan to iterate - on other code in PyMedPhys!

darcymason:
> @darcymason not sure how keen you might be to review PyMedPhys code, up to you. But this is certainly up your alley if you wanted to give it a squiz. No pressure though :).
> 
> Please don't feel obligated at all.

Just noticed my @ mention now - I'll see if I can catch up and take a look.  I should start getting to know my way around pymedphys better.

Matthew-Jennings:
I like the `Planning` object. But what do you think about calling it `Patient` or `TpsPatient` or similar (I prefer the latter)?

`Planning` feels a little... too focussed on plans, and isn't obviously scoped to a patient (which I *think* is what we want) `Tps` makes the application/analogous data clear and `Patient` the scope clear...

SimonBiggs:
> I like the `Planning` object. But what do you think about calling it `Patient` or `TpsPatient` or similar (I prefer the latter)?
> 
> `Planning` feels a little... too focussed on plans, and isn't obviously scoped to a patient (which I _think_ is what we want) `Tps` makes the application/analogous data clear and `Patient` the scope clear...

Issue is, you said the Pinnacle files weren't even necessarily patient scoped...

SimonBiggs:
> Just noticed my @ mention now - I'll see if I can catch up and take a look. I should start getting to know my way around pymedphys better.

Thanks @darcymason :)

Matthew-Jennings:
@SimonBiggs, that's right, but I'm not really imagining this object to match the .tar file scope. Pinnacle internally arranges patient-related data into patient-scoped structures. Tar files (archives) can just have multiple of these. It houses machine models too, but these can conceivably be considered patient data as well. They are really included to supplement patient plan info, as opposed to sharing machine models to be used for *new* plans

I'm imagining a function for which a Pinnacle .tar file is the input and a list/tuple of `PinnaclePatient` objects is the output. Each `PinnaclePatient` could house both its corresponding DICOM data and non-DICOM data present in the .tar file.

The purpose of the object is simply to house any extra TPS- and patient-specic, non-DICOM data.

Thing is, I can't imagine any scenarios where a user might want to manipulate this extra data outside of Pinnacle anyway...

Matthew-Jennings:
Which is a roundabout way of saying, if I were to write this from scratch, it would be functional (😮) and generate a list/tuple of DICOMs and ignore extra info in the first instance.

SimonBiggs:
> Which is a roundabout way of saying, if I were to write this from scratch, it would be functional (😮) and generate a list/tuple of DICOMs and ignore extra info in the first instance.

😎 🎉

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/exporttool/export_pinnacle.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
pchlap: @SimonBiggs So you might have to help clarify how to handle the licenses exactly. It's my understanding that all code in this repository is licensed under the AGPL-3.0+? So I would place the common header in my files with our copyright?

To make it a little more complicated, the original code I worked off was under a MIT license. That's why I've left that license file in the package. Can these two exist in there together? I know you have studied this licensing issue extensively so I will follow your advice on this one! :)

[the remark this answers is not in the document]
SimonBiggs: They certainly may coexist. Any file that contains work from the original MIT license, include your copyright with your license, and then say something like, "this work is derived from work released under the following license" then include the MIT license header with the original copyright. 

You are not forced to use the AGPL for your code. You may use any license that is compatible with the GPL (MIT included) https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses

It is certainly far easier (and my preference) if your release the code under the same as the rest of the library. But if you have a good reason to choose a different (and compatible) license you are free to do so.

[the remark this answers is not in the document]
pchlap: Fantastic, thanks for clarifying. Yes I don't see any reason not to put this under that AGPL as well, so I will take your suggested approach, and reference the original MIT license in the header :)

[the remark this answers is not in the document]
SimonBiggs: Glad to help :).

SimonBiggs:
> Which is a roundabout way of saying: if I were to write this from scratch, it would be a simple function () that reads a .tar file and generates a list/tuple of DICOM objects (probably `DicomStudy`s?) and ignores extra info in the first instance.

I believe an output of DicomCollection would probably be the aim here. I believe with DicomCollection we envisioned allowing it to be multiple patients. Quite literally a bag of DicomStudy's. One thought some time ago is its base representation would be a pandas dataframe, even though under the hood it would have a lot more in it. That was before the thoughts about going immutable though.

SimonBiggs:
... but maybe it really doesn't make sense for DicomCollection to be immutable. Anyway, that's off topic.

Matthew-Jennings:
Ah I see! Sorry, I'd wrong assumed that `DicomCollection` was patient-scoped. Do we then need a DicomPatient object as well?

Most TPSs nest planning data according to:

Patient --> Case / Examination / study --> Plan --> Beam Set / Trial (Which can also map to Plan - there's no equivalent in DICOM, IIUC)

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/exporttool/pinnacle_plan.py
```diff
@@ -0,0 +1,234 @@
...
+    plan_inst_uid = None  # UID for RTPlan instance
+    dose_inst_uid = None  # UID for RTDose instance
+    struct_inst_uid = None  # UID for RTStruct instance
+
+    def __init__(self, pinnacle, path, plan):
+        self.pinnacle = pinnacle
+        self.path = path
+        self.plan_info = plan
+        self.logger = pinnacle.logger
+
```
Matthew-Jennings: Excellent. Resolves my question earlier.

pchlap:
@SimonBiggs  @Matthew-Jennings  Thanks to both of you for a really great review of the code with some excellent thoughts and comments. This is really a great demonstration of the advantage of a repository like this (something you've both been telling me I know, but it's another thing to experience it first hand).

At this stage I'm working my way through your comments, focusing on the items that I think are essential for this pull request. Once we agree all agree on merging this pull request, I'll look at pulling out the other points into separate issues to work on moving forward.

SimonBiggs:
> At this stage I'm working my way through your comments, focusing on the items that I think are essential for this pull request. Once we agree all agree on merging this pull request, I'll look at pulling out the other points into separate issues to work on moving forward.

Great plan

packages/pymedphys_pinnacle/tests/test_pinnacle.py
```diff
@@ -0,0 +1,92 @@
...
+    from pymedphys_pinnacle import Pinnacle
+    return Pinnacle(archive, None)
+
+
+def test_pinnacle(pinn):
+    plans = pinn.get_plans()
+    assert len(plans) == 3
+
+
+def test_dose(pinn):
```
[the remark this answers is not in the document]
pchlap: During development, I use a script which simply outputs all of the differences between the two DICOM files. However I don't think this is useful for tests at this stage, since there are going to be a lot of differences. For now we just want to make sure the stuff we're interested in is matching I think. Also for the contour data, the sequence of points might be different, but that doesn't mean its necessarily wrong.

packages/pymedphys_pinnacle/tests/test_pinnacle.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Make sure the maximum values are in the same locations
+    assert exported_vol.argmax() == pinn_vol.argmax()
+
+    # Get the absolute difference between these volumes
+    diff = exported_vol - pinn_vol
+    abs_diff = numpy.absolute(diff)
+
+    # Make sure the absolute difference is (close to) zero
+    for val in numpy.nditer(abs_diff):
```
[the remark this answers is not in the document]
pchlap: Great, thanks! :)

packages/pymedphys_pinnacle/tests/test_pinnacle.py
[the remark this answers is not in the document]
SimonBiggs: A straight equality check will be quite difficult.

You've probably already thought of something like this, but might there be value in creating a mask based off of the contours and then verifying that the masks agree for each named contour?

SimonBiggs:
@pchlap thanks by the way for going through the review process :). It can be a bit of a back and forward. Thank you.

packages/pymedphys_pinnacle/tests/test_pinnacle.py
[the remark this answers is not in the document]
pchlap: Yep, that's exactly the approach I was planning to take... So... next question, is there a function in pymedphys which will generate a mask from an RTStruct for me? :-P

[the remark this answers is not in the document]
SimonBiggs: Something like this :)

https://github.com/pymedphys/pymedphys/blob/master/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/dose.py#L187-L219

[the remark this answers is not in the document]
SimonBiggs: It's not exactly right...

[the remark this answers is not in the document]
SimonBiggs: That masks a dose grid. In reality the "masking" feature should be refactored into it's own function.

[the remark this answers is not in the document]
pchlap: Awesome, yeah that's pretty close to what I'm after! :)

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/exporttool/pinnacle_image.py
```diff
@@ -1,3 +1,53 @@
...
+# Affero General Public License. These additional terms are Sections 1, 5,
+# 6, 7, 8, and 9 from the Apache License, Version 2.0 (the "Apache-2.0")
+# where all references to the definition "License" are instead defined to
+# mean the AGPL-3.0+.
+
+# You should have received a copy of the Apache-2.0 along with this
+# program. If not, see <http://www.apache.org/licenses/LICENSE-2.0>.
+
+# This work is derived from:
+# https://github.com/AndrewWAlexander/Pinnacle-tar-DICOM
```
SimonBiggs: Might even be helpful to include the commit hash in this link (and if you're keen the file path this particular code was derived from).

It's certainly not necessary, but it enables someone to go and use/compare to the original code if they wish.

Certainly not a must. Up to you.

SimonBiggs: For example https://github.com/pymedphys/pymedphys/blob/master/app/src/components/select-script.tsx#L27-L28

Matthew-Jennings:
Why is a straight equality check difficult? It's already implemented in `pydicom`...

Just to be clear, I'm suggesting setting a well-established baseline and running simple equality test for constancy.

packages/pymedphys_pinnacle/tests/test_pinnacle.py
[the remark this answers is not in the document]
Matthew-Jennings: Why is a straight equality check difficult? It's already implemented in `pydicom`.

Just to be clear, I'm suggesting setting a well-established baseline and running simple equality test for constancy.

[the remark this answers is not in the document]
SimonBiggs: In this case I don't think a baseline will be anywhere near as helpful as comparing to an available ground truth. Take a true DICOM export from Pinnacle and using that as the ground truth compare the created DICOMs to it. That will do the job of a baseline while also actually testing it works (not just testing that it is the same as it was before).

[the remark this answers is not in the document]
Matthew-Jennings: Agree that comparison with ground truth is most important.

But if for some reason full equality cannot or will not be achieved (e.g. missing info in the .tar file or even developer discretion), a baseline check is still valuable to protect against the effect of future code changes that could have adverse consequences outside of the ground truth test scope.

If full equality CAN be achieved against ground truth, then it is implicitly a full baseline check :)

My thoughts were: if @pchlap is happy with the performance of the converter from his own experience and manual testing, but doesn't have the auto tests written, perhaps a baseline test is a satisfactory minimum to include prior to merge?

But if you prefer the stronger constraint of automated tests verifying against ground truth with some minimum level of comprehensiveness, I certainly won't object :)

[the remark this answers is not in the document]
SimonBiggs: In this case the ground truth is readily accessible. And I would argue a complete equality with ground truth isn't required. For example certain headers can be left out. Order of things can be different as long as meaning is the same.

A forced complete baseline has the issue of forcing some things to stay the same which shouldn't need to stay the same. Which can be problematic. I think @pchlap approach here is ideal for this case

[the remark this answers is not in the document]
Matthew-Jennings: I disagree.

A completely test enforces nothing. It simply flags (possibly unintended) changes. A developer can simply alter the test to accommodate. Thus is best practice. A catch-all is prudent.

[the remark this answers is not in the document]
SimonBiggs: Okay... But should it catch that "GE Medical Systems" is hard coded into the header?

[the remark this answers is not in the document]
SimonBiggs: And should it catch that contour A is saved in the file before contour B?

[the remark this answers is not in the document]
SimonBiggs: Test specificity is important too. If tests often fail when no code logic is broken, then we will get "test failing fatigue". It's okay for tests to throw false negatives, but if from the outset we can catch just as many true positives with less false negatives I'm keen for that approach.

If a test is brittle, and fails because the order of something changed, and that happens on multiple tests throughout the code base, that hampers the efficiency of important refactoring work. Tests are supposed to make refactoring safer and more efficient. If the design of tests make refactoring painful, because say, "we don't want to run this for loop in a different order, as that will break 10 tests", we just leave the for loop as is.

Baselines are certainly important. I do agree. But it's baselines on the logic and meaningful outcomes. Things such as order, noise, non-determinism, non-logically relevant equalities should be kept out of the baselines as much as is plausible.

An example, if a swathe of tests relied on something such as dictionary key ordering, then when upgrading from Python 3.5 --> 3.6 all those tests would have then failed... (https://stackoverflow.com/a/39980744/3912576) ... which would have been very painful.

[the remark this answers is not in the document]
Matthew-Jennings: > Okay... But should it catch that "GE Medical Systems" is hard coded into the header?

Actually, yes, it absolutely should! If this has changed, we should be aware and adjust the test accordingly. I think the problem you envisage here is not solved by omitting it from testing, but by not hard coding it in the first place...

> And should it catch that contour A is saved in the file before contour B?

No, but that shouldn't be caught by the equality operator anyway.

I think we roughly agree on the endpoint, but differ on the means to get there. I agree that you shouldn't write tests to be too strict. You shouldnt require a certain order when order is known to be unimportant. My point is that you ONLY want to omit *known* false positives from your testing. I'm saying that one ought to test with maximum coverage and then narrow down from there. E.g. if your initial, very strictly written test fails for something like dict key order (it shouldnt anyway, but we'll use it as an example), then adjust your test to accommodate different key orders and iterate from there. That way, you are only ever accommodating known things, you are far less likely to miss unintended adverse changes as compared to the bottom-up, "only write unit tests for things you can think of" approach.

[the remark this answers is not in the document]
SimonBiggs: > That way, you are only ever accommodating known things, you are far less likely to miss unintended adverse changes as compared to the bottom-up, "only write unit tests for things you can think of" approach.

Fair point. I agree.

> I think the problem you envisage here is not solved by omitting it from testing, but by not hard coding it in the first place...

Also fair point.

Yup.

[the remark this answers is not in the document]
SimonBiggs: > No real objection to this approach; plenty of good sanity checks here! Though would a simple test of equality for your extracted DICOMs vs expected DICOMs be sufficient? Indeed perhaps superior, since all tags are checked?

@Matthew-Jennings for this request to use a straight equality to be useful to @pchlap I think you might have to provide a DICOM structure equality test example that still evaluates to `True` even if the structure slices are out of order, or if the structure points are defined anti-clockwise as opposed to clockwise for example. Essentially, if the structure definitions define the same volume this said equality check would need to evaluate to `True`.

I'm asking this... Because I think the request is non-trivial. And I believe that @pchlap current trajectory approach is sufficient.

SimonBiggs:
@Matthew-Jennings this appears to be your tests failing, yet @pchlap didn't actually make any changes to your code:

image

Are you able to have a peak and see what might have caused your tests to fail?

Cheers,
Simon

pchlap:
> @Matthew-Jennings this appears to be your tests failing, yet @pchlap didn't actually make any changes to your code:
> 
> image
> 
> Are you able to have a peak and see what might have caused your tests to fail?
> 
> Cheers,
> Simon

Yeah it's also not clear to me why that test is failing, especially because all tests are passing when I run them locally... I will continue to look into this :)

SimonBiggs:
@pchlap I wouldn't stress about that test just now. Let @Matthew-Jennings have a look see if he can get to the bottom of it.

pchlap:
Hi @SimonBiggs & @Matthew-Jennings 
As you can see I've been making some improvements based off your feedback. From my perspective I think the last thing to do for this pull request would be to add some docs. Is there a pymedphys module which is particularly well documented which I could work off? At this stage my plan is to add some docstrings as well as a page under 'docs/projects'? Would this be appropriate/sufficient?

Also, from your point of view is there anything else you would like me to address for this pull request?

SimonBiggs:
> Is there a pymedphys module which is particularly well documented which I could work off? 

Here is an example docstring:

https://github.com/pymedphys/pymedphys/blob/master/packages/pymedphys_mudensity/src/pymedphys_mudensity/mudensity/core.py#L43-L209

And here is the line within the docs that includes that docstring within the documentation:

https://github.com/pymedphys/pymedphys/blob/master/docs/user/library/mudensity.rst#L22

Which results in the following rendered documentation:

https://docs.pymedphys.com/user/library/mudensity.html#pymedphys.mudensity.calc_mu_density

Here is an example reference about how to write the docstrings:

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

SimonBiggs:
> At this stage my plan is to add some docstrings as well as a page under 'docs/projects'? Would this be appropriate/sufficient?

I would recommend to have a document included under https://docs.pymedphys.com/user/library/index.html much the same as the `mudensity` library.

I would also recommend inclusion within https://docs.pymedphys.com/user/interfaces/index.html#command-line-interfaces

I don't think `docs/projects` would be ideal.

SimonBiggs:
For documentation, could you also include an addition the changelog:

https://github.com/pymedphys/pymedphys/blob/master/changelog.md

And certainly add yourself to the README.rst file if you're comfortable with that:

https://github.com/pymedphys/pymedphys/blob/pc-pinnacle-export/README.rst

SimonBiggs:
And then lastly with the tests, it would be great to have that mask compare for the structures:

https://github.com/pymedphys/pymedphys/pull/402/files?file-filters%5B%5D=.py#diff-90818d5abdf138de42e3428f217380b6R187

And if the plan compare isn't yet functional:

https://github.com/pymedphys/pymedphys/pull/402/files?file-filters%5B%5D=.py#diff-90818d5abdf138de42e3428f217380b6R255

Would we be able to make the exposed API `PinnacleExport` raise a `NotImplementedError` https://docs.python.org/3/library/exceptions.html#NotImplementedError:

https://github.com/pymedphys/pymedphys/pull/402/files?file-filters%5B%5D=.py#diff-e033a2d701991a0859b1192749f5f360R167

Also make sure the CLI raises that same error:

https://github.com/pymedphys/pymedphys/pull/402/files?file-filters%5B%5D=.py#diff-e033a2d701991a0859b1192749f5f360R167

SimonBiggs:
As adamant as @Matthew-Jennings was about the baseline concept, I'm not convinced it is entirely practical here without better implementation of `__eq__`. So for now, I think comparing the structures via a mask and then removing the plan export from the public API should be sufficient.

SimonBiggs:
More than happy for you to disagree with me also @pchlap :) I'm no arbiter of truth :).

SimonBiggs:
Also, can I say, thank you @pchlap this is an awesome pull request. For me it is a huge encouragement.  I hope this can be the first of many. And I hope any pain points regarding contributing can be made easier.

packages/pymedphys_pinnacle/tests/test_pinnacle.py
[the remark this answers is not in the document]
Matthew-Jennings: I still disagree, pending evidence that such an implementation *is* required for this request. I.e. writing my suggested simple test and demonstrating that it would actually result in an unwieldy number of false positive failures. If written intelligently (yet still trivial), I’m not so sure it would... :)

Unfortunately I won’t be in a position to do this myself for at least another month!

That said, I don’t mean to suggest that I would wish to block this merge for this request (note my original comment)! Most of this is academic. Though I would prefer if my suggestion were at least *tried*, please feel free to proceed. :)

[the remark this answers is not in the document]
Matthew-Jennings: Remember I’m suggesting relative to baseline, not ground truth (so in *addition* to the existing tests)

SimonBiggs:
Also, following the "Find Bugs Once" method, can we include a test that covers the case of the offending DICOM orientation referred to by @Matthew-Jennings 

> Find Bugs Once
> Once a human tester finds a bug, it should be the last time a human tester finds that bug. Automatic tests should check for it from then on.

From https://pragprog.com/the-pragmatic-programmer/extracts/tips

pchlap:
@SimonBiggs I'm having some trouble getting my head around compiling and running the docs. I think the info here is possibly outdated: https://docs.pymedphys.com/developer/documentation.html ? I'd be happy to update that except I'm not sure the correct way to do it.

My approach has been to run:
`yarn pip:install:docs`
`yarn docs`

This didn't seem to compile the html itself, so before running `yarn docs` I had to run `make html` inside the docs directory. This at least generated some html but it's not updating as I would expect as I update my docstrings etc... Could you point me in the right direction to get this running? Thanks :)

SimonBiggs:
> @SimonBiggs I'm having some trouble getting my head around compiling and running the docs. I think the info here is possibly outdated: [docs.pymedphys.com/developer/documentation.html](https://docs.pymedphys.com/developer/documentation.html) ? I'd be happy to update that except I'm not sure the correct way to do it.
> 
> My approach has been to run:
> `yarn pip:install:docs`
> `yarn docs`
> 
> This didn't seem to compile the html itself, so before running `yarn docs` I had to run `make html` inside the docs directory. This at least generated some html but it's not updating as I would expect as I update my docstrings etc... Could you point me in the right direction to get this running? Thanks :)

Yes you're right that is out of date. And you are also right on the second account, `yarn pip:install:docs` then `yarn docs` is what you need to run. Unfortunately, changing docstrings doesn't yet trigger a rebuild or a difference.

A workflow when updating docstrings would be to have a little script that deletes `docs/_build`, reruns `sphinx-build -W docs docs/_build/html`, and then runs `python -m http.server --directory docs/_build/html`. Ideally however, I would find a way to have docstrings invalidate the docs build cache and rerun the sphinx build automatically on file edit.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/pinnacle_plan.py
```diff
@@ -125,13 +125,7 @@ def trials(self):
             if isinstance(self._trials, dict):
                 self._trials = [self._trials['Trial']]
 
-            # Select trial with FINAL in name as active trial for this plan
-            # If no FINAL trial found then select first trial
-            for trial in self._trials:
-                if 'FINAL' in trial['Name'].upper():
```
SimonBiggs: Did you want this to be accessible via a parameter? It's quite important that you ["Eat your own dog food"](https://en.wikipedia.org/wiki/Eating_your_own_dog_food), so ideally the code you write here can be the code you end up using internally. Therefore, you would want it to be able to support your internal processes.

SimonBiggs: What I mean is have a parameter like `trial_name_selector` which defaults to `None` but allows you to adjust it to something like `FINAL`

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/constants.py
```diff
@@ -0,0 +1,98 @@
...
+
+RTDOSEModality = 'RTDOSE'
+RTPLANModality = 'RTPLAN'
+RTSTRUCTModality = 'RTSTRUCT'
+
+RTDoseSOPClassUID = '1.2.840.10008.5.1.4.1.1.481.2'
+RTStructSOPClassUID = '1.2.840.10008.5.1.4.1.1.481.3'
+RTPlanSOPClassUID = '1.2.840.10008.5.1.4.1.1.481.5'
+
+Manufacturer = 'Pinnacle Philips'
```
SimonBiggs: Potentially even Manufacturer here should by `PyMedPhys`

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/image.py
```diff
@@ -0,0 +1,268 @@
...
+import struct
+import shutil
+
+# Constants
+RTDOSEModality = 'RTDOSE'
+GTransferSyntaxUID = '1.2.840.10008.1.2'
+GImplementationClassUID = '1.2.826.0.1.3680043.8.498.75006884747854523615841001'
+RTDoseSOPClassUID = '1.2.840.10008.5.1.4.1.1.481.2'
+RTPlanSOPClassUID = '1.2.840.10008.5.1.4.1.1.481.5'
+Manufacturer = 'Pinnacle Philips'
```
SimonBiggs: Import these from `constants.py`?

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/image.py
```diff
@@ -0,0 +1,268 @@
...
+        ds.SpatialResolution = 0.35  # ???????
+        # ds.DistanceSourceToDetector = #???
+        # ds.DistanceSourceToPatient = #????
+        ds.GantryDetectorTilt = 0.0  # ??
+        ds.TableHeight = -158.0  # ??
+        ds.RotationDirection = "CW"  # ???
+        ds.ExposureTime = 1000  # ??
+        ds.XRayTubeCurrent = 398  # ??
+        ds.GeneratorPower = 48  # ??
+        ds.FocalSpots = 1.2  # ??
```
SimonBiggs: I do think these hard coded results need to be removed as @Matthew-Jennings mentioned in a previous comment.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/exporttool/image.py
```diff
@@ -0,0 +1,218 @@
...
+        #ds.SeriesInstanceUID = seriesuid
+        ds.FrameOfReferenceUID = info["FrameUID"]
+        ds.StudyInstanceUID = info["StudyInstanceUID"]
+        ds.SeriesInstanceUID = info["SeriesUID"]
+        # problem, some of these are repeated in image file so not sure
+        # what to do with that
+        ds.InstanceNumber = slicenum
+        ds.ImagePositionPatient = [-float(image_header["x_pixdim"])*10 * float(image_header["x_dim"]) / 2, -float(
+            image_header["y_pixdim"])*10 * float(image_header["y_dim"]) / 2, sliceloc]
+        if "HFS" in currentpatientposition or "FFS" in currentpatientposition:
```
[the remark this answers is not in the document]
SimonBiggs: @pchlap would you be able to have a second testing data set that compares the gold standard (planning system official export) to this tool for FFS and FFP data?

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/pinnacle_plan.py
pchlap: Perhaps in a future version, but I will script that functionality from where I call on the API and I would encourage others to do the same.

pchlap:
> I personally don't have Pinnacle, so all of my reviews are going to be relatively superficial. I think before the final merge it would be of huge value if we do wait until @Matthew-Jennings gets back from his A/L and takes it for a spin at his site.

While I agree that there are still a number of issues to resolve, and the cleanest option would definitely be to work with @Matthew-Jennings to fix these before merging, I have got a number groups who are eager to jump in and start using this, as is... No doubt with their input we will be able to identify further issues and solve them all together.

I am off to Germany for a month from tomorrow on, so my time will become quite limited. Ultimately these other groups will begin using the code whether or not we have merged into the master branch. I just thought it would be easier or them to get started if they could pip install pymedphys and then get started with the command line.

I don't want to pressure you guys to merge anything that you don't feel is ready for the master branch, so I would be happy to send out some instructions to the others on how they can get started using this branch :)

SimonBiggs:
> While I agree that there are still a number of issues to resolve, and the cleanest option would definitely be to work with @Matthew-Jennings to fix these before merging, I have got a number groups who are eager to jump in and start using this, as is... No doubt with their input we will be able to identify further issues and solve them all together.

That's okay. Given the time constraints, let's work on getting this merged in before you go with the intent to address issues or bugs in a future pull request.

SimonBiggs:
@pchlap was there anything else you wanted to do before a merge?

SimonBiggs:
The example renders quite nicely:

https://deploy-preview-402--docs-pymedphys.netlify.com/user/examples/pinnacle_export.html

SimonBiggs:
Noticed that the CLI isn't in the docs yet?

https://deploy-preview-402--docs-pymedphys.netlify.com/user/interfaces/index.html#command-line-interfaces

pchlap:
> @pchlap was there anything else you wanted to do before a merge?

Yes I'm just working on some docs :)

SimonBiggs:
Kk, let me know once you're ready for a merge, we'll pull it in, and then I'll cut the `pymedphys` `0.10.0` release so that it can be used from a standard pip install.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/pinnacle_plan.py
SimonBiggs: Makes sense

pchlap:
I have added some docs, so from my point of view I would be happy for this to merge so that others can begin experimenting with it. I'm sure over the next weeks a number of issues will be opened, as well as opening issues from the outstanding items in this pull request. Once I'm back I'll look at tackling all of those in the next iteration of development and hopefully that will result in a much more stable version of this tool :)

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+>>> from pymedphys.pinnacle import (
+...     PinnacleExport
+... )
+
+.. WARNING::
+   The DICOM objects exported by this tool are not the same as the DICOM
+   objects exported from within the Pinnacle application. Certain DICOM
+   modalities correspond fairly well to the Pinnacle exported objects
+   for some versions of Pinnacle. Others are not working as expected or
+   have not been validated.
```
SimonBiggs: This is a good warning. Thanks @pchlap.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/pinnacle_plan.py
```diff
@@ -219,16 +312,32 @@ def iso_center(self, iso_center):
         self._iso_center = iso_center
 
     def is_prefix_valid(self, prefix):
+        """Check is a UID prefix is valid.
```
SimonBiggs: "is" --> "if" here

docs/user/library/pinnacle.rst
```diff
@@ -0,0 +1,24 @@
...
+
+.. automodule:: pymedphys_pinnacle.export
+    :no-members:
+
+
+***
+API
+***
+
+.. autoclass:: pymedphys_pinnacle.export.PinnacleExport
```
SimonBiggs: From here reference `pymedphys.PinnacleExport` instead of `pymedphys_pinnacle`. The sub packages should be transparent to users, never a part of the public API.

docs/user/library/pinnacle.rst
```diff
@@ -0,0 +1,24 @@
...
+   :members:
+
+.. autoclass:: pymedphys_pinnacle.export.PinnaclePlan
+   :members:
+
+.. autoclass:: pymedphys_pinnacle.export.PinnacleImage
```
SimonBiggs: Here for the `PinnaclePlan` and `PinnacleImage` also.

docs/user/library/pinnacle.rst
SimonBiggs: @pchlap this is an easy fix, so I'll do this one, then merge it.

pchlap:
> One important point is that by referencing `pymedphys_pinnacle` instead of the top level `pymedphys` package with `autoclass` it has actually detailed within the docs that submodule import:
> 
> https://deploy-preview-402--docs-pymedphys.netlify.com/user/library/pinnacle.html#pymedphys_pinnacle.export.PinnacleExport
> 
> It's okay to use `pymedphys_pinnacle` with `automodule` with `no-members`, but you can't use it anywhere else. `pymedphys_pinnacle` needs to be completely hidden from the user as it is a part of the private API which we reserve the right to completely change at any release.
> 
> For example, we might change `pymedphys_pinnacle` to `pymedphys_tps`, and that needs to not break any user code.

OK great, yeah I understand.

I see you already made the changes, thanks a lot! :)

SimonBiggs:
> I see you already made the changes, thanks a lot! :)

One last thing before you leave today :) So that I can make a release. Can you update the following section within the following file:

https://github.com/pymedphys/pymedphys/blob/master/changelog.md#unreleased

That's what I use for the release notes on each GitHub release, and it details the changes for this release on the docs.

SimonBiggs:
@pchlap if you could give us a heads up if you're doing that change log, otherwise I'll write something.

pchlap:
> @pchlap if you could give us a heads up if you're doing that change log, otherwise I'll write something.

Yep just putting something in now :)

SimonBiggs:
> Yep just putting something in now :)

Awesome :) Thanks :)

changelog.md
```diff
@@ -35,6 +35,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Re-exposed `convert2_ratio_perim_area` and `create_transformed_mesh` from
   `pymedphys.electronfactors`.
+- Pinnacle module providing a tool to export raw Pinnacle data to DICOM
+  objects.
+  - A CLI is provided: See
+  [the Pinnacle CLI docs](../user/interfaces/cli/pinnacle.rst).
```
SimonBiggs: Does this link work in the built docs?

changelog.md
```diff
@@ -35,6 +35,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+  - As well as an API: See
+  [the Pinnacle library docs](../user/interfaces/library/pinnacle.rst).
```
SimonBiggs:
```suggestion
  [the Pinnacle library docs](https://docs.pymedphys.com/user/interfaces/library/pinnacle.html).
```

SimonBiggs:
Merged! :)

And, `pymedphys` version `0.10.0` is now on pypi if you want to take it for a spin :).

SimonBiggs:
Thanks for this epic pull request @pchlap :)

SimonBiggs:
It's also live in the docs:

https://docs.pymedphys.com/user/interfaces/cli/pinnacle.html

pchlap:
> Thanks for this epic pull request @pchlap :)

Thank you! Yep I can confirm the new version is installing from pypi! That's awesome.

This has been a great experience, look forward to more of this in the very near future.

In the meantime, I'm off to Europe for the next month. As a side note I will be attending EuroPython in Basel next week, so I'll report back on how that goes! Oh and my wedding of course :-P

SimonBiggs:
:) I hope your wedding goes absolutely brilliantly :).

Very keen to hear about EuroPython. I haven't actually yet attended a Python conference. Very keen to however.

Matthew-Jennings:
Thank you very, very much @pchlap!! Great stuff! I hope for and look forward to more in the future!

Enjoy Europe! I really hope your wedding goes magnificently!

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[0].RTBeamLimitingDeviceType = 'ASYMX'
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[1].RTBeamLimitingDeviceType = 'ASYMY'
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[2].RTBeamLimitingDeviceType = 'MLCX'
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[0].NumberOfLeafJawPairs = '1'
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[1].NumberOfLeafJawPairs = '1'
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[2].NumberOfLeafJawPairs = p_count/2
+                bounds = ['-200', '-190', '-180', '-170', '-160', '-150', '-140', '-130', '-120', '-110', '-100', '-95', '-90', '-85', '-80', '-75', '-70', '-65', '-60', '-55', '-50', '-45', '-40', '-35', '-30', '-25', '-20', '-15', '-10', '-5', '0', '5', '10', '15', '20', '25', '30', '35', '40', '45', '50', '55', '60', '65', '70', '75', '80', '85', '90', '95', '100', '110', '120', '130', '140', '150', '160', '170', '180', '190', '200']
```
SimonBiggs: So much of this file contains hard coded linac specific parameters. To work around this for the logfile --> DICOM conversion code I required the user to provide a DICOM template. I think all of the hard coded parameters in this file should be removed, and instead the function should take a template DICOM file which the user provides. Then, the new DICOM file will just be that template DICOM file with the pinnacle results mapped onto it.

See the logic over at https://github.com/pymedphys/pymedphys/blob/master/packages/pymedphys_dicom/src/pymedphys_dicom/delivery/core.py#L157-L186.

I think the first clean up pass through for this code needs to be the removal of any hard coded values.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+
+    metersetweight = ['0']
+
+    num_fractions = 0
+    beam_count = 0
+    beam_list = trial_info['BeamList'] if trial_info['BeamList'] else []
+    if len(beam_list) == 0:
+        plan.logger.warning("No Beams found in Trial. Unable to generate RTPLAN.")
+        return None
+    for beam in beam_list:
```
SimonBiggs: Here do:

```python
for beam_index, beam in enumerate(beam_list):
    do_stuff()
```

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+
+        beam_count = beam_count + 1
```
SimonBiggs: replace this with `beam_index` defined above

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+
+        plan.logger.info("Exporting Plan for beam: " + beam['Name'])
+
+        ds.PatientSetupSequence.append(Dataset())
```
SimonBiggs: Instead of appending this at the start and then having to reference it with `ds.PatientSetupSequence[beam_count - 1]` call it `patient_setup` and append it once it is built.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+        ds.PatientSetupSequence[beam_count - 1].PatientPosition = patient_position
+        ds.PatientSetupSequence[beam_count - 1].PatientSetupNumber = beam_count
+
+        ds.FractionGroupSequence[0].ReferencedBeamSequence.append(Dataset())
+        ds.FractionGroupSequence[0].ReferencedBeamSequence[beam_count - 1].ReferencedBeamNumber = beam_count
+        ds.BeamSequence.append(Dataset())
```
SimonBiggs: Instead of appending this at the start and then having to reference it with `ds.BeamSequence[beam_count - 1]` call it `beam` and append it once it is built. Or, even better, build it with `dicom_dataset_from_dict` https://github.com/pymedphys/pymedphys/blob/master/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/create.py#L42-L68

Used like:

https://github.com/pymedphys/pymedphys/blob/master/packages/pymedphys_dicom/tests/test_from_dict.py#L39-L41

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+        ds.BeamSequence[beam_count - 1].BeamName = beam['FieldID']
+        ds.BeamSequence[beam_count - 1].BeamDescription = beam['Name']
+
+
+        if 'Photons' in beam['Modality']:
+            ds.BeamSequence[beam_count - 1].RadiationType = 'PHOTON'
+        elif 'Electrons' in beam['Modality']:
+            ds.BeamSequence[beam_count - 1].RadiationType = 'ELECTRON'
+        else:
+            ds.BeamSequence[beam_count - 1].RadiationType = ""
```
SimonBiggs: I would rather raise an exception here, if its a modality we are not sure about we should noisily tell the user that we are unsure about what to do.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+                wedgetype = "STANDARD"
+                wedgeflag = True
+                numwedges = 1
+                wedgeangle = cp['WedgeContext']['Angle']
+                wedgeinorout = ""
+                wedgeinorout = cp['WedgeContext']['Orientation']
+                if int(wedgeangle) == 15:
+                    numberinname = '30'
+                elif int(wedgeangle) == 45:
+                    numberinname = '20'
```
SimonBiggs: What's this for?

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+                elif int(wedgeangle) == 30:
+                    numberinname = '30'
+                elif int(wedgeangle) == 60:
```
SimonBiggs: Having these cases drawn out like this is quite brittle.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+                    wedgeorientation = '90'  # temporary until I find out what to put here
+                elif "WedgeLeftToRight" == wedgeinorout:
+                    wedgename = "W" + str(int(wedgeangle)) + "L" + numberinname  # + "U"
+                    wedgeorientation = '270'
+                elif "WedgeTopToBottom" == wedgeinorout:
+                    wedgename = "W" + str(int(wedgeangle)) + "OUT" + numberinname  # + "U"
+                    wedgeorientation = '180'  # temporary until I find out what to put here
+                elif "WedgeBottomToTop" == wedgeinorout:
+                    wedgename = "W" + str(int(wedgeangle)) + "IN" + numberinname  # + "U"
+                    wedgeorientation = '0'  # temporary until I find out what to put here
```
SimonBiggs: Would rather raise a `NotImplementedError` if we don't know what to do.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+                        ds.BeamSequence[beam_count - 1].ControlPointSequence[j].WedgePositionSequence.append(Dataset())
+                        ds.BeamSequence[beam_count - 1].ControlPointSequence[j].WedgePositionSequence[0].WedgePosition = "IN"
+                        ds.BeamSequence[beam_count - 1].ControlPointSequence[j].WedgePositionSequence[0].ReferencedWedgeNumber = '1'
+
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence.append(Dataset()) # This will be the x jaws
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence.append(Dataset()) # this will be the y jaws
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence[0].RTBeamLimitingDeviceType = 'ASYMX'
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence[0].LeafJawPositions = [x1, x2]
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence[1].RTBeamLimitingDeviceType = 'ASYMY'
```
SimonBiggs: Elekta don't have ASYMY (see template comment)

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+                        ds.BeamSequence[beam_count - 1].ControlPointSequence[j].WedgePositionSequence[0].ReferencedWedgeNumber = '1'
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence.append(Dataset()) # This will be the x jaws
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence.append(Dataset()) # this will be the y jaws
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence[0].RTBeamLimitingDeviceType = 'ASYMX'
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence[0].LeafJawPositions = [x1, x2]
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence[1].RTBeamLimitingDeviceType = 'ASYMY'
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence[1].LeafJawPositions = [y1, y2]
+                    ds.BeamSequence[beam_count - 1].ControlPointSequence[j].BeamLimitingDevicePositionSequence.append(Dataset()) # this will be the MLC
```
SimonBiggs: This repeated logic here needs to be factored out

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
```diff
@@ -0,0 +1,548 @@
...
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[0].RTBeamLimitingDeviceType = 'ASYMX'
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[1].RTBeamLimitingDeviceType = 'ASYMY'
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[2].RTBeamLimitingDeviceType = 'MLCX'
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[0].NumberOfLeafJawPairs = '1'
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[1].NumberOfLeafJawPairs = '1'
+                ds.BeamSequence[beam_count - 1].BeamLimitingDeviceSequence[2].NumberOfLeafJawPairs = p_count/2
+                bounds = ['-200', '-190', '-180', '-170', '-160', '-150', '-140', '-130', '-120', '-110', '-100', '-95', '-90', '-85', '-80', '-75', '-70', '-65', '-60', '-55', '-50', '-45', '-40', '-35', '-30', '-25', '-20', '-15', '-10', '-5', '0', '5', '10', '15', '20', '25', '30', '35', '40', '45', '50', '55', '60', '65', '70', '75', '80', '85', '90', '95', '100', '110', '120', '130', '140', '150', '160', '170', '180', '190', '200']
```
SimonBiggs: repeated logic again, but would likely be fixed by a template

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
SimonBiggs: In any case where there is any chance the code might be doing "the wrong thing" let's raise an error. I want this to fail fast and noisily if it is ever sees something that is unexpected.

packages/pymedphys_pinnacle/src/pymedphys_pinnacle/export/rtplan.py
SimonBiggs: Also these I believe are Varian MLC leaf bounds.

PR #406, opened 2019-07-01
[TEST] commit useless file

SimonBiggs:
This is a useless file to see if the tests are now failing


SimonBiggs:
@pchlap @Matthew-Jennings as you can see these tests fail even on a fresh clone of master.

SimonBiggs:
@Matthew-Jennings it appears that the part that is failing is the section where an item is removed from the global baseline. For some reason that global state trickery isn't working anymore.

Potentially a better way to test it is to just add a tag that isn't in the baseline dict?

PR #409, opened 2019-07-01
[MRG] Str(e) to str(e.value)

Matthew-Jennings:


SimonBiggs:
I feel a bit bad taking you away from your holiday...

SimonBiggs:
I still don't get what caused this...

SimonBiggs:
There we go, that's the difference:

pytest version 5.0.0:

Changelog:

https://docs.pytest.org/en/latest/changelog.html#removals

Relevant issue:

 https://github.com/pytest-dev/pytest/issues/5412

SimonBiggs:
@pchlap has fixed this issue. It was due to an upgrade in pytest to 5.0.0. see the above comment. This can now be merged into your branch.

Matthew-Jennings:
Yeah, I was thinking it had to be an update to a dependency. I was actually thinking Python itself, but hadn’t gotten around to checking the logs between passing and failing tests to interrogate package versions! Anyway, nice find :)

SimonBiggs:
Oh? You didn't find that? How did you fix it without finding that?

SimonBiggs:
Ahh, your prints :)

Matthew-Jennings:
I re-checked the `pytest.raises()` example code and noticed it had changed from asserting x in `str(excinfo)` to asserting x in `str(excinfo.value)`.

I hadn’t made the leap to recognising `pytest` had updated, I’d actually thought it was Python itself. I’d mistakenly thought `excinfo` Referred to Python Exception info, but now I see that was silly, it’s actually part of `pytest`.

SimonBiggs:
I'll let @pchlap do the merge given it's his branch

PR #414, opened 2019-07-04
[MRG] improve robustness of pcolor_grid creation

SimonBiggs:
Only need one review on this for it to be merged. @pchlap not sure how open you are to doing a review here and there. Let me know if you're willing.

### DIFF
--- a/packages/pymedphys_mudensity/src/pymedphys_mudensity/plt/helpers.py
+++ b/packages/pymedphys_mudensity/src/pymedphys_mudensity/plt/helpers.py
@@ -29,8 +29,10 @@ import numpy as np
 
 def pcolormesh_grid(x, y, grid_resolution=None):
     if grid_resolution is None:
-        grid_resolution = x[1] - x[0]
-        assert grid_resolution == y[1] - y[0]
+        diffs = np.hstack([np.diff(x), np.diff(y)])
+        assert np.all(np.abs(diffs - diffs[0]) < 10**-12)
+
+        grid_resolution = diffs[0]
 
     new_x = np.concatenate(
         [x - grid_resolution/2, [x[-1] + grid_resolution/2]])
--- a/changelog.md
+++ b/changelog.md
@@ -31,7 +31,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
-- nil
+### Improvements
+
+- Made the resolution detection of `pymedphys.plt.pcolormesh_grid` more robust.
 
 ## [0.10.0]
 


SimonBiggs:
By the way, just a primer on the pull request tags:

* [WIP] means "work in progress", usually not even worth a review
* [RFR] means "ready for review" but please don't merge yet
* [MRG] means please "merge" once tests and reviews pass -- as in the reviewer may do the merge.

packages/pymedphys_mudensity/src/pymedphys_mudensity/plt/helpers.py
```diff
@@ -29,8 +29,10 @@
 
 def pcolormesh_grid(x, y, grid_resolution=None):
     if grid_resolution is None:
-        grid_resolution = x[1] - x[0]
-        assert grid_resolution == y[1] - y[0]
+        diffs = np.hstack([np.diff(x), np.diff(y)])
+        assert np.all(np.abs(diffs - diffs[0]) < 10**-12)
```
Matthew-Jennings: Not at all required, but [`numpy.allclose`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html) is perhaps neater here?

SimonBiggs:
No need. I'll let him jump in at some point if he thinks it's appropriate. Otherwise let's just continue as normal :).

packages/pymedphys_mudensity/src/pymedphys_mudensity/plt/helpers.py
SimonBiggs: I thought about that, but it's actually not as neat. All close I believe requires two arrays to be the same size. This is an array and a scalar. I could be wrong though.

SimonBiggs:
Also, collapsing and merging has actually be confusing simultaneous commits causing some commits to occur twice. Let's not squish and merge any longer.

packages/pymedphys_mudensity/src/pymedphys_mudensity/plt/helpers.py
Matthew-Jennings: Numpy accepts a scalar in either position. It simply broadcasts.

I think that's the point of the "array or *array-like*" phrase present in many numpy parameter docstring sections - so long as it can broadcast, it's fine. Scalars should generally work, I think.

https://stackoverflow.com/questions/41363751/numpy-is-there-an-allclosenp-array-scalar

SimonBiggs:
Okay cool :). I don't mind leaving it as it is now, but I shall use all
close next time :).

Matthew-Jennings:
> Also, collapsing and merging has actually be confusing simultaneous commits causing some commits to occur twice. Let's not squish and merge any longer.

Really? Where? That's weird...

Squashing and committing is a very, very weak preference of mine. I like the cleaner history it provides ("readibility is king" and all that). And since we tend to keep our feature branches short anyway, I think the further granularity afforded by avoiding squashing is generally unnecessary.

I don't think duplicate commits are a consequence of squashing and merging. But I could be wrong.

Regardless, if this is a strong preference of yours, I'll avoid squashing from now on. :)

SimonBiggs:
It's happened when I have had a branch that was running longer, it took a while to get it reviewed, so I leave a branch as is and then fork that branch and continue it elsewhere in a new pull request with the aim that the original get merged and the new will have its first bunch of commits lopped off once the original gets merged to master.

For a while there it worked, and then it stopped working. I suspect it was due to squishing. Maybe an experiment is in order :).

packages/pymedphys_mudensity/src/pymedphys_mudensity/plt/helpers.py
SimonBiggs: Thanks for that @Matthew-Jennings I believe you're right :).

Issue #416, opened 2019-07-04
Documentation re-write


SimonBiggs:
Highly consider reworking PyMedPhys documentation based on the following advice:

https://www.divio.com/blog/documentation/

Found via link at: https://github.com/kylelobo/The-Documentation-Compendium

SimonBiggs:
From the blog:

The secret
Documentation needs to include and be structured around its four different functions: tutorials, how-to guides, explanation and technical reference. Each of them requires a distinct mode of writing. People working with software need these four different kinds of documentation at different times, in different circumstances - so software usually needs them all.

And documentation needs to be explicitly structured around them, and they all must be kept separate and distinct from each other.

TUTORIALS
A tutorial:

is learning-oriented
allows the newcomer to get started
is a lesson
Analogy: teaching a small child how to cook

HOW-TO GUIDES
A how-to guide:

is goal-oriented
shows how to solve a specific problem
is a series of steps
Analogy: a recipe in a cookery book

EXPLANATION
An explanation:

is understanding-oriented
explains
provides background and context
Analogy: an article on culinary social history

REFERENCE
A reference guide:

is information-oriented
describes the machinery
is accurate and complete
Analogy: a reference encyclopaedia article

This division makes it obvious to both author and reader what information goes where. It tells the author how to write, and what to write, and where to write it. It saves the author from wasting a great deal of time trying to wrestle the information they want to impart into a shape that makes sense, because each of these kinds of documentation has only one job.

In fact, it’s extremely hard to maintain good documentation that doesn’t implicitly or explicitly recognise the quadrants of this scheme. The demands of each kind are different from those of the others, so any attempt at documentation that fails to maintain this structure suffers, as it’s pulled in different directions at once.

Once you understand the structure, it becomes a very useful tool for analysing existing documentation, and understanding what needs to be done to improve it.

SimonBiggs:
I propose, right at the top level of the docs splitting up according to those categories. Hpw-to-guides need to be self contained. Meaning any data needs to be able to be downloaded by the notebook (or a PyMedPhys helper package).

We need to change our pattern from using test files in the how to guides as that relies on people using the guide needing to clone the repo. Which is not the right requirements for someone at the stage of a how-to guide.

SimonBiggs:
Hence, a Zenodo solution with a built in data downloader function in PyMedPhys would likely be our best bet to handle large files. (Far better than waiting for git VFS).

Pandas does this, and so does pylinac.

PR #425, opened 2019-07-18
[MRG] expose trf2pandas, add a few experimental notebook examples

SimonBiggs:


Matthew-Jennings:
`trf2pandas()` is neither documented nor tested. I do think we should start to be strict with this for the `master` branch of PyMedPhys proper. Postponing until some other time prior to next release is a sure-fire recipe for it never getting done!

Perhaps interested users can be given early, undocumented, untested access to functions like this in separate branches?

SimonBiggs:
`trf2pandas` certainly is tested. See https://github.com/pymedphys/pymedphys/blob/master/packages/pymedphys_fileformats/tests/trf/test_decode.py

Matthew-Jennings:
I see `trf2csv`, but no `trf2pandas`...

SimonBiggs:
trf2csv is just trf2pandas with pandas in built (tested) to_csv method.

SimonBiggs:
https://github.com/pymedphys/pymedphys/blob/master/packages/pymedphys_fileformats/src/pymedphys_fileformats/trf/trf2csv.py#L65-L68

Matthew-Jennings:
Ah yes, I see. My mistake.

But the documentation point still remains.

SimonBiggs:
> Ah yes, I see. My mistake.
> 
> But the documentation point still remains.

I agree with the docs. I had a read of the following https://www.divio.com/blog/documentation/

I noted it here https://github.com/pymedphys/pymedphys/issues/416

I am actually considering a complete re-work of the docs based on it. And I want to do that in unison with our next major API adjustment.

I have been talking to @msobolewski, I think I might be able to get some time in the nearish future to undergo this documentation re-write and fill out.

Once that is complete, I will expect that every exposed API will have a Reference, How-To, and Explanation (as defined in that linked article).

Until that time, lets not block things due to documentation. But the day when we block based on docs will come, and it will come before a 1.0.0 release.

Matthew-Jennings:
Okay :)

PR #440, opened 2019-07-26
[MRG]: Tps meas compare - a  helping hand

Matthew-Jennings:
Heya @SimonBiggs,

Just thought I'd give you a hand and take the liberty to make a few of my suggested changes that you agreed to. =)

It occurs to me that `require_patient_orientation_be_HFS()` is not necessarily specific to RTDOSE files, but I figured it was fine there for now. But if you think it's important to move it, we can :)


SimonBiggs:
Thanks @Matthew-Jennings :)

PR #442, opened 2019-07-27
[MRG]: Update Docs with Matthew-Jennings & Git configuration recommendations. Fix bleeding edge install instructions.

### LINKED ISSUE #192: Developer setup guidelines no longer work after the big split (#171)

Matthew-Jennings:
fixes #192

### DIFF
--- a/docs/getting-started/installation.rst
+++ b/docs/getting-started/installation.rst
@@ -46,20 +46,27 @@ Installing the "bleeding edge" version from GitHub
 --------------------------------------------------
 
 You can install the very latest, "bleeding edge", development version of
-PyMedPhys by directly cloning its GitHub repo and installing its dependencies
-using the following commands from within a terminal:
+PyMedPhys, but you will first need Git and Yarn. In turn, Yarn requires Nodejs.
+Perhaps the easiest way to install these is to install Chocolatey from within
+a terminal in Administrator mode using `these instructions <https://chocolatey.org/docs/installation>`__.
+Once installed, reload your terminal and run the following commands:
 
 .. code:: bash
 
-    git clone https://github.com/pymedphys/pymedphys.git
+    choco install git nodejs yarn
+
+If you already have Git installed, you can of course remove "``git``" from the
+above command.
+
+Once installation is complete, reload the terminal again and run the following
+commands to clone the ``pymedphys`` GitHub repo and install the ``pymedphys``
+package along with its dependencies:
+
+.. code:: bash
+
+    git clone git@github.com:pymedphys/pymedphys.git
+    # If you prefer HTTPS, use this link instead: https://github.com/pymedphys/pymedphys.git
     cd pymedphys
 
     conda config --add channels conda-forge
-    conda install pymedphys --only-deps
-    pip install -e .
-
-
-Note that this requires Git to be installed on your workstation. Instructions
-for installing Git can be found in the `Developer Guide`_.
-
-.. _`Developer Guide`: ../developer/contributing.html#chocolatey
+    yarn bootstrap
--- a/docs/getting-started/installation.rst
+++ b/docs/getting-started/installation.rst
@@ -46,27 +46,16 @@ Installing the "bleeding edge" version from GitHub
 --------------------------------------------------
 
 You can install the very latest, "bleeding edge", development version of
-PyMedPhys, but you will first need Git and Yarn. In turn, Yarn requires Nodejs.
-Perhaps the easiest way to install these is to install Chocolatey from within
-a terminal in Administrator mode using `these instructions <https://chocolatey.org/docs/installation>`__.
-Once installed, reload your terminal and run the following commands:
+PyMedPhys, but you will first need Git and Yarn. You can find instructions for
+installing Git `here <https://www.atlassian.com/git/tutorials/install-git>`__
+and Yarn `here <https://yarnpkg.com/en/docs/install>`__.
+
+Once you have Git and Yarn, run the following commands in a terminal to clone
+the ``pymedphys`` GitHub repo and install the ``pymedphys`` package along with
+its dependencies:
 
 .. code:: bash
 
-    choco install git nodejs yarn
-
-If you already have Git installed, you can of course remove "``git``" from the
-above command.
-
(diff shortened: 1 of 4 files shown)


SimonBiggs:
While you're in there are you able to fix the following also:

https://docs.pymedphys.com/getting-started/installation.html#installing-the-bleeding-edge-version-from-github

We have left it to still say `pip install -e .`

docs/getting-started/installation.rst
```diff
@@ -46,20 +46,27 @@ Installing the "bleeding edge" version from GitHub
...
+.. code:: bash
 
-Note that this requires Git to be installed on your workstation. Instructions
-for installing Git can be found in the `Developer Guide`_.
+    git clone git@github.com:pymedphys/pymedphys.git
+    # If you prefer HTTPS, use this link instead: https://github.com/pymedphys/pymedphys.git
+    cd pymedphys
 
-.. _`Developer Guide`: ../developer/contributing.html#chocolatey
+    conda config --add channels conda-forge
```
SimonBiggs: This shouldn't be needed...

packages/pymedphys/setup.py
```diff
@@ -93,7 +93,7 @@ def get_variable_from_file(filepath, variable):
             'xlwings >= 0.15.4'
         ],
         'lint': [
-            'autopep8',
+            'black',
```
SimonBiggs: We'll need to put this within the vscode settings also.

docs/getting-started/installation.rst
```diff
@@ -46,20 +46,27 @@ Installing the "bleeding edge" version from GitHub
 --------------------------------------------------
 
 You can install the very latest, "bleeding edge", development version of
-PyMedPhys by directly cloning its GitHub repo and installing its dependencies
-using the following commands from within a terminal:
+PyMedPhys, but you will first need Git and Yarn. In turn, Yarn requires Nodejs.
+Perhaps the easiest way to install these is to install Chocolatey from within
+a terminal in Administrator mode using `these instructions <https://chocolatey.org/docs/installation>`__.
```
SimonBiggs: Actually, maybe we should keep this one OS independent. If someone is wanting to install bleeding edge they can work some things out.

Potentially the following link should be sufficient:

https://yarnpkg.com/en/docs/install

docs/getting-started/installation.rst
```diff
@@ -46,20 +46,27 @@ Installing the "bleeding edge" version from GitHub
...
+Once installation is complete, reload the terminal again and run the following
+commands to clone the ``pymedphys`` GitHub repo and install the ``pymedphys``
+package along with its dependencies:
 
+.. code:: bash
 
-Note that this requires Git to be installed on your workstation. Instructions
-for installing Git can be found in the `Developer Guide`_.
+    git clone git@github.com:pymedphys/pymedphys.git
+    # If you prefer HTTPS, use this link instead: https://github.com/pymedphys/pymedphys.git
```
SimonBiggs: Probably best just to write https here, users can swap to ssh if they want to develop, but if they're just cloning https is far easier...

docs/getting-started/installation.rst
Matthew-Jennings: Good point!

docs/getting-started/installation.rst
Matthew-Jennings: No worries :)

packages/pymedphys/setup.py
Matthew-Jennings: Will do :)

Matthew-Jennings: I didn't worry because VS Code doesn't have `autopep8` explicitly enabled in `settings.json`. I assumed `black` wouldn't need it either, and that we basically weren't enforcing *autoformatting*.

Or do you want to use `black` (with the `--check`) option for linting as well?

SimonBiggs: I was thinking of doing the following:

https://code.visualstudio.com/docs/python/editing#_general-formatting-settings

python.formatting.provider = black

Matthew-Jennings: Added :)

docs/getting-started/installation.rst
```diff
@@ -46,20 +46,16 @@ Installing the "bleeding edge" version from GitHub
...
-    conda config --add channels conda-forge
-    conda install pymedphys --only-deps
-    pip install -e .
-
-
-Note that this requires Git to be installed on your workstation. Instructions
-for installing Git can be found in the `Developer Guide`_.
-
-.. _`Developer Guide`: ../developer/contributing.html#chocolatey
+    yarn bootstrap
```
SimonBiggs: And maybe a "see the developer guide for more detailed instructions on how to set up PyMedPhys for development".

Matthew-Jennings: Sure :) done!

docs/getting-started/installation.rst
```diff
@@ -46,20 +46,23 @@ Installing the "bleeding edge" version from GitHub
...
     git clone https://github.com/pymedphys/pymedphys.git
     cd pymedphys
+    yarn bootstrap
 
-    conda config --add channels conda-forge
-    conda install pymedphys --only-deps
-    pip install -e .
 
+Windows users can find detailed instructions for installing Git and Yarn in the
+_`Developer Guide`.
```
SimonBiggs:
```suggestion
`Developer Guide`_.
```

SimonBiggs:
:) thanks @Matthew-Jennings

PR #447, opened 2019-07-28
[MRG] Blacken PyMedPhys

### LINKED ISSUE #444: Blackifying codebase
Black aims to minimise diffs allowing reviewing to be easier and have the git blames be better assigned to who changed the line.

However, in the transition it will do exactly the opposite. And potentially cause conflicts in open pull requests. I vote, we get current major pull requests merged. Then blackify the whole code base before open a new pull request.

Either that, or iteratively blackify things and we will to go through a bit of review pain with excessive diffs initially...

Matthew-Jennings:
Resolves #444 

Black couldn't parse nearly 50 files, all of which were either in:

- `packages/pymedphys_sphinxtheme/node_modules/` or
- `node_modules/`

I think at least some of these files are python 2?

The first commit was achieved by running:

```bash
black . --exclude "^.*/node_modules/.*$
```

No other changes were made.


Matthew-Jennings:
Curiously, significantly more lines were added than removed, despite the increased line length.

I think it might be due to `black`'s convention of putting closing parentheses on new lines.

Matthew-Jennings:
So, `yarn tree` conflicts with `black` convention.

`black` triggered "file has changed", but `yarn tree` does not add a trailing comma in the lists contains in the `_install_requires.py` files (conflicting with `black`) and populates `version_info` with single-quote strings in the `_version.py` files (conflicting with `black`)

Can we configure `yarn tree` to conform to `black`? Or shall we just ignore those files when blackifying?

SimonBiggs:
Let's make yarn tree conform.

Matthew-Jennings:
> Let's make yarn tree conform.

I'm not sure how to do that. I had a bit of a look, but it wasn't clear to me. Shall we do that in this PR or later?

Matthew-Jennings:
Actually, I think I worked it out (messily, mind you). The relevant changes can be found [here](https://github.com/pymedphys/pymedphys/pull/447/commits/a765f3f54b55fbd698d4dbe946a3142565488e4d) in `packages/pymedphys_monomanage/src/pymedphys_monomanage/propagate/dependencies.py` and [here](https://github.com/pymedphys/pymedphys/pull/447/commits/7f707724faa8e4d325b7230d0b77ec972902471f) in `packages/pymedphys_monomanage/src/pymedphys_monomanage/propagate/versions.py`

SimonBiggs:
Broke things I think :) https://github.com/pymedphys/pymedphys/blob/a765f3f54b55fbd698d4dbe946a3142565488e4d/packages/pymedphys_sphinxtheme/src/pymedphys_sphinxtheme/_install_requires.py

SimonBiggs:
Tidier solution might be to literally call the black command over the strings produced.

Matthew-Jennings:
>Tidier solution might be to literally call the black command over the strings produced.

That is cleverer, yes!

SimonBiggs:
Much better :)

SimonBiggs:
I was also expecting that you'd use the python API not the CLI:

https://github.com/psf/black/issues/779

Interestingly because you chose the CLI version black itself is not being picked up as a dependent...

Matthew-Jennings:
> I was also expecting that you'd use the python API not the CLI:

Yeah, using the Python API is probably better. I honestly completely missed it.

SimonBiggs:
Given Black's API isn't stable, create a wrapper function adapted that when
passed a string will output that same string blackified. Then we just call
that one function everywhere.

That way if Black's API changes we can just update it in one place.

Matthew-Jennings:
> Given Black's API isn't stable, create a wrapper function adapted that when
> passed a string will output that same string blackified. Then we just call
> that one function everywhere.
> 
> That way if Black's API changes we can just update it in one place.

Okay, will do :) In which package do you think the wrapper is best placed? Just `pymedphys_monomanage` for now?

SimonBiggs:
Yeah monomanage seems appropriate

Matthew-Jennings:
Okay, we might be good now :)

.pylintrc
```diff
@@ -0,0 +1,2 @@
+disable=bad-continuation
+
```
SimonBiggs: Here is where we would also be setting he pylint line length to 88.

Matthew-Jennings:
@SimonBiggs, read for merge I *think* :)

Matthew-Jennings:
Which rulers should we keep/remove from here `.vscode/settings.json`, do you think?

```python
 "[python]": {
    "editor.insertSpaces": true,
    "editor.tabSize": 4,
    "editor.rulers": [
      72,
      79,
      119
    ]
```

.vscode/settings.json
```diff
@@ -13,8 +13,7 @@
   "[python]": {
     "editor.rulers": [
       72,
-      79,
-      119
+      88
```
SimonBiggs: Yup, that works well. 72 for docs/long strings. And the 88 is helpful.

Matthew-Jennings:
Did you want to merge it, old mate?

SimonBiggs:
Woohoo :) it passes. Merge it is.

PR #453, opened 2019-08-01
[MRG] Remove file-structure page from docs for now

Matthew-Jennings:


docs/developer/file-structure.rst
```diff
@@ -1,3 +1,5 @@
+:orphan:
```
SimonBiggs: Oooh, cool, I didn't know about this :)

Matthew-Jennings: Neither did I! Google is our friend :)

PR #459, opened 2019-08-15
[MRG] Do not treat group length as unknown tag in DICOM anonymisation

Matthew-Jennings:
So I came across a CT that contained a bunch of "Group Length" elements that were treated as unknown by our anonymisers. This triggered a few improvements in the first two commits here:

-   https://github.com/pymedphys/pymedphys/commit/e0eac3cb093f345e2b9571c64b10632d7948fcc4 improved printing of unknown tag information by including the hex tag number as well as the keyword. In my CT case, the keywords were empty strings and so it took some manual debugging to discover that the Group Length elements were the problem. Adding tag number helps here, I think.
- https://github.com/pymedphys/pymedphys/commit/f7f48ef9f6d9f77ba60f0262434f0d8eefcdde6a fixes another error I got where basically `numpy`'s `bool` doesn't always play nice with native Python `bool`.
- https://github.com/pymedphys/pymedphys/commit/342041a906a437e2016599af7ea645eb2a7deb92 addresses the idea that maybe Group Length elements shouldn't be considered "unknown". Strictly they are "known", just retired. [DICOM basically recommends that these elements be discarded](http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_7.2.html) and [`pydicom` itself actually does discard them, at least when writing to file](https://github.com/pydicom/pydicom/pull/595). We could just continue as is, since users can strictly just pass `ignore_unknown_tags=True` to the anonymisation functions. But since deletion of Group Lengths is usually happening anyway, maybe we ought not to treat them as "unknown"? It's debatable, and I don't have a strong opinion.

### DIFF
--- a/dependencies.json
+++ b/dependencies.json
@@ -148,6 +148,7 @@
         "json",
         "operator",
         "os",
+        "pprint",
         "re",
         "warnings"
       ]
--- a/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/anonymise.py
+++ b/packages/pymedphys_dicom/src/pymedphys_dicom/dicom/anonymise.py
@@ -164,7 +164,7 @@ def anonymise_dataset(
             "of date, please raise an issue on GitHub at "
             "https://github.com/pymedphys/pymedphys/issues.".format(
                 printer.pformat(unknown_tags_to_print)
-        )
+            )
         )
 
     elif delete_unknown_tags:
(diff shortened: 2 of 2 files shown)


SimonBiggs:
> * We could just continue as is, since users can strictly just pass `ignore_unknown_tags=True` to the anonymisation functions. But since deletion of Group Lengths is usually happening anyway, maybe we ought not to treat them as "unknown"? It's debatable, and I don't have a strong opinion.

Ahh it's okay I think this change is helpful. I think I understand the issue. By having them in there it will unnecessary prompt users to discard group lengths.

Maybe the neater solution is to submit a pull request to pydicom to have those tags ignored on load? ... I'm happy for us to handle it here as you have done, but maybe also see if it is a good idea to have them ignored on load within pydicom itself... I'll leave that to your judgement as I have only spent a few minutes thinking about (usually not the best way to make good decisions)

Issue #462, opened 2019-08-15
Move long running tests to a single pipeline


SimonBiggs:
Make long running tests only run on one pipeline on one OS.

Highlight all long running tests, mark them with "slow" and then have only the Linux pipeline run them. This will allow an improved machine turn around time for CI testing.

SimonBiggs:
https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option

SimonBiggs:
Filter the following to see the long running tests:

https://dev.azure.com/pymedphys/pymedphys/_build/results?buildId=3804&view=ms.vss-test-web.build-test-results-tab

PR #463, opened 2019-08-15
[MRG] Skip slow

SimonBiggs:
Skip all slow tests. Any test that took longer than 1 s is being skipped by default. Only the Linux pipeline will run these tests.

Fixes https://github.com/pymedphys/pymedphys/issues/462

@Matthew-Jennings and @pchlap some of these tests I am skipping are yours, so probably worth having both of your approvals before merge.

I imagine there might be some value in the future in speeding up these tests that have been marked as slow. But for now, still having them run, but just on one pipeline, I think is the way to go.

### DIFF
--- a/app/src/python/tests/test_user_scripts.py
+++ b/app/src/python/tests/test_user_scripts.py
@@ -27,6 +27,7 @@ def temp_workingdir(tmpdir_factory):
     return workingdir
 
 
+@pytest.mark.slow
 def test_gamma_calc(temp_workingdir):
     os.chdir(temp_workingdir)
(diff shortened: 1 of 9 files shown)


Matthew-Jennings:
I think I'm happy so long as they're run at least once on at least one platform. :)

SimonBiggs:
> I think I'm happy for my tests in particular to just run at least once on at least one platform. :)

Yup, that they are.

SimonBiggs:
I'm going to merge, it's easy enough to undo if @pchlap has any issues with it.

PR #464, opened 2019-08-19
[MRG] Windows Pipeline Broken (already!)

SimonBiggs:
...

Attempting fix in hope that this won't break so easily again... (no longer holding breath)


SimonBiggs:
Fixed the Windows pipeline.

PR #479, opened 2019-08-30
[MRG] Remove app

SimonBiggs:
I have already copied a frozen version of https://app.pymedphys.com over to https://liberapp.net. Also https://app.pymedphys.com now redirects to https://liberapp.net.

It is now safe to remove the app directory.

The app is now located at https://github.com/libreapp/libreapp/tree/master/legacy

[EDIT] ... except for the fact that some of the docs calls files in there. oops. fixing now. [/EDIT]


SimonBiggs:
@Matthew-Jennings this one here is hanging too...

PR #484, opened 2019-09-10
[MRG] Remove sphinxtheme

SimonBiggs:


SimonBiggs:
Just a note here, a downside of not using our own vendored theme is that we can't put the custom advertising on the page. We can address that in the future once again hopefully without vendoring the theme and needing a javascript stack. For now though I am removing the notes regarding sustainable open source.

PR #485, opened 2019-09-10
[WIP] Remove monomanage

SimonBiggs:


SimonBiggs:
I am here removing my dependency tree code (as well as a lot of other stuff). We might bring this back, but we can use the git history to find what we need.

PR #494, opened 2019-09-19
[MRG] Expand installation instructions

SimonBiggs:


SimonBiggs:
I found getting people to go to the developer page when they're just starting out is quite confusing. Better off to not be DRY here.

PR #558, opened 2019-10-23
Mwillam branch

mwillam:
Expand the linux contributer setup doc


SimonBiggs:
The command to build the docs locally is the following:

https://github.com/pymedphys/pymedphys/blob/1d47609289fa02b6232eb08f4fd5076fb1684f97/tox.ini#L26

docs/contrib/tutes/setup-linux.rst
```diff
@@ -11,6 +11,7 @@ Overview
 ========
 
 * Install Python 3.7
+* Install pip ``sudo apt-get install python3-pip``
```
SimonBiggs: I think Python 3.7 comes with pip preinstalled when it is installed via pyenv, so I don't think we need this one.

docs/contrib/tutes/setup-linux.rst
```diff
@@ -27,11 +28,17 @@ You're good to go.
 Opinionated Recommendations
 ===========================
 
-* Install Python with `pyenv`_
+* Install Python with pyenv
+  * Install dependencies: ```sudo apt-get install -y make build-essential libssl-dev zlib1g-dev libbz2-dev \
```
SimonBiggs: Here maybe best to just link to the prerequisites document? https://github.com/pyenv/pyenv/wiki/common-build-problems#prerequisites

docs/contrib/tutes/setup-linux.rst
```diff
@@ -27,11 +28,17 @@ You're good to go.
 Opinionated Recommendations
 ===========================
 
-* Install Python with `pyenv`_
+* Install Python with pyenv
```
SimonBiggs: reStructured text needs a new line here before sub dot points. ... It's write like that...

docs/contrib/tutes/setup-linux.rst
SimonBiggs: Also I believe this line is what the tests are bawking about:

https://github.com/pymedphys/pymedphys/pull/558/checks?check_run_id=271464861#step:10:197

I don't think triple ``` is valid reStructured text. But if this line goes and is replaced by the requirements link that should fix that issue.

PR #565, opened 2019-10-28
Working on WLutz

SimonBiggs:


SimonBiggs:
Make a modulo all close function, use it within the loop.

PR #567, opened 2019-10-30
Further WLutz work

SimonBiggs:


SimonBiggs:
@pchlap this would be a great pull to get a review on if you're available :).

SimonBiggs:
@Matthew-Jennings, I won't hold back a merge if you don't get a chance to review this, but I thought you might be interested.

This here is implementing both the Zenodo download method as well as WLutz field and BB finding.

pchlap:
> @pchlap this would be a great pull to get a review on if you're available :).

Sure, I will give this a go this afternoon :)

SimonBiggs:
> Sure, I will give this a go this afternoon :)

Thanks :)

pymedphys/labs/winstonlutz/createaxis.py
```diff
@@ -25,14 +25,8 @@
...
-import matplotlib.transforms
-
-
-def transform_axis(x, y, translation, rotation):
-    transform = matplotlib.transforms.Affine2D()
-    transform.rotate_deg(-rotation)
-    transform.translate(*translation)
 
+def transform_axis(x, y, transform):
     transformed_x = transform @ np.vstack([x, np.zeros(len(x)), np.ones(len(x))])
```
pchlap: I'd never seen this @ operator for matrix multiplication before.. Cool!

pymedphys/labs/winstonlutz/interppoints.py
```diff
@@ -128,3 +130,45 @@ def apply_transform(xx, yy, transform):
...
+
+    x = []
+    y = []
+    dist = []
+
+    for _, distance in enumerate(distances):
+        (
+            new_x,
+            new_y,
+        ) = pymedphys._utilities.createshells.calculate_coordinates_shell_2d(  # pylint: disable = protected-access
```
pchlap: Just curious, do we have a general rule of thumb as to when we allow access to protected members and when is best to avoid it?

pymedphys/labs/winstonlutz/createaxis.py
SimonBiggs: Yeah it is neat :)

pymedphys/labs/winstonlutz/interppoints.py
SimonBiggs: In this case `_utilities` is a module private to the `pymedphys` library. Any function within `pymedphys` can use these private modules, any external function should not as they will not be governed by the semver upgrade rules (https://semver.org/).

SimonBiggs:
> I suppose some more docstrings would help someone like me understand the context a little better

Certainly a fair point there :) Will do.

> Overall your coding style and structure is a thing of beauty, I can really learn a lot from it. 

Thanks @pchlap, made me smile. 

> Feel free to keep requesting reviews from me, but you probably would also want someone with a stronger medical physics background to critique the nitty gritty of stuff like this.

Unfortunately @Matthew-Jennings has mentioned that he'll need to be stepping back from reviewing so at the moment, until someone else steps up, this won't be viable in the near future. Nevertheless, I am making the code output figures which explicitly declare the results of each step of the code. This will allow physicists to judge for themselves if the code is working as it should be at each run time for their particular images.

PR #604, opened 2019-12-12
[MRG] Implement optional imports, begin work on JupyterLab bundling

SimonBiggs:


CHANGELOG.md
```diff
@@ -29,6 +29,22 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - nil -->
 
+
+## [Unreleased]
+
+### Breaking Changes
+
+- Moved `pymedphys pinnacle` cli command to be nested under
+ `pymedphys labs pinnacle`
```
SimonBiggs: Heya @pchlap,

I wanted to give you a heads up about this one. I wanted to make it clear within the PyMedPhys CLI which parts of the CLI are still within the labs. This would make `0.14.0` a breaking release for users of the pinnacle cli.

Let me know if you have any thoughts regarding this.

Cheers,
Simon

pchlap: Thanks for the info Simon. Yeah I think that's reasonable. It's probably about time I jump in and fix the underlying issues with that code so that we can move it out of labs and make it more stable moving forward.

I'll put that on my TODO list for January, at least to identify the issues that need to be addressed for the Pinnacle export tool. In the meantime I have no problem with there being breaking changes to the CLI.

Thanks!

SimonBiggs: Thanks @pchlap :)

PR #630, opened 2019-12-17
[WIP] JupyterLab bundle cli

SimonBiggs:


SimonBiggs:
Attempt building embedded Python with a miniconda installation. Install `jlpm` and use it to create the electron bundle.

Move `tdqm` from optional dependency to non-optional.

PR #673, opened 2020-01-03
Auto segmentation

matthewdeancooper:


examples/labs/auto-segmentation/dicom_input_pipeline.py
```diff
@@ -0,0 +1,73 @@
...
+    return np.array([ax_slice.pixel_array for ax_slice in dicom_volume])
+
+
+def scale_volume(volume, scale):
+    """
+    Get a list of sets of virtual ancestors for the given types
+    """
+    # TODO
+    # we are holding 2*volume in memeory - maybe look at generators
+    return np.array([(resize(image, scale)) for image in volume])
```
SimonBiggs: Here, so it's easier to see where resize came from (whether it was defined in your code or imported) it ends up being tidier to write either `skimage.transform.resize` or `transform.resize` instead of `resize` (I honestly prefer the whole hog and writing `skimage.transform.resize`)

See:
https://google.github.io/styleguide/pyguide.html#22-imports

examples/labs/auto-segmentation/dicom_input_pipeline.py
```diff
@@ -0,0 +1,73 @@
...
+    Quick hack to view a slice
+    """
+    # TODO Scale pixel values for CT
+    plt.imshow(pixel_array[index], cmap=plt.cm.bone)
+    plt.show()
+
+
+def normalise_pixel_array(pixel_array):
+    """
+    Return a normalise pixel array
```
SimonBiggs:
```suggestion
    Return a normalised pixel array
```

examples/labs/auto-segmentation/dicom_input_pipeline.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    # TODO
+    return pixel_array / np.max(pixel_array)
```
SimonBiggs: Is normalising to max the right thing to do here? ... This could make the pixel values of the region of interest (say vac bag) vary widely. Might be better to just normalise by dividing by the max possible pixel value (possibly one of `2**16` or `2**32` or `2**64` depending on the pixel_array type)

examples/labs/auto-segmentation/dicom_input_pipeline.py
```diff
@@ -0,0 +1,73 @@
...
+    file_names = glob(data_path + "/*.dcm")
+    file_names.sort()
+    return file_names
+
+
+def read_volume(file_names):
+    """
+    Returns a dicom volume from slices in file_names
+    """
+    return [pydicom.dcmread(file, force=True) for file in file_names]
```
SimonBiggs: There's going to need to be an extra processing step here. In the deep mind paper each data input was actually focused on a single slice, with "padding" slices either side of it. If a slice padding value of 5 is chosen, then each DICOM file will need to have each of its slices stepped through and exported as a new array which includes the slice in question and 5 slices either side.

This also means that the top 5 slices and bottom 5 slices of each set can't be used because they don't have sufficient slice padding.

On this point, the neural net will get very confused when there are slices that look like vacbag in the reference data that hasn't been contoured. I would recommend an initial processing step where all slices that don't have vacbag are removed from the training data.

examples/labs/auto-segmentation/dicom_input_pipeline.py
```diff
@@ -1,16 +1,21 @@
...
+
+import pydicom
+
 
 def get_file_names(data_path):
     """
     Returns a sorted list of all files in data_path with ext .dcm
     """
-    file_names = glob(data_path + '/*.dcm')
+    file_names = glob(data_path + "/*.dcm")
```
SimonBiggs: The fact that the pre-commit bot did something for you suggests that you haven't actually installed pre-commit by running:

```bash
poetry run pre-commit install
```

matthewdeancooper: Nope! Will get on that for next commit!

SimonBiggs:
Once you've read/commented on the notes above feel free to merge this. Also, as a follow up pull request it would be worth adding yourself to the front page:

https://github.com/pymedphys/pymedphys#our-team

examples/labs/auto-segmentation/dicom_input_pipeline.py
matthewdeancooper: Fixed.

examples/labs/auto-segmentation/dicom_input_pipeline.py
SimonBiggs: Just as an extra note, the Deep Mind paper used a padding of 10 slices either side (which is why the input size was 21, the slice in question + 10*2). I would recommend initially using less padding than that just to speed up the training.

A z slice is removed from either side each time an unpadded convolution is undergone, therefore the number of extra z slice included corresponds to how many layers are in the encoding arm of the network... I would recommend keeping this small for now until we want to spin up some fancy computing...

matthewdeancooper: Great. Thanks for the input. 

I'm planning to implementing a get_padding function for my next pull request once we have extracted the pixel arrays from each slice and created a resized pixel volume. From here we can loop through each slice and get its padding. You're right that we will need to remove the boundary slices once we know the padding size. However, i have included them in the above read_volume function because we will still need them as part of the padding for some interior slices.

I am preempting that we will eventually need to write these extracted arrays to a file as we iterate through our total data set; which will later be read into the model. However, I have not yet a look at the best practise for doing so :)

Let me know your thoughts!

Agreed. If there are slices which include the vac bag and it has not been labelled then this will need to be cleaned from the data. It is hard enough without mixed messages!

matthewdeancooper:
> Once you've read/commented on the notes above feel free to merge this. Also, as a follow up pull request it would be worth adding yourself to the front page:
> 
> https://github.com/pymedphys/pymedphys#our-team

Hi Simon,
It appears that i do not have permission to merge the PR.

examples/labs/auto-segmentation/dicom_input_pipeline.py
SimonBiggs: `import skimage.transform`

PR #679, opened 2020-01-06
move input pipeline

SimonBiggs:


SimonBiggs:
@matthewdeancooper just moving where this file is sitting

PR #682, opened 2020-01-07
Lightrad bug with ROI windows fixed

peterg1t:
Added the following changes to Lightrad:
Output filename now has a timestamp as well as displays machine name.
ROI graphs now have axis in cm instead of image pixels.


SimonBiggs:
Awesome :) thanks @peterg1t, if you hit "update branch":

image

You should be able to then merge it yourself :)

PR #744, opened 2020-02-24
Unet model

matthewdeancooper:
Hey Simon,

Most of the code is presented in two jupyter notebooks

1) dicom_process_example.ipynb for dicom to usable model input data
TODO: 
Only returns one structure but an easy fix by appending
TODO 
Account for fact that a structure may have multiple ROI per slice. I had a function that i thought was working great until I discovered that different ROIs for the same structure on the same image are stored in different indexes
TODO: Save data for each case with slice context and labels as a model input instance - all the data cant be held in memory at once.

2) model_building.ipynb
All the core architecture should be there, however, I'm having a bit of trouble with the up (decoder) path with respect to tensor sizes. - so a bit of debugging is still required here.


examples/protyping/auto-segmentation/model_building.ipynb
```diff
@@ -0,0 +1,466 @@
...
+    "    print(\">>>> upblock arg down_in:\", down_input)\n",
+    "    print(\">>>> upblock arg up_input:\", up_input)\n",
+    "    \n",
+    "    if upscale == True:\n",
+    "        up = tf.keras.layers.UpSampling3D(size=(1,2,2))(up_input)\n",
+    "        print(\">>>> upscale layer:\", up)\n",
+    "    else:\n",
+    "        up = up_input\n",
+    "    \n",
+    "    up = tf.keras.layers.concatenate([down_input, up], axis=0)\n",
```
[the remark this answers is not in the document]
SimonBiggs: I personally would have started with the above example, and then slowly adjusted it to make it work for 3D, trying to change as little as possible each time. That way, by starting with something that works, debugging is orders of magnitude quicker. If it stops working there are only a small number of things that could have broken it (the thing I last changed).

examples/protyping/auto-segmentation/dicom_process.py
```diff
@@ -0,0 +1,383 @@
...
+
+def get_binary_masks(structures, dicom_series, images):
+    # Should z location for each slice in patient space
+    z_slice_locations = [(dicom.ImagePositionPatient[2]) for dicom in dicom_series]
+    # an array full of zeros the same shape as images
+    labels = np.zeros_like(images, dtype=np.int16)
+
+    # NOTE: This function only works for a single label volume output
+    # See line for structure in structures[0:1] to choose
+    for structure in structures[0:1]:
```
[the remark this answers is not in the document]
SimonBiggs: Think the unix philosophy of "do one job, and do it well" for functions too. Ideally one could read the name of a function and pretty much know everything it should run off and do. Anything that doesn't come to mind, it shouldn't really go and do in secret. If a function accepts a list of structures, as the caller of the API I would expect it to return to me a list of results on all those structures (or raise an error if it could not). If the function is designed to only return one structure, then have the API only receive one structure also :).

examples/protyping/auto-segmentation/dicom_process.py
```diff
@@ -0,0 +1,383 @@
...
+    for file in dicom_files:
+        if hasattr(file, "ImageType"):
+            dicom_series.append(file)
+        elif hasattr(file, "StructureSetName"):
+            dicom_structures.append(file)
+        elif hasattr(file, "BeamSequence"):
+            dicom_plan.append(file)
+        else:
+            # TODO Add condition - will it always be DICOM dose file?
+            dicom_dose.append(file)
```
SimonBiggs: Better pattern here is to explicitly test for DICOM dose, and then have a final else that raises a `ValueError` saying something like "Unsupported DICOM type found."

examples/protyping/auto-segmentation/dicom_process.py
```diff
@@ -0,0 +1,383 @@
...
+    file_names.sort()
+    return file_names
+
+
+def read_dicom_files(file_names):
+    """
+    Returns a list of DICOM files read in from each file_name
+    in list file_names.
+    """
+    return [pydicom.dcmread(file_name, force=True) for file_name in file_names]
```
SimonBiggs: Interestingly, if you wanted to, you could actually take advantage of `stop_before_pixels`:

https://pydicom.github.io/pydicom/stable/reference/generated/pydicom.filereader.dcmread.html?highlight=header#pydicom.filereader.dcmread

What you can do, is read only the header for all the DICOM files, work out which ones are structures, and then initially, just read those. Then, if you wanted to, you could then make it so that you only read the pixel data from a given CT slice if it actually had pixel data relevant to a slice within a structure set.

* I could imagine a neat workflow would be, load all DICOM header data only (not pixel data)
* Pick the structure you care about, extract the coords
* Select a slice within the structures
* Define which slices are then needed to be loaded into memory from the hard disk by going +/- however many slices are required for your UNet layers (each layer drops two slices, one superior, and one inferior (https://en.wikipedia.org/wiki/Anatomical_terms_of_location#Superior_and_inferior)).
* Then look up the list of CT headers that you have recorded, find which files correspond to those slices you need
* Then only load those CT DICOM files and no more.

matthewdeancooper:
Ah, Simon you are a legend. Thank you for putting in so much work reviewing
the commit last night!

SimonBiggs:
> Ah, Simon you are a legend. Thank you for putting in so much work reviewing the commit last night!

My pleasure :)

examples/protyping/auto-segmentation/dicom_process_example.ipynb
```diff
@@ -0,0 +1,616 @@
...
+    "print(xyz_points[0:10])"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "labels = dicom_process.get_binary_masks(structures, dicom_series, images)"
```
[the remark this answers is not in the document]
SimonBiggs: Also, just so you have a few more tricks up your sleeve, the following line:

```python
names_to_pull = [name for name in structure_names_in_dicom if name in STRUCTURE_NAMES]
```

Could also be written as:

```python
set(structure_names_in_dicom).intersection(STRUCTURE_NAMES)
```

examples/protyping/auto-segmentation/dicom_process.py
matthewdeancooper: elif hasattr(file, "BeamSequence"):
            dicom_plan.append(file)
        elif hasattr(file, "DoseType"):
            dicom_dose.append(file)    
        else:
            raise ValueError('Unsupported DICOM type found', dicom_files)
    return dicom_series, dicom_structures, dicom_plan, dicom_dose`
`

examples/protyping/auto-segmentation/model_building.ipynb
```diff
@@ -9,7 +9,11 @@
   },
   {
    "cell_type": "code",
+<<<<<<< HEAD
+   "execution_count": 15,
+=======
    "execution_count": null,
+>>>>>>> 90184c9cc5017638aee4f6267356ad6324e88111
```
SimonBiggs: oops :), this'll be a bit painful. These are previous merge conflicts that needed to be resolved...

SimonBiggs:
DICOM usage within Tensorflow:

https://www.tensorflow.org/io/tutorials/dicom
https://www.tensorflow.org/io/api_docs/python/tfio/image/decode_dicom_data
https://www.tensorflow.org/io/api_docs/python/tfio/image/decode_dicom_image

examples/protyping/auto-segmentation/dicom_process.py
matthewdeancooper: After digging further I believe something resembling this process would be a good idea.

In regards to our conversation about time spent processing data, I have estimated the following:

**Saving all slices as .npy**
ie. pairs of image_slice.npy and mask_volume_for_image_slice.npy
Time per patient (averaged): 21 s
Time for 1000 patients (estimated): 5 h 51 m
Disk usage for 1000 patients (estimated): 567 G

The statistics for TIFF are very similar. 

**Dicom format**
Disk usage for 1000 patients (estimated):  80 G

It seems like building our own generator function to handle dicom files will pay off in the future.  Especially if we can cache the structure files for each batch.

SimonBiggs: Sounds like a good plan :). Also I think once you've got a handle on it the logic won't be too difficult. The hardest bit I imagine will be wrangling with TensorFlow Dataset objects within those pipeline functions.

PR #747, opened 2020-02-26
show tests failing with pydicom 1.4.1 -> 1.4.2

SimonBiggs:


SimonBiggs:
Heya @darcymason @scaramallion @mrbean-bremen :),

The only thing changed from master here is upgrading from `1.4.1 -> 1.4.2`. These tests passed when using `1.4.1`. I haven't had the chance to debug this, but it seems that there may have been some breaking changes introduced with the `1.4.1 -> 1.4.2` release.

https://github.com/pydicom/pydicom/compare/v1.4.1...v1.4.2

Cheers,
Simon

scaramallion:
> AttributeError: 'Dataset.is_little_endian' and 'Dataset.is_implicit_VR' must be set appropriately before saving.

They all look like write failures due to unset transfer syntax. `dcmwrite` got changed to make it primary (as its supposed to be) and `save_as` as secondary.

scaramallion:
Hmm, I'm not completely sure why this worked before. Those attributes default to `None`... it might be this is the danger in using the code style `if ds.is_little_endian:` rather than `if ds.is_little_endian is True:`

darcymason:
I think they did not originally default to None, so at some point when that was added these checks did become vulnerable.  Definitely agree that they should be explicit.

SimonBiggs:
Thanks for your help here :).

I'm still a little lost. Is this a case of "I should have actually been using `pydicom` differently" or is it a case of "something in `pydicom` needs to be fixed"? I don't mind too much either way. My current solution was to make the most recent `pymedphys-0.22.0` release be pinned to not be allowed to install `pydicom-1.4.2` (https://github.com/pymedphys/pymedphys/blob/0.22.x/pyproject.toml#L59).

scaramallion:
It's a case of needing to use pydicom differently. v1.4.2 fixed the bug in `dcmwrite` that let you use it wrongly.

darcymason:
To expand on @scaramallion's answer a little, pydicom should have enforced this correct use all along, but didn't in one arm of the writing capabilities, but did enforce in the other.  Those two should have had identical behavior.  We also had a test in our test suite which I had to fix during this change.  I do apologize that this was a breaking change however; I had hoped that our breaking test was not typical.  I'm happy to consider another patch release if we can reverse the "break", but ultimately it is much preferred for the user code to be explicit about the transfer syntax.

SimonBiggs:
All good :). I'm happy to update my usage. I shall do that before the next
release.

SimonBiggs:
Thanks for your help. This has been fixed in https://github.com/pymedphys/pymedphys/pull/748.

PR #758, opened 2020-03-05
MC UNet

SimonBiggs:


SimonBiggs:
@matthewdeancooper any luck with making it not crash? :)

SimonBiggs:
@matthewdeancooper thought you might be interested in the following:

https://www.youtube.com/watch?v=AgkfIQ4IGaM

SimonBiggs:
image

@matthewdeancooper should there be a `Dense` between the following two lines?

```python
result = tf.keras.layers.ReLU()(result)
result = tf.keras.layers.Add()([crop, result])
```

Also, if you're interested, I have put together a pipeline based off of a Python generator over here for a different type of task all together:

https://github.com/pymedphys/pymedphys/blob/wlutz-tensorflow-experimentation/examples/site-specific/cancer-care-associates/production/Winston%20Lutz/prototyping/tf_model/002-pixel-distance-method.ipynb

I also had my own take on how to style the creation of the model, I purposefully made it very similar to DeepMind's model. I just used less channels and double pooling. Managed to get it down to 300 000 trainable parameters.

matthewdeancooper:
Hey Simon,
I was able to use my generator to train on all the data i have collected
with a very simple model and make predictions! So the input pipeline is
working without any preprocessing.

The model contained just a single convolutional layer (1, 1, 1 filter) and
a single max pooling layer over all depth slices. I believe that this
architecture is only able to extract features at an individual pixel level
(ie. brightness) averaged over all slices. A model this simple was able to
learn that there are no contours outside the circular diameter of the ct
bore. :)

Plan is to attempt to train on the unet architecture today. Very exciting!

matthewdeancooper:
I have also started reading through the pragmatic programmer. I will get a
chance to read through your example this morning once i get to university.

SimonBiggs:
> Hey Simon, I was able to use my generator to train on all the data i have collected with a very simple model and make predictions! So the input pipeline is working without any preprocessing. The model contained just a single convolutional layer (1, 1, 1 filter) and a single max pooling layer over all depth slices. I believe that this architecture is only able to extract features at an individual pixel level (ie. brightness) averaged over all slices. A model this simple was able to learn that there are no contours outside the circular diameter of the ct bore. :) Plan is to attempt to train on the unet architecture today. Very exciting!

A great start, definitely worth getting something working to flex the pipeline and usage of TensorFlow itself :).

As a side note, there's a lot of value in making a callback for your model, ideally one that plots some key activations of each layer, that way you can get some insight into what the model is actually doing internally. Can really help with debugging.

SimonBiggs:
@matthewdeancooper Also, make sure to send Matt Sobolewski your lit review. :)

PR #764, opened 2020-03-06
WLutz TensorFlow Experimentation

SimonBiggs:


SimonBiggs:
@matthewdeancooper out of interest, I'm not a fan of the unintended effect of causing a lowres grid to appear via using pooling as per the DeepMind paper:

image

Out of interest, the Pix2Pix UNet downsampling doesn't use pooling, instead it uses a stride of 2 with a kernel size of 4. All the normal DeepMind convolutions could be undergone, but then as a conclusion instead of pooling, a convolution with a kernel size of 4 and a stride of 2 might have the same effect of downsampling without biasing the whole image into sub pixel areas:

```python
def downsample(filters, size, apply_batchnorm=True):
  initializer = tf.random_normal_initializer(0., 0.02)

  result = tf.keras.Sequential()
  result.add(
      tf.keras.layers.Conv2D(filters, size, strides=2, padding='same',
                             kernel_initializer=initializer, use_bias=False))

  if apply_batchnorm:
    result.add(tf.keras.layers.BatchNormalization())

  result.add(tf.keras.layers.LeakyReLU())

  return result
```
(https://www.tensorflow.org/tutorials/generative/pix2pix)

I'm just guessing here though. So for now possibly easiest to leave it as is and then when you're training make sure to plot activation images so that you can see what your layers are doing at the end of each epoch. ... helps diagnose if it's doing something silly...

SimonBiggs:
Note to self, for the next iteration use the following network instead:

https://arxiv.org/pdf/1902.09212.pdf

PR #829, opened 2020-05-04
Segmentation

matthewdeancooper:
Hi Simon, sorry we missed each other today. I was still asleep at 8am :)

Included is the code I've used for training with a few different loss functions to compare (specified in the config file imported in training.py)

I think the next step to take the model further would be to increase the amount of training data that we have access to. After so many epochs (between 60 to 140 depending on which loss function is used) the loss stops improving on the validation set. The DeepMind paper stated around 600 patients. We are currently sitting at about 10 with many of the slices removed due to missing contours.

I think the results look pretty good considering we are using such a small data set and considering that we are using clinical data that has not be tailored for our purpose.


SimonBiggs:
Yup, were you wanting to go down the path of getting more data? I can investigate remote access today if you would like...

examples/protyping/pymedphys-segmentation/training.py
```diff
@@ -0,0 +1,372 @@
...
+import paths as paths
+import tensorflow as tf
+import unet as unet
+
+physical_devices = tf.config.experimental.list_physical_devices("GPU")
+assert len(physical_devices) > 0, "Not enough GPU hardware devices available"
+config = tf.config.experimental.set_memory_growth(physical_devices[0], True)
+
+
+# The meaning of life
```
SimonBiggs, line 26: :)

matthewdeancooper:
From my end that is definately the end goal. However, churning out the
report has to come first. Lets revisit collecting more data in a week or
two.

SimonBiggs:
> From my end that is definately the end goal. However, churning out the report has to come first. Lets revisit collecting more data in a week or two.

Good plan.

PR #831, opened 2020-05-12
Updated qc-graticule

peterg1t:
Updated qc-grraticule and added oncentra utilities


peterg1t:
Hi Simon,
I closed the previous pull and started this one. I updated a couple of things but nothing major. The sanity checks seem to fail but I'm not sure how to fix these issues.

image
and when I re-run the checks I got the following then,
image
Let me know what you think. I ran poetry run pre-commit run --all-files and recommitted but not sure if that would help.
Thanks,
Pedro

SimonBiggs:
Let me see what I can do about that.

SimonBiggs:
I think you may just need to run the following:

```
poetry run pre-commit install
```

Then you should be set :)

PR #850, opened 2020-05-31
Move large files to zenodo

SimonBiggs:


tests/dicom/test_coords.py
```diff
@@ -1,97 +1,83 @@
...
+    assert set(expected_xyz.keys()) == set(ORIENTATIONS_SUPPORTED)
 
     test_ds_dict = {
         key: pydicom.dcmread(get_data_file(key)) for key in ORIENTATIONS_SUPPORTED
     }
-    print()
-    for orient, dicom in test_ds_dict.items():
-        test_xyz = xyz_axes_from_dataset(dicom, coord_system)
-
-        if save_new_baseline or print_:
```
SimonBiggs, line 45: @Matthew-Jennings this was an interesting case. These tests here weren't actually ever running because `print_` was set to `True`...

PR #877, opened 2020-06-13
Upgrade pylibjpeg from 1.0.0 --> 1.1.0

SimonBiggs:
Heya @scaramallion :),

I'm not sure what's causing this, and from what I can tell I don't really need to upgrade to 1.1.0. But I thought you might appreciate knowing that something seems to be up with pylibjpeg-libjpeg on pypi.

I don't have any idea what could be causing it...

I'm not fussed re a fix. Just wanted to give a heads up.

Cheers,
Simon


scaramallion:
I'm not sure what the problem is. I can install to a 64-bit Ubuntu 18.04.1 virtual machine with Python 3.6.7, pip 18.1 and numpy1.18.5 OK. 

Could it be a caching issue?

SimonBiggs:
Maybe...

SimonBiggs:
It appears to have been Python 3.6 vs Python 3.7 ...

PR #883, opened 2020-06-15
[WIP] Soft surface dice, mask with falloff, and pipeline

SimonBiggs:


SimonBiggs:
Heya @matthewdeancooper,

I have been moving the functions from the notebook over to `pymedphys.labs.autosegmentation`.

I'd be keen for you to take a squiz and give any feedback you might have. Here is an example notebook using the created functions:

https://github.com/pymedphys/pymedphys/blob/soft-surface-dice/protyping/auto-segmentation/sb/12-moving-into-pymedphys-labs.ipynb

Cheers,
Simon

SimonBiggs:
Heya @matthewdeancooper,

The following is probably the best example of the usage of the pipeline. I had to remove the dataset for ID 22026 since its right eye structure was corrected with the name "ODnew". But after removing that dataset I was then able to produce the following:

[16-sanity-tests-contours-make-sense.pdf](https://github.com/pymedphys/pymedphys/files/4784968/16-sanity-tests-contours-make-sense.pdf)

Using this notebook here:

https://github.com/pymedphys/pymedphys/blob/soft-surface-dice/protyping/auto-segmentation/sb/16-sanity-tests-contours-make-sense.ipynb

This demonstrates using the filters to just train on a subset of the data available. My thought was I could use this "eye" + "lens" dataset as a bit of a toy example that contains small organs to see how the soft surface dice fairs.

matthewdeancooper:
@SimonBiggs 

`# Of note, the DICOM file directory structure need not have any further
# organisation beyond being placed somewhere within one of the three
# 'training', 'validation', or 'testing'. They can be organised into
# directories by patient but that is not a requirement.`

Just a thought on storing data this way...

Ideally what we want is to be able to change which files are used for testing and validation to allow for k-fold cross validation. By using different training/valid/test partitions we can get a better idea of the architectures true performance i.e. some validation sets with contain doggier dogs than others.

SimonBiggs:
Yup that makes sense. Would it be better to maintain a study UID list that
assigns different studies to different datasets?

SimonBiggs:
As in keep all the DICOM files in one directory, and just have a json file that assigns different DICOM files to different datasets?

matthewdeancooper:
> As in keep all the DICOM files in one directory, and just have a json file that assigns different DICOM files to different datasets?

Yep. I think this is the way to go! Noting that we need to split at the patient level so that we have training/valid/test independence.

SimonBiggs:
Yup, the json file could contain study set UIDs I think.

matthewdeancooper:
Do you remember what the download rate was on your end? When I ran the script last night I was getting ~10-30 kB/s from a wired connection. Which means a few hours of download time.

SimonBiggs:
Yuck :(. I'm actually driving to Sydney at the moment, but I've had a few
ideas on how to be able to have the dataset usable before having to
download it all. As in, at the moment we have to download all the DICOM
files, then we can start working with it. Ideally, one would only need to
download config files and then the DICOM files that are about to be used
will be downloaded using TensorFlow's dataset prefetch method. In that way,
the next 10 datasets will be downloading while the mask is being
calculated, and all that is required is one ct file and one structure set
file to begin.

I likely won't be able to implement that until Sunday though when I get
back from the wedding.

To briefly outline the idea, instead of uploading the DICOM files as large
collective ZIPs to Zenodo, zip them all up individually. Upload the UID
mapping caches as well. That way, the user downloads the UID maps, and the
can run their filter on those maps (before even downloading any DICOM
files). After they've run their filter, they will have a list of study IDs
that still contain slices after filtering, then a set of study IDs are sent
to the dataset creator which will begin fetching and masking DICOM files.
But importantantly it will only fetch the DICOM files needed in order to be
"10 DICOM files ahead of the current mask" for example. I'm hoping to be
able to make it relatively seemless so that you only download what you need
to, and you shouldn't notice too much the effects of needing to download
the data.

The other important step will be to make it so that the npz mask cache is
split up by structure, that way it shouldn't need to be cleared when
different structure masks are wanted. Also, because the mask cache will end
up being huge, I'm also going to compress the masks before saving them to
disk. Slight time loss in loading them up, but massively less disk space
usage.

Anyway, we'll see how we go.

SimonBiggs:
Also which script? Was it number 17?
https://github.com/pymedphys/pymedphys/blob/master/protyping/auto-segmentation/sb/17-use-downloadable-data.ipynb

Also, if you're interested, I already attempted making it so that the cache
is directly downloaded:

https://github.com/pymedphys/pymedphys/blob/add-e2e-auto-contour-data/protyping/auto-segmentation/sb/24-stream-compressed-dataset-from-zenodo.ipynb

But, some stale cache crept in and some of those files aren't right :/. ...
Part of the reason for reworking that idea except making sure to still
download the DICOM files and calculate the mask locally, so any code
changes, or DICOM file changes, propegate readily.

PR #880, opened 2020-06-15
Added plan complexity code to labs

db72-git:
I have added the code I have written for calculating a couple of plan complexity metrics to the labs as discussed via email a couple of weeks ago.

PR #896, opened 2020-06-21
Plan complexity

SimonBiggs:
A continuation of https://github.com/pymedphys/pymedphys/pull/880 by @db72-git


SimonBiggs:
So, @db72-git, the main thing that was causing this to fail was due to not implementing the following pattern: 

https://stackoverflow.com/a/419185

So I have fixed that, as well as fixed up some importing due to the need to import external packages from `pymedphys._imports` so as to support cases where users want to just install pymedphys with:

```
pip install pymedphys --no-deps
```

SimonBiggs:
I also just fixed a few cases of spelling, and refactored out some functions that didn't use self. The only API change is any instance of Apperture in the name has been changed to Aperture.

SimonBiggs:
@db72-git to simplify things a little on your end, now that this works (all except for the updates CI) I'm going to merge this.

A next step I think would be to expose your tkinker GUI via the pymedphys CLI. But let's have that in another pull request, and if you're not sure how to go about that, I can do that for you.

db72-git:
Thanks so much for all this, that may have taken some time to figure that out! I will have a bash at incorporating the CLI as soon as I get a chance

SimonBiggs:
> Thanks so much for all this, that may have taken some time to figure that out!

My pleasure :)

Issue #908, opened 2020-06-30
Propose renaming `labs` to `experimental`


SimonBiggs:
This makes the nature of the code in this region more explicit, and it comes into line with other libraries that follow the same style:

https://discuss.streamlit.io/t/adding-beta-and-experimental-channels-to-streamlit/2765
https://github.com/google/jax/tree/master/jax/experimental

SimonBiggs:
@sjswerdloff, how about I make this change now so that when you release your CLI it can be released with the new name from the get go. That way you won't have to change your usage of it.

I'm expecting that the CLI will be something like:

```
pymedphys experimental dicom anonymise --pseudo /input/file.dcm -o /output/directory/
```

This experimental scope of `dicom anonymise` will import from the non-experimental version, so that it will act exactly the same, except have extra option(s) added. I can set this up for you if you like?

This will be designed with the aim to in the future drop the "experimental" scope once it's been put through its paces.

SimonBiggs:
Heya @pchlap,

Are you okay with the name of `labs` changing to `experimental`? This will effect the pinnacle CLI usage. The intent is in the future to have both `experimental` and `beta` scopes so that when PyMedPhys goes to version `>= 1.0.0` portions of the CLI can be actively iterated on without needing to change the major version number of PyMedPhys. The name `experimental` is used in other libraries, so this brings us in line with what others are doing. It is also a much clearer name which has the name carry with it the warnings that should be inherent in that portion of PyMedPhys.

Let me know your thoughts. If needed I can make a pointer from `labs` through to `experimental` so that the CLI doesn't break for a few versions. But, given that it is the labs, personally I would prefer to be able to make a clean break and not have to have to manage a staged roll out.

sjswerdloff:
Sure.  While that wasn't my initial design direction, that is how things will end up based on a refactoring approach.
Now that I know what is involved (from the work that went in to my draft PR), it's a lot easier for me to identify the refactorings that will result in the design implied by the CLI proposal below.

> @sjswerdloff, how about I make this change now so that when you release your CLI it can be released with the new name from the get go. That way you won't have to change your usage of it.
> 
> I'm expecting that the CLI will be something like:
> 
> ```
> pymedphys experimental dicom anonymise --pseudo /input/file.dcm -o /output/directory/
> ```
> 
> This experimental scope of `dicom anonymise` will import from the non-experimental version, so that it will act exactly the same, except have extra option(s) added. I can set this up for you if you like?
> 
[...]

PR #911, opened 2020-06-30
pseudonymisation support

sjswerdloff:
new module for pseudonymise
added identify_uids.json to parallel identifying_keywords
  to avoid forcing anonymise to use the uids
updated identifying_keywords to include sequences commonly found in RT
added test_pseudonymise and a test file to parallel anonymise
updated cli to include pseudonymise
added RT Ion Plan as "RN" to core.py


SimonBiggs:
Hi @sjswerdloff,

Before we dive into the PR and review process would you be able to have a read through the two documents available at:

https://github.com/google/eng-practices

We aim to follow Google's review process. The documents above details that large pull requests must be avoided, and mentions that the first pull request for significant feature changes should be a standalone refactoring pull request. So with that in mind, I have three initial requirements.

Firstly, this is a large pull request with some significant changes. I have a few design requirements which are often brought to light when a pull request is opened or discussed with the first few initial small code commits. Unfortunately you have already done a large amount of work which will make these requirements more painful. Sorry about that. That is one key reason that small iterative pull requests are needed. So, the first requirement, is this change is undergone in smaller bites.

Secondly, the first pull request needs to have no inclusion of your own coding logic, instead, simply have it be a refactor of the current code so that the functions that already exist can be easily reused by your code. If there is already a function somewhere in the code base that achieves a certain logical task, that is not to be repeated in another file. Search for "DRY" within the following book:

https://www.cin.ufpe.br/~cavmj/104The%20Pragmatic%20Programmer,%20From%20Journeyman%20To%20Master%20-%20Andrew%20Hunt,%20David%20Thomas%20-%20Addison%20Wesley%20-%201999.pdf

The following blog is also another article discussing the DRY principle:

https://thevaluable.dev/dry-principle-cost-benefit-example/

To make the whole code base DRY you need to refactor the current functions that are there to be callable by your code. This refactoring is a pull request on its own.

Thirdly, given the newness of this all, I would like this to be within the labs. Therefore, once the current anonymisation code has been refactored to support reuse by your code, then your code can import those functions while having your code be written into the labs itself. The labs is the "experimental" part of PyMedPhys. In fact, I plan to rename the "labs" module to "experimental" in the near future (see https://github.com/pymedphys/pymedphys/issues/908).

sjswerdloff:
sure.  I failed to do DRY in the test file.  I've got that queued up, but I'm struggling to remember the set of process steps for running all of the tests prior to commit.  I do have poetry and tox, etc. installed.

If something is moved to lab or experimental, is there still a cli available?
That's how I was able to do my testing against a realistic set of data (where I have an RT Dose, RT Plan, RT Structure Set, and CT, as a "connected" set of data).
Or does something in lab/experimental need to come with it's own script to give it a cli ?

SimonBiggs:
> If something is moved to lab or experimental, is there still a cli available?
> That's how I was able to do my testing against a realistic set of data (where I have an RT Dose, RT Plan, RT Structure Set, and CT, as a "connected" set of data).
> Or does something in lab/experimental need to come with it's own script to give it a cli ?

CLI is certainly still available (encouraged even). See the following CLI from the labs:

https://docs.pymedphys.com/ref/cli/pinnacle.html


> I've got that queued up, but I'm struggling to remember the set of process steps for running all of the tests prior to commit. I do have poetry and tox, etc. installed.

Running `poetry run pymedphys dev tests` should run all of the main tests. To hit the ones that are marked slow, run `poetry run pymedphys dev tests --run-only-slow`. If you want to lint all your code you can run `poetry run pymedphys dev tests --pylint`. However when it comes to linting, I find it much easier to just use VSCode and have it lint the code while I'm writing it.

SimonBiggs:
Also, if you ever want to know what is being tested on the server, you can cheat and peak into the following file:

https://github.com/pymedphys/pymedphys/blob/master/.github/workflows/on-push.yml#L241-L269

SimonBiggs:
Hi @sjswerdloff,

I have made a mock CLI entry point under the experimental scope set up to mimic the DICOM CLI:

https://github.com/pymedphys/pymedphys/blob/dfb15988f0073ef12992aaf9060fd0e4c07852db/pymedphys/cli/experimental/dicom.py#L15-L31

It is now merged into master. Hopefully that'll help :).

Cheers,
Simon

SimonBiggs:
@sjswerdloff do you think it might be better to have pymedphys depend upon pydicom/deid? Build CLIs etc based upon the work done in deid?

sjswerdloff:
> @sjswerdloff do you think it might be better to have pymedphys depend upon pydicom/deid? Build CLIs etc based upon the work done in deid?

My preference is to continue down the path in pymedphys.  My intent is to work things out in pymedphys, where there is a clear focus on Radiation Oncology, and then re-implement in pydicom/deid (which is more oriented towards diagnostic).  Once pydicom/deid seems mature enough, it would be reasonable to strangle anonymise/pseudonymise by utilising the outward facing API (including CLI) and replacing the internal implementation with calls in to pydicom/deid.

Sometimes the direct route is not as feasible as the indirect route.

sjswerdloff:
obsoleted by experimental pseudonymisation strategy PR #985 (and refactoring anonymisation so that code duplication was kept to a minimum)

PR #923, opened 2020-07-05
Find anonymise test data via zenodo

sjswerdloff:
I took the existing test plan file, zipped it, uploaded that to zenodo.
I left the old approach in as comments, but I can delete those now or delete them when I add the RT Ion Beams Treatment Record "Not Quite Anonymised" test file.
This counts as a refactoring, while putting in the additional test file counts as new logic, so keeping those separate


SimonBiggs:
> while putting in the additional test file counts as new logic, so keeping those separate

Have a look at the following (document linked to before regarding pull requests):

https://github.com/google/eng-practices/blob/master/review/developer/small-cls.md#what-is-small-what_is_small

Importantly:

> * The CL makes a minimal change that addresses just one thing. This is usually just one part of a feature, rather than a whole feature at once. In general it's better to err on the side of writing CLs that are too small vs. CLs that are too large. Work with your reviewer to find out what an acceptable size is.
> * **The CL should include related test code.**

Treat the tests that test the "changed thing" as part of the same pull request. It is important in the long run that each pull request also contains the testing code that validated that that pull request is doing the right thing.

SimonBiggs:
@sjswerdloff have you installed pre-commit as per the contributor install instuctuions?

https://docs.pymedphys.com/howto/advanced/setup-linux.html

sjswerdloff:
In refactoring, I re-ran 
poetry run pymedphys dev tests --run-only-pydicom
I have not put in a "new" test file.  I've only changed where the test file was gathered from.
However, when I do put in a new test file (the treatment record), there will be an additional test that uses that test file.

I had run the poetry run pre-commit install...
however, because I had done a fresh clone after that, it got lost. mea-culpa.
So I've run it again, and will keep in mind that it's necessary if I do a fresh clone.

pymedphys/tests/dicom/test_anonymise.py
```diff
@@ -31,10 +32,9 @@
 from pymedphys._dicom.utilities import remove_file
 from pymedphys.dicom import anonymise as anonymise_dataset
 
-HERE = dirname(abspath(__file__))
-DATA_DIR = pjoin(HERE, "data", "anonymise")
-TEST_FILEPATH = pjoin(DATA_DIR, "RP.almost_anonymised.dcm")
-
+DATA_PATHS = download.zip_data_paths("rtplan-anonymisation.zip")
```
SimonBiggs, line 35: By placing this as a constant at the top of the file this will make it so that this data is downloaded even it is not used.

Could this file path logic either be placed at the top of the function using it, or if it is used by multiple functions have it be placed either within a pytest.fixture (https://docs.pytest.org/en/stable/fixture.html), or a function inside a functools.lrucache decorator?

SimonBiggs:
> I had run the poetry run pre-commit install...
> however, because I had done a fresh clone after that, it got lost. mea-culpa.
> So I've run it again, and will keep in mind that it's necessary if I do a fresh clone.

Awesome :) All good. There are a few infrastructure steps with PyMedPhys. I try to find a good balance between set up complexity and tools that help long term maintainability. Hopefully I am still striking a good balance.

SimonBiggs:
> I have not put in a "new" test file. I've only changed where the test file was gathered from.
> However, when I do put in a new test file (the treatment record), there will be an additional test that uses that test file.

Ahh perfect, yup. That's brilliant. Thanks that makes sense. Refactor out the current test file so that you can add more to it.

Cheers :)

pymedphys/tests/dicom/test_anonymise.py
```diff
@@ -31,10 +32,9 @@
...
+TEST_POSIX_PATH = next(x for x in DATA_PATHS if x.name == "RP.almost_anonymised.dcm")
+TEST_FILEPATH = TEST_POSIX_PATH.absolute().as_posix()
```
SimonBiggs, line 37: Could this be why it's failing on Windows?

pymedphys/tests/dicom/test_anonymise.py
```diff
@@ -31,10 +32,9 @@
 from pymedphys._dicom.utilities import remove_file
 from pymedphys.dicom import anonymise as anonymise_dataset
 
-HERE = dirname(abspath(__file__))
-DATA_DIR = pjoin(HERE, "data", "anonymise")
-TEST_FILEPATH = pjoin(DATA_DIR, "RP.almost_anonymised.dcm")
```
SimonBiggs, line 36: Worth deleting this file now it's also not needed.

SimonBiggs:
> In refactoring, I re-ran
> poetry run pymedphys dev tests --run-only-pydicom

You might appreciate the following command:

```
poetry run pymedphys dev tests --lf
```

It will run only the tests that failed on the last run. If no tests failed on the last run it'll run all tests.

pymedphys/tests/dicom/test_anonymise.py
sjswerdloff: yes.
I'll try str() instead of as_posix().  It's not clear from the documentation for zip_data_paths what gets returned other than it being pathlib related, but there's a hierarchy of classes that are possible.
Once I have the windows test working, I'll go back in and see about using a fixture or lru_cache.
So I won't expect a merge until I've got that worked out.

SimonBiggs: The `zip_data_paths` function returns a list containing all of the extracted paths that were found within the zip as the same type returned by `pathlib.Path`, this will be a Windows path on Windows systems, and a Posix path on Mac and Linux.

sjswerdloff:
utilising lru_cache approach.
The higher level call is a bit unnecessary, but it's not wrong (it's just how I got to where I was going).

SimonBiggs:
I've approved this commit, and I have just invited you to be a member of the pymedphys org. That way, once you're happy with it you can merge at your own discretion.

sjswerdloff:
I'm honoured (and accepted).
But GitHub seems to have different ideas about whether I can merge in to a protected branch.
I don't have experience as a group member on GitHub.
I do have experience being in charge of the equivalent on BitBucket Enterprise <cough>
I'll do my best to chip in on the reviews and merges (after making sure I'm clear on the entire set of house rules)

SimonBiggs:
I suspect you might have to click on the following:

image

sjswerdloff:
That was not the UI I had.
If I run in to something similar again, I'll take a screenshot.
Thank you for finishing off the PR.

SimonBiggs:
> That was not the UI I had.
> If I run in to something similar again, I'll take a screenshot.

Kk, GitHub is refreshing their UI, so that would likely be quite helpful.

> Thank you for finishing off the PR.

My pleasure :)

PR #929, opened 2020-07-11
Pinnacle orientations bug fix

pchlap:
The dose grid exported was only correct when patients were in HFS. Some tests have been added to test other orientations and the dose export has been corrected. Change log has been updated with the bug fix. Let me know if there's anything else you'd like before merging. Once again this doesn't require a release, happy to have this rolled out with the next update :)

Thanks
Phil


pymedphys/_experimental/pinnacle/rtdose.py
```diff
@@ -201,10 +201,14 @@ def convert_dose(plan, export_path):
...
-    elif "HFP" in patient_position or "FFP" in patient_position:
-        ds.ImageOrientationPatient = [-1.0, 0.0, 0.0, 0.0, -1.0, -0.0]
+    if "HFS" in patient_position:
+        ds.ImageOrientationPatient = [1, 0, 0, 0, 1, 0]
+    elif "HFP" in patient_position:
+        ds.ImageOrientationPatient = [-1, 0, 0, 0, -1, 0]
+    elif "FFS" in patient_position:
+        ds.ImageOrientationPatient = [-1, 0, 0, 0, 1, 0]
+    elif "FFP" in patient_position:
+        ds.ImageOrientationPatient = [1, 0, 0, 0, -1, 0]
```
SimonBiggs, line 211: Could we make a dictionary that looks like the following:

```python
image_orientation_map = {
	'FFDL': [0, 1, 0, 1, 0, 0],
	'FFDR': [0, -1, 0, -1, 0, 0],
	'FFP': [1, 0, 0, 0, -1, 0],
	'FFS': [-1, 0, 0, 0, 1, 0],
	'HFDL': [0, -1, 0, 1, 0, 0],
	'HFDR': [0, 1, 0, -1, 0, 0],
	'HFP': [-1, 0, 0, 0, -1, 0],
	'HFS': [1, 0, 0, 0, 1, 0]
}
```

From:

https://github.com/pymedphys/pymedphys/blob/40780daff9ccbf92dccc34bbc777207e0a409a72/pymedphys/_dicom/coords.py#L90-L101

And then, we can simply write:

```python
ds.ImageOrientationPatient = image_orientation_map[patient_position]
```

Given how ubiquitous that map is, it should probably sit within a new file placed at `pymedphys/_dicom/constants/orientation.py`.

And then all logic that uses this concept within PyMedPhys can import from there. For example the following function within PyMedPhys:

https://github.com/pymedphys/pymedphys/blob/40780daff9ccbf92dccc34bbc777207e0a409a72/pymedphys/_dicom/dose.py#L251-L256

Could be refactored to be:

```python
def require_patient_orientation(ds, patient_orientation):
    if not np.array_equal(
        ds.ImageOrientationPatient, 
        np.array(image_orientation_map[patient_position])
    ):
        raise ValueError(
            "The supplied dataset has a patient "
            f"orientation other than {patient_orientation}."
        )
```

pymedphys/_experimental/pinnacle/rtdose.py
```diff
@@ -338,6 +342,14 @@ def convert_dose(plan, export_path):
...
         for i in range(3):
             idx[i] = -(origin[i] - prescription_point[i]) / spacing[i]
+
+        if patient_position in ("HFP", "FFS"):
+            idx[0] = -idx[0]
+        if patient_position in ("HFP", "FFP"):
+            idx[1] = -idx[1]
+        if patient_position in ("FFS", "FFP"):
+            idx[2] = -idx[2]
+
```
SimonBiggs, line 352: Is there a way to implement this logic in a general way that applies to all orientations by creating a transform from by `ds.ImageOrientationPatient`?

pymedphys/_experimental/pinnacle/rtdose.py
```diff
@@ -415,6 +427,11 @@ def convert_dose(plan, export_path):
     pixel_binary_block = struct.pack("%sh" % len(pixelvaluelist), *pixelvaluelist)
     ds.PixelData = pixel_binary_block
 
+    # If Feet first, flip the dose grid
+    if patient_position in ("FFS", "FFP"):
+        arr = ds.pixel_array
+        ds.PixelData = np.flip(arr, axis=0).tostring()
```
SimonBiggs, line 433: It seems like we should have a function that is within PyMedPhys that converts pixel data from one orientation (say 'HFS')  to another (for example 'FFS'):

```python
supported_orientations = ('HFS', ... etc)  # maybe even image_orientation_map.keys() ?

def convert_pixel_orientation(pixel_data, from_orientation, to_orientation):
    if not set((from_orientation, to_orientation)).issubset(supported_orientations):
        raise ValueError("Provided orientation not supported.")
	...
	do_stuff()    
    ...

	return transformed_pixel_data
```

Does that seem doable? And then that function could be called here.

SimonBiggs: A function like that would happily reside in a place such as a new file at `pymedphys/_dicom/pixelarray.py`

pchlap:
> Thanks @pchlap, this is just a first pass initial review. I would like to also review the tests done, haven't done that yet.
> 
> Lovely PR :).

Thanks for the review @SimonBiggs. Some great suggestions, I'll address those tomorrow. A general transform between patient orientations would of course be the correct way to do it, I'll see if I can figure it out :)

As for the tests, I fear that you're just going to find some more end-to-end tests :( Reworking those to proper unit tests is definitely on the (relatively long) TODO list for this tool.

SimonBiggs:
> on the (relatively long) TODO list for this tool.

I think a manageable way to achieve this over time is actually just how it has been happening already. If step by step we find neat reusable functions and logical units that can be refactored out into PyMedPhys (like the above `convert_pixel_orientation` and `image_orientation_map` for example) then as time goes on more and more of the building blocks of this tool will have moved itself out of experimental.

That also makes it so whenever I do a review it is always of a relatively small understandable piece, which is a big plus. :).

SimonBiggs:
Also, if it helps, I just did a development release of this branch:

```
pip install pymedphys==0.32.0dev0
```

SimonBiggs:
> As for the tests, I fear that you're just going to find some more end-to-end tests :(

That's completely fine, does mean there's probably not much I can review, but, end-to-end tests are still exceptionally useful :). Thanks @pchlap!

I'll set the review to approve so that you can be free to merge once you're happy with your changes :).

pymedphys/tests/dicom/test_dose.py
```diff
@@ -79,19 +79,25 @@ def test_require_patient_orientation_be_HFS():
...
-
-        else:
-            with pytest.raises(ValueError) as ev:
-                require_patient_orientation_be_HFS(ds)
-            assert (
-                "The supplied dataset has a patient orientation "
-                "other than head-first supine" in str(ev.value)
-            )
+
+        for test_orient in test_orientations:
```
SimonBiggs, line 86: Beautiful :). Thanks :)

pymedphys/tests/dicom/test_dose.py
```diff
@@ -79,19 +79,25 @@ def test_require_patient_orientation_be_HFS():
...
+                    ea.value
+                )
+
+            else:
+                with pytest.raises(ValueError) as ev:
+                    require_patient_orientation(ds, test_orient)
+                assert (
+                    "The supplied dataset has a patient orientation "
+                    f"other than {test_orient}" in str(ev.value)
+                )
```
SimonBiggs, line 103: What's your opinion here? I've never been that much of a fan directly asserting the text of a raised message... Do you think this was needed in the original test? Or do you think it may have been better just writing:

```python
with pytest.raises(ValueError) as ev:
    require_patient_orientation(ds, test_orient)
```

pchlap: I agree, I think testing that the error is raised is sufficient. The text being "correct" doesn't affect the functionality. I'll make the adjustment :)

SimonBiggs: Thanks @pchlap

pymedphys/_experimental/pinnacle/rtdose.py
pchlap: Hi @SimonBiggs , I'm not sure how to do this to be honest. I think a general function for this would need to adjust the axis in all directions to convert between the orientations. But for some reason the Pinnacle dose only needs to be flipped to convert between Head first and feet first, the other orientations seem to already be accounted for.

Any ideas why this might be? Am I missing something obvious? If not perhaps we'll leave it for now? Since this is probably only Pinnacle specific there wouldn't be much benefit in pulling out a general function.

SimonBiggs: > Any ideas why this might be? Am I missing something obvious? If not perhaps we'll leave it for now? Since this is probably only Pinnacle specific there wouldn't be much benefit in pulling out a general function.

Hmm, not sure why that is. I agree, let's leave it for now.

PR #930, opened 2020-07-12
moved

matthewdeancooper:
Adding some new files in the auto-segmentation directory.
Moving all old files into mc_old_#. Mostly what we still need from these old files is the training and augmentation functions.


SimonBiggs:
Heya @matthewdeancooper :),

There's quite a bit going on here. In the near future I would be quite keen to see some of the reusable pieces of this find their way into the following part of the library -- https://github.com/pymedphys/pymedphys/tree/master/pymedphys/_experimental/autosegmentation

I think, I'll focus on reviewing code as it gets migrated out of prototyping into the experimental part of the library itself.

:)

protyping/auto-segmentation/create_rs_file.py
```diff
@@ -0,0 +1,202 @@
...
+ds = add_referenced_frame_of_reference_sequence(ds, dcms)
+ds = add_structure_set_roi_sequence(ds)
+
+# ds = add_roi_contour_sequence(ds, dcms)
+ds = add_rt_roi_observations_sequence(ds)
+
+for x in ds:
+    print(x)
+
+ds.save_as("Test")
```
SimonBiggs, line 202: Probably should save this file in a temp directory, or put the filename in a `.gitignore`. Good rule of thumb to never commit to version control build artefacts.

:)

Issue #246, opened 2019-05-02
Easier to read error logs


SimonBiggs:
Something that might be worth making liberal use of ...

https://github.com/cknd/stackprinter

Centrus007:
I had been thinking that we ought to use Python's native [logging ](https://docs.python.org/3/library/logging.html) for our CLIs, at least. Output could be redirected to a GUI widget as well.

I'm pretty well experienced with it - we use it liberally at the RAH.

Centrus007:
(I meant that as a relevant aside: it need not be `logging` *or* `stackprinter`. I'm just very used to using `logging` to output more useful error messages to users, to provide information on execution progress, to increase verbosity and debug issues, etc.)

SimonBiggs:
Yeah let's do that. I'm just not experienced with it, more than happy to
learn though :).

PR #931, opened 2020-07-13
incorporate use of python library logging with a fallback configuration

sjswerdloff:
log the presence of an unexpected private element at the warning level
see Issue Easier to read error logs #246

### DIFF
--- a/pymedphys/_dicom/anonymise.py
+++ b/pymedphys/_dicom/anonymise.py
@@ -15,6 +15,7 @@
 
 import functools
 import json
+import logging
 import os.path
 import pprint
 from copy import deepcopy
@@ -35,6 +36,18 @@ from pymedphys._dicom.constants import (
 from pymedphys._dicom.utilities import remove_file
 
 HERE = dirname(abspath(__file__))
+
+logger = logging.getLogger(__name__)
+# basicConfig will only have effect if it was not called previously
+# initial testing indicated that it has not been invoked previously
+# however, future work to add logging to pymedphys and the CLI
+# in particular should precede and hence supercede this invocation
+logging.basicConfig(
+    filename="pymedphys_anonymise.log",
+    format="%(asctime)s|%(name)s|%(message)s",
+    datefmt="%Y/%m/%d %H:%M:%S",
+    level=logging.INFO,
+)
 
 IDENTIFYING_KEYWORDS_FILEPATH = pjoin(HERE, "identifying_keywords.json")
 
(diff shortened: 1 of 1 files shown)


sjswerdloff:
I found it difficult to understand why an anonymisation test would fail.
Logging seems an appropriate and unobtrusive way of addressing that.
I will be following up descent in to sequences within the is_anonymised_dataset() function, and logging of findings within that descent.  It turns out that the lack of descent in to sequences in the is_anonymised_dataset() function was failing to find that earlier versions would result in datasets that in fact were not anonymised, including the test RT Plan.
So the approach to PRs of which this is the first in a set is:
add logging to anonymisation.py
add recursive descent in to sequences in is_anonymised_dataset() with logging
add logging to test_anonymise.py
refactor test_anonymise.py for more flexible testing (ability to add test data files)
add test data files to test_anonymise.py (accessed through zenodo)

pymedphys/_dicom/anonymise.py
```diff
@@ -36,6 +37,18 @@
...
+# basicConfig will only have effect if it was not called previously
+# initial testing indicated that it has not been invoked previously
+# however, future work to add logging to pymedphys and the CLI
+# in particular should precede and hence supercede this invocation
+logging.basicConfig(
+    filename="pymedphys_anonymise.log",
+    format="%(asctime)s|%(name)s|%(message)s",
+    datefmt="%Y/%m/%d %H:%M:%S",
+    level=logging.INFO,
+)
```
SimonBiggs, line 50: Setting the logging config here will cause it to be adjusted on this module import (not when a particular function is run). This will result in awkward initiation logic.

Logging is currently already initialised within the CLI:

https://github.com/pymedphys/pymedphys/blob/cede62015203b09f9490e99c6eb6f653ed43c968/pymedphys/cli/__init__.py#L73

Could you use the same logging methods as used within https://github.com/pymedphys/pymedphys/blob/cede62015203b09f9490e99c6eb6f653ed43c968/pymedphys/_icom/listener.py#L24

And then, if you want to call the CLI with a debug log level you can run:

```
pymedphys --debug dicom anonymise
```

SimonBiggs:
> So the approach to PRs of which this is the first in a set is:
> add logging to anonymisation.py
> add recursive descent in to sequences in is_anonymised_dataset() with logging
> add logging to test_anonymise.py
> refactor test_anonymise.py for more flexible testing (ability to add test data files)
> add test data files to test_anonymise.py (accessed through zenodo)

Yup, this order seems logical to me.

sjswerdloff:
I am opening a new PR that adheres to the request above regarding use of the top level logger

PR #940, opened 2020-07-20
added logging of elements that do not appear to be anonymised

sjswerdloff:
see Issue Easier to read error logs #246
logs at info level when private element is found and not expected
logs at info level a public element that is in the anonymisation list (keyword dictionary and not otherwise excluded)
uses logging directly and therefore defaults to the configuration of the top level logger

### DIFF
--- a/pymedphys/_dicom/anonymise.py
+++ b/pymedphys/_dicom/anonymise.py
@@ -15,6 +15,7 @@
 
 import functools
 import json
+import logging
 import os.path
 import pprint
 from copy import deepcopy
@@ -466,10 +467,13 @@ def is_anonymised_dataset(ds, ignore_private_tags=False):
                     float(elem.value), float(dummy_value)
                 ):
                     continue
-
+                logging.info("%s is not considered to be anonymised", elem.name)
                 return False
 
         elif elem.tag.is_private and not ignore_private_tags:
+            logging.info(
+                "%s is private and private tags are not being ignored", elem.tag
+            )
             return False
 
     return True


pymedphys/_dicom/anonymise.py
```diff
@@ -466,10 +467,13 @@ def is_anonymised_dataset(ds, ignore_private_tags=False):
                     float(elem.value), float(dummy_value)
                 ):
                     continue
-
+                logging.info("%s is not considered to be anonymised", elem.name)
```
SimonBiggs, line 470: See the following table:


Level | When it’s used
-- | --
DEBUG | Detailed information, typically of interest only when diagnosing problems.
INFO | Confirmation that things are working as expected.
WARNING | An indication that something unexpected happened, or indicative of some problem in the near future (e.g. ‘disk space low’). The software is still working as expected.
ERROR | Due to a more serious problem, the software has not been able to perform some function.
CRITICAL | A serious error, indicating that the program itself may be unable to continue running.

from -- https://docs.python.org/3/howto/logging.html#when-to-use-logging

Based on that table, is info the right logging level here?

sjswerdloff: I believe so.  Things are working as expected for the programmer.
It isn't the library programmer who is debugging whether their data is anonymised or not.
It is the user or API programmer who wants to understand why their data is not considered anonymised.
is_anonymised_dataset is a public function.  If it were private (e.g. _is_anonymised_dataset() and only intended to be called by testing code, or was itself test code only), I'd still consider it to be info level if one expected the return value to be true in the vast majority of the typical use cases.

There are a number of different arguments to the CLI that would result in different behaviour operating on the same data.  So the use of varying arguments might result in outcomes that the user might need to understand, but the library programmer would consider to be absolutely correct behaviour that does not require debugging.  The library programmer would think that there might be a user training issue (assuming The Fine Manual was clear, and simply had not been Read...)

Consider what you would do if someone said "this isn't working right, my data is being called not anonymised, but I anonymised it according to the same rules as pymedphys as far as I can tell".  Yes, you could tell them to put it in debug logging and then slog through the debug logs.  Or you could ask them to log at the info level to confirm that things are operating as expected.

The distinction I make is whether the "problems" are unexpected or not for considering it DEBUG.
DEBUG content is only beautiful to the programmer who put the content in.  Stuff is not working as expected so the programmer needs to know what happened before, what happened after, and needs significant context to understand what might be going on.
INFO content should be meaningful to a service engineer or support person, or knowledgable user.

The fact that I ran in to this while acting as a library programmer and found a defect in the code is separate from the need for a user to be able to receive confirmation that things (a dataset is found not to be anonymised) are working as expected (by being informed of the reason why the dataset is not considered to be anonymised).

SimonBiggs:
Just quickly responding on my phone to one point there. Given this is
within a private module _dicom, everything in here is actually a private
function.

 The only anonymised function that forms part of the public API of
pymedphys is "anonymise":

https://github.com/pymedphys/pymedphys/blob/56fd325ccdc91b1de682fdb669f3f78d80d4f401/pymedphys/dicom.py#L6

And it is accessible at pymedphys.dicom.anonymise

SimonBiggs:
And yup, that logic all makes sense to me. From that point of view, do you
think you'd hope to have filename and line details when in INFO mode?

sjswerdloff:
> And yup, that logic all makes sense to me. From that point of view, do you think you'd hope to have filename and line details when in INFO mode?
> […](#)

No.  It's enough to understand why something was considered to not be anonymised.
The user or support person would not need to know where the code was.  Just which element was involved.

If this were troubleshooting for a dataset issue that was not intended to be anonymised, having the SOP Instance UID in the log message would be useful (to uniquely identify which SOP instance was involved if there was more than one dataset involved).  But because this is anonymisation, one might already have anonymised the UID (not the case currently, but it is clearly something mentioned as a goal, and a high order goal for pseudonymisation).  However... one could make an argument for referring to the SOP Instance UID available in this scenario as well.  

And I stand corrected on the content of the module as a whole being private (what might be referred to as "internal" in other programming languages).  Excellent reminder to me to keep in mind the import/namespace hierarchy!

SimonBiggs:
> If this were troubleshooting for a dataset issue that was not intended to be anonymised, having the SOP Instance UID in the log message would be useful (to uniquely identify which SOP instance was involved if there was more than one dataset involved).

On this one I'm happy to defer to your judgement regarding whether or not it should be included.

> Externalizing the logging configuration in a configuration file (with a pragmatic set of defaults) allows the logging to be "all things to all people".

When it's being called as a python library (via `pymedphys.pydicom.anonymise`) the user is free to configure the logging themselves. Regarding the CLI, I agree, it would be very neat to have it within a configuration file.

I actually have a config.toml file set up at `$HOME/.pymedphys/config.toml`. This is growing to become the central place to store all `pymedphys` configuration. An example configuration file can be found at https://github.com/pymedphys/pymedphys/blob/master/site-specific/cancer-care-associates/config.toml

It would make sense to have an entry that looks like the following:

```toml
[cli]

	[cli.logging]
	format = "%(asctime)s %(levelname)-8s %(message)s"
	level = "warning"
	date_format = "%Y-%m-%d %H:%M:%S"
```

Is that something you think we should implement? Did you want me to do that?

SimonBiggs:
@sjswerdloff LGTM, feel free to merge when you're ready

SimonBiggs:
> Is that something you think we should implement? Did you want me to do that?

I liked the idea... I decided to dive in and just implement it. See https://github.com/pymedphys/pymedphys/pull/942

PR #942, opened 2020-07-20
Add logging configuration

SimonBiggs:
Based on the idea presented at https://github.com/pymedphys/pymedphys/pull/940#issuecomment-660902155.

In the end, supported all keyword arguments that `logging.basicConfig` supports, and did this by passing the configuration through. An example `$HOME/.pymedphys/config.toml` would be:

```toml
[cli]

	[cli.logging]
	level = "info"	
	format = "%(asctime)s %(levelname)-8s %(message)s"
	datefmt = "%Y-%m-%d %H:%M:%S"
```

### DIFF
--- a/pymedphys/_vendor/patchlogging.py
+++ b/pymedphys/_vendor/patchlogging.py
@@ -28,7 +28,9 @@ _basicConfig = logging.basicConfig
 
 
 def apply_logging_patch():
-    if sys.version_info.major > 3 or sys.version_info.minor >= 8:
+    if sys.version_info.major > 3 or (
+        sys.version_info.major == 3 and sys.version_info.minor >= 8
+    ):
         return
 
     global _patch_applied  # pylint: disable = global-statement
--- a/pymedphys/_vendor/patchlogging.py
+++ b/pymedphys/_vendor/patchlogging.py
@@ -19,6 +19,7 @@
 # https://github.com/python/cpython/blob/a667e1c66a/Lib/logging/__init__.py#L1957-L1961
 
 import logging
+import sys
 
 from pymedphys import _config
 
@@ -27,6 +28,9 @@ _basicConfig = logging.basicConfig
 
 
 def apply_logging_patch():
+    if sys.version_info.major > 3 or sys.version_info.minor >= 8:
+        return
+
     global _patch_applied  # pylint: disable = global-statement
 
     if not _config.is_cli:
--- a/pymedphys/_vendor/patchlogging.py
+++ b/pymedphys/_vendor/patchlogging.py
@@ -33,8 +33,8 @@ def apply_logging_patch():
         raise ValueError(
             "This patch globally adjusts the logging module. This patch is "
             "not to be used within pymedphys while it is in library mode "
-            "as standard library monkey patches can be exceptionally confusing "
-            "to a library user. "
+            "as monkey patches on libraries/modules external to pymedphys "
+            "can be exceptionally confusing to a library user. "
             "This is only to be used when pymedphys is being called via "
             "CLI."
         )
--- a/pymedphys/_vendor/patchlogging.py
+++ b/pymedphys/_vendor/patchlogging.py
@@ -1,3 +1,4 @@
+# Copyright (C) 2020 Simon Biggs
 # Copyright 2001-2017 by Vinay Sajip. All Rights Reserved.
 #
 # Permission to use, copy, modify, and distribute this software and its
--- a/pymedphys/_vendor/patchlogging.py
+++ b/pymedphys/_vendor/patchlogging.py
@@ -36,7 +36,7 @@ def apply_logging_patch():
             "as monkey patches on libraries/modules external to pymedphys "
             "can be exceptionally confusing to a library user. "
             "This is only to be used when pymedphys is being called via "
-            "CLI."
+            "the pymedphys CLI."
         )
 
     if _patch_applied:
--- a/pymedphys/_vendor/patchlogging.py
+++ b/pymedphys/_vendor/patchlogging.py
@@ -19,17 +19,34 @@
 
 import logging
 
+from pymedphys import _config
+
+_patch_applied = False
 _basicConfig = logging.basicConfig
 
 
-def basicConfig(**kwargs):
-    force = kwargs.pop("force", False)
-    if force:
-        for h in logging.root.handlers[:]:  # pylint: disable = no-member
-            logging.root.removeHandler(h)  # pylint: disable = no-member
-            h.close()
+def apply_logging_patch():
+    global _patch_applied  # pylint: disable = global-statement
 
-    _basicConfig(**kwargs)
+    if not _config.is_cli:
+        raise ValueError(
+            "This patch globally adjusts the logging module. This patch is "
+            "not to be used within pymedphys while it is in library mode. "
+            "This is only to be used when pymedphys is being called via "
+            "CLI."
+        )
 
+    if _patch_applied:
+        raise ValueError("This patch has already been applied.")
 
-logging.basicConfig = basicConfig
+    def basicConfig(**kwargs):
+        force = kwargs.pop("force", False)
+        if force:
+            for h in logging.root.handlers[:]:  # pylint: disable = no-member
+                logging.root.removeHandler(h)  # pylint: disable = no-member
+                h.close()
+
+        _basicConfig(**kwargs)
+
+    logging.basicConfig = basicConfig
+    _patch_applied = True
--- a/pymedphys/_vendor/patchlogging.py
+++ b/pymedphys/_vendor/patchlogging.py
@@ -31,7 +31,9 @@ def apply_logging_patch():
     if not _config.is_cli:
         raise ValueError(
             "This patch globally adjusts the logging module. This patch is "
-            "not to be used within pymedphys while it is in library mode. "
+            "not to be used within pymedphys while it is in library mode "
+            "as standard library monkey patches can be exceptionally confusing "
+            "to a library user. "
             "This is only to be used when pymedphys is being called via "
             "CLI."
         )
--- a/pymedphys/_vendor/patchlogging.py
+++ b/pymedphys/_vendor/patchlogging.py
@@ -15,8 +15,8 @@
 # IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT
 # OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
-# Code vendored from:
-# https://github.com/python/cpython/blob/a667e1c66a62d509c39d30abf11778213a1e1ca0/Lib/logging/__init__.py#L1896-L1999
+# Code vendored and modified from:
+# https://github.com/python/cpython/blob/a667e1c66a/Lib/logging/__init__.py#L1957-L1961
 
 import logging
 
--- a/pymedphys/cli/__init__.py
+++ b/pymedphys/cli/__init__.py
@@ -85,11 +85,6 @@ def run_logging_basic_config(args):
 
     if "level" not in logging_config:
         logging_config["level"] = logging.WARNING
-    else:
-        try:
-            logging_config["level"] = getattr(logging, logging_config["level"].upper())
-        except AttributeError:
-            pass
 
     # Allow command line options to override the config.toml options
     if args.logging_verbose:
@@ -109,7 +104,7 @@ def run_logging_basic_config(args):
     if "datefmt" not in logging_config:
         logging_config["datefmt"] = "%Y-%m-%d %H:%M:%S"
 
-    logging.basicConfig(**logging_config)
+    logging.basicConfig(**logging_config, force=True)
 
     logging.info(
         "Set `logging.basicConfig` with:\n%(logging_config)s",
--- a/pymedphys/cli/__init__.py
+++ b/pymedphys/cli/__init__.py
@@ -113,7 +113,7 @@ def run_logging_basic_config(args):
     if args.logging_debug and args.logging_verbose:
         logging.warning(
             "Both --verbose and --debug were defined. Verbose mode was "
-            "ignored and instead --debug was used."
+            "ignored. Debug mode has been used."
         )
 
     logging.info(
--- a/pymedphys/cli/__init__.py
+++ b/pymedphys/cli/__init__.py
@@ -80,15 +80,18 @@ def get_logging_config():
     return logging_config
 
 
-def run_logging_basic_config(args):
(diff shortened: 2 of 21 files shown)


sjswerdloff:
While this works for basicConfig, https://docs.python.org/3/library/logging.config.html provides for a richer means of addressing configuration.
Personally, I would favour referencing a separate configuration file from the config.toml (if file config is intended).
basicConfig (or at least it's use here) seems like it is a dictConfig() based approach, so it should be feasible to go with the current PR, and then enhance later with a fileConfig() based approach or even a listen() based approach.  The listen() based approach is useful for long running applications (dynamically enable increasingly verbose logging without having to restart the application, e.g. from within an iCOM listener).

SimonBiggs:
Ooo, that is neat. I'll have a read. I don't think this'll be too hard to
get right the first time. Lets go through a few iterations to make it
support that live case.

SimonBiggs:
> Personally, I would favour referencing a separate configuration file from the config.toml (if file config is intended).
> basicConfig (or at least it's use here) seems like it is a dictConfig() based approach, so it should be feasible to go with the current PR, and then enhance later with a fileConfig() based approach or even a listen() based approach. The listen() based approach is useful for long running applications (dynamically enable increasingly verbose logging without having to restart the application, e.g. from within an iCOM listener).

So, a few thoughts here after doing a bit of reading. I'm quite keen to have the `config.toml` file be the "reference" configuration location. I do have features in there such as redirect, which allows the `$HOME/.pymedphys/config.toml` file to have a `redirect` key which allows the user to specify a different configuration file:

https://github.com/pymedphys/pymedphys/blob/90bf2532b79b1e4a3d96e0974f0957abfd56836e/pymedphys/_config.py#L36-L43

I'd like to stick to having all configuration be able to be done in that file, more than happy to be able to support multiple `redirects`, as in, have the top level configuration point to a range of files which each are merged (and verified to not have any conflicting keys). This would allow the user to split the config file up as needed.

Justifications for using the `toml` format can be seen at the following link:

https://www.python.org/dev/peps/pep-0518/#file-format

All that said, it seems if we go with the `fileConfig` approach we will be pigeon-holed into using a `.ini` format in a standalone file losing the ability to have a common location for all configuration.

One thought regarding the ability to live update logging during a program's running may be to set up live listening to the `config.toml` file itself? This could have some serious benefits for other parts of PyMedPhys. For example, running `pymedphys gui`, it would be quite neat if one could update the `config.toml` file and have the effects of the configuration be represented within the GUI live. The pymedphys CLI (or another GUI) could then freely write to this file, and its changes would be automatically propagated into the running software live.

To achieve this file listening one can use watchdog (https://python-watchdog.readthedocs.io/).

To internally achieve this propegation, `get_config` could return a BehaviorSubject (https://rxpy.readthedocs.io/en/latest/reference_subject.html#rx.subject.BehaviorSubject) instead of a dictionary. This observable subject could be set to update whenever watchdog triggers that the file has changed. Internally within the library, this observable configuration can then be subscribed to in order to propagate the configuration changes live.

What do you think of this design approach?

Cheers,
Simon

SimonBiggs:
If you're interested, some more details on file format choices:

https://gist.github.com/njsmith/78f68204c5d969f8c8bc645ef77d4a8f

sjswerdloff:
if you can get listen() to respond to the updated config.toml file or achieve the equivalent, I think the approach has a lot of value.
I like event driven approaches.
But keep in mind that in a networked environment, you could have your home directory accessed by a number of different hosts (you can be working on a handful of Unix systems simultaneously...) so you may have the same config.toml unintentionally acting on multiple different systems.
You also then require that the modifications to the configuration be driven by the user.
Trade off if you want to have someone else come in and do support (they shouldn't have to "be" you".

SimonBiggs:
> But keep in mind that in a networked environment, you could have your home directory accessed by a number of different hosts (you can be working on a handful of Unix systems simultaneously...) so you may have the same config.toml unintentionally acting on multiple different systems.

I hadn't thought of this. I haven't actually seen an environment set up like this. This could be frustrating. Any thoughts on how best to handle this?

> You also then require that the modifications to the configuration be driven by the user.
> Trade off if you want to have someone else come in and do support (they shouldn't have to "be" you".

This one has an easy fix I think, for example, on all the computers at work that use this config file I have a `redirect = '\\path\to\some\common\network\directory'`. That way, the config file being used by each system is a common one to all users across the sites.

SimonBiggs:
Regarding storing the config file in the home directory, on further thought, should one home directory for a given user be spread across multiple machines there are going to be all sorts of software that has the issue described. For example the bashrc file is stored in home, x configuration is in home, in the python world, Jupyter, pip, streamlit, and others store their config in home.

...my thoughts are, maybe a big difference is that those other tools also allow the user to provide the path to a config file directly within the CLI. Maybe that's the workaround in those cases. Allow the user to customise the config location by passing a path via a CLI option?

What do you think?

sjswerdloff:
Yes.  Have a default and allow configuration of where the config file is by environment variable.

Multi platform environment was common back in the day when engineering workstations were playing leap frog, and successive purchases resulted in mixed environments and machines with different performance and software licenses.

SimonBiggs:
> Yes. Have a default and allow configuration of where the config file is by environment variable.

Yup, I like that. That way the user can set the environment variable in a boot script if need be to cover weird cases.

> Multi platform environment was common back in the day when engineering workstations were playing leap frog, and successive purchases resulted in mixed environments and machines with different performance and software licenses.

Yup, makes sense. In the past I have tried to restore a complete home directory to a new Ubuntu install... it was messy. But I can see the need.

So, what do you think, let's merge this one then as the first step in a few other PRs. Subsequent PRs will:

* Allow watchdog and RxPy to observe the config file in a subscribable way.
* Have the logging configuration subscribe to changes to the config file and have it reset the logging configuration when that observable fires.
* Have the `config.toml` path be setable via both environment variable and CLI parameter. Have CLI parameter override environment variable.
* Change the `config.toml` logic to use `imports` which accepts a list of filepaths which will be pulled into the configuration to allow users to split their config file across a range of files, while still keeping common configurations in a core config file, and flip between them depending on production/development/staging environments, for example.

pymedphys/cli/__init__.py
```diff
@@ -63,22 +65,63 @@ def define_parser():
...
+def run_logging_basic_config(args):
+    logging_config = get_logging_config()
 
+    if "level" not in logging_config:
+        logging_config["level"] = logging.WARNING
+    else:
+        try:
+            logging_config["level"] = getattr(logging, logging_config["level"].upper())
+        except AttributeError:
+            pass
```
sjswerdloff, line 92: if someone misconfigures the level of logging, e.g. they put in warnning (I intentionally misspelled that... as an example of what someone might do) or status (a logging level available in NLog, Log4J, and other related logging frameworks), then... what happens?
I guess this would default to WARNING because that's what logging defaults to?
It looks like printing (because logging isn't working yet) the AttributeError content would be a kindness, e.g.:

except AttributeError as error:
    print(str(error))

or start adding to a list of errors found during logging configuration build up, and then log them all once the basicConfig is finished, and do so at the error level...

alternatively, I think one could just allow the inappropriate string value to be assigned to the (local variable) logging_config dict and leave it to basicConfig (which I think is a wrapper of dictConfig) to report the error.  But in that scenario, the misconfigured value would not show up if --verbose or --debug was in use on the command line.

pymedphys/cli/__init__.py
```diff
@@ -63,22 +65,63 @@ def define_parser():
...
+        logging_config["format"] = "%(asctime)s %(levelname)-8s %(message)s"
 
-    logging.basicConfig(format=logformat, level=loglevel, datefmt="%Y-%m-%d %H:%M:%S")
+        if logging_config["level"] <= logging.DEBUG:
+            logging_config["format"] += "\n    %(pathname)s#%(lineno)d"
+
+    if "datefmt" not in logging_config:
+        logging_config["datefmt"] = "%Y-%m-%d %H:%M:%S"
+
+    logging.basicConfig(**logging_config)
```
sjswerdloff, line 112: If you intend to call this method more than once during the lifetime of the cli "instance", I think you want to pass force=True.  I'm not 100% sure about that, but that's my read of
https://docs.python.org/3.8/library/logging.html#logging.basicConfig
Given that it includes "This function does nothing if the root logger already has handlers configured, unless the keyword argument force is set to True."

Given that this method will only be called once at the moment, that's not a defect.
If this method will only ever be called once, it's not a defect.
Otherwise it's a latent defect.

pymedphys/cli/__init__.py
SimonBiggs: > if someone misconfigures the level of logging, e.g. they put in warnning (I intentionally misspelled that... as an example of what someone might do) or status (a logging level available in NLog, Log4J, and other related logging frameworks), then... what happens?
> I guess this would default to WARNING because that's what logging defaults to?

It wouldn't default to WARNING it would just leave `logging_config["level"]` as it began (exactly as it appeared in the config file), in the example you gave it would try pass `level='warnning'` to basicConfig, resulting in the following:

```python
In [1]: import logging                                                                    

In [2]: logging.basicConfig(level='warnning')                                             
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-954734a85d38> in <module>
----> 1 logging.basicConfig(level='warnning')

~/.pyenv/versions/3.7.8/lib/python3.7/logging/__init__.py in basicConfig(**kwargs)
   1914             level = kwargs.pop("level", None)
   1915             if level is not None:
-> 1916                 root.setLevel(level)
   1917             if kwargs:
   1918                 keys = ', '.join(kwargs.keys())

~/.pyenv/versions/3.7.8/lib/python3.7/logging/__init__.py in setLevel(self, level)
   1351         Set the logging level of this logger.  level must be an int or a str.
   1352         """
-> 1353         self.level = _checkLevel(level)
   1354         self.manager._clear_cache()
   1355 

~/.pyenv/versions/3.7.8/lib/python3.7/logging/__init__.py in _checkLevel(level)
    190     elif str(level) == level:
    191         if level not in _nameToLevel:
--> 192             raise ValueError("Unknown level: %r" % level)
    193         rv = _nameToLevel[level]
    194     else:

ValueError: Unknown level: 'warnning'
```

Which I think is sufficiently informative :).

But, my whole point of that code was to allow users to provide an integer number such as `10` or text value such as `'debug'`, but turns out `basicConfig` already supports that out of the box. So it was superfluous therefore, I removed that piece of code all together.

pymedphys/cli/__init__.py
SimonBiggs: Yup good point, it's not currently a defect, but if I forget to change it later, it would be a defect then. Changed :). Cheers :).

SimonBiggs:
Ahh bumma:

> Changed in version 3.8: The force argument was added.

-- https://docs.python.org/3/library/logging.html#logging.basicConfig

For now can't use basicConfig like this...

SimonBiggs:
@sjswerdloff, just re-pinged for review as I made a few more changes after you approved.

Cheers :)

SimonBiggs:
Thanks for the review by the way :)

image

pymedphys/cli/__init__.py
```diff
@@ -85,11 +85,6 @@ def run_logging_basic_config(args):
 
     if "level" not in logging_config:
         logging_config["level"] = logging.WARNING
-    else:
```
sjswerdloff: The attribute search/validation may be needed.  DictConfig (not available until 3.2) states that it will validate and raise specific Errors.
basicConfig does not say it will raise those errors, nor does it say what it will do if provided a level it doesn't understand, either way.  I had assumed that if DictConfig would raise the errors, basicConfig would as well, but now I don't think that's a valid assumption.  I don't know if basicConfig is a wrapper around DictConfig, but now looking at the evolution of logging configuration, I'm not willing to assume they refactored to ensure DRY.
And separate from the validation, this change also resulted in loss of using the level specified in the config.toml file (e.g. ERROR, if someone really doesn't want much logging, or INFO if they didn't want to have to use the command line --verbose option to get INFO level logging...)

sjswerdloff:
> @sjswerdloff, just re-pinged for review as I made a few more changes after you approved.
> 
> Cheers :)

Shame, I didn't realize force=True was limited to 3.8.  I missed that (it did say it).
But having gone back through various versions of the documentation, and then walked through the tutorials and cookbook with a more cynical eye, I think the incorporation of the logging level specified in the config.toml file should go back in, and so should the attribute checking (I hope my review comment on the obsolete commit shows up for you, I was trying to point out the code that I thought should get brought back).

Sorry...

SimonBiggs:
> And it's unfortunate that one can not "un-approve". BitBucket allows that...

Ahh, you can, just click "request changes"

SimonBiggs:
> Sorry...

No need for the sorry :) massively appreciate the review

SimonBiggs:
> Shame, I didn't realize force=True was limited to 3.8. I missed that (it did say it).

The point still stands nevertheless. It needs to be addressed if I am to recall that function when the `config.toml` file changes.

pymedphys/cli/__init__.py
SimonBiggs: In the end, I believe the main reason I needed this code back in so that the following line worked:

https://github.com/pymedphys/pymedphys/blob/fc9063d992ad93f65a3910c69b5e12a6660d62cf/pymedphys/cli/__init__.py#L105

SimonBiggs:
I think I have covered the issues you raised :) I also included some tests that really should have been there from the beginning in hindsight.

sjswerdloff:
Feel free to Merge...
LGTM

SimonBiggs:
> Would it be possible to use the logging.DEBUG constant instead of 10 and similar for the other defined integer values?

Do you mean in the config file or in the test?

I can certainly swap over to referring to the constant in the test, that makes sense. And in the config file if "debug" or "DEBUG" was given that would work too.

sjswerdloff:
>Do you mean in the config file or in the test?
Test

SimonBiggs:
Yup, thanks. I hope the monkey patch wasn't too crazy an idea...

pymedphys/_vendor/patchlogging.py
```diff
@@ -27,6 +28,9 @@
 
 
 def apply_logging_patch():
+    if sys.version_info.major > 3 or sys.version_info.minor >= 8:
```
sjswerdloff: This only works because Python 2.7 was end of line.
As a rule of thumb, I would take the following approach without having prior knowledge of where earlier versions stopped incrementing their minor version.

if sys.version_info.major > 3 or (sys.version_info.major == 3 and sys.version_info.minor >= 8):

SimonBiggs:
Yeah, as soon as I did that one I realised the error in my ways.

PR #945, opened 2020-07-23
recurse through sequences to check anonymisation of a dataset

sjswerdloff:
For each sequence item in a sequence descend and identify any elements not considered to be anonymised.
Log the name of the sequence itself (for each sequence item).
Potential improvements would be to:
identify the index (in order of iteration) which sequence item(s) is/(are) involved
only log the sequence name once if there are multiple sequence items that are not considered to be anonymised.


sjswerdloff:
Agreed.
Incorporated your suggestion nearly verbatim.

sjswerdloff:
SnoopyHappyDance

SimonBiggs:
Cool Snoopy

PR #951, opened 2020-07-24
Diagnose on-update test error

SimonBiggs:
@sjswerdloff, if you're interested here is my process to isolate the on-updates issue. If there is an issue I can pass it upstream to the corresponding dependency.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -195,10 +195,10 @@ description = "The AWS SDK for Python"
 name = "boto3"
 optional = false
 python-versions = "*"
-version = "1.14.20"
+version = "1.14.27"
 
 [package.dependencies]
-botocore = ">=1.17.20,<1.18.0"
+botocore = ">=1.17.27,<1.18.0"
 jmespath = ">=0.7.1,<1.0.0"
 s3transfer = ">=0.3.0,<0.4.0"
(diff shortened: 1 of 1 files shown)


SimonBiggs:
btw, the `poetry.lock` file is created by running `poetry update`. So this was essentially:

```bash
git checkout master
git pull
git checkout -b new-branch-name
poetry update
git add poetry.lock
git commit -m "commit message"
git push
```

sjswerdloff:
While I see what has been done, I still don't understand why the mypy hash (SHA digest) mismatch was called out and yet this was resolved by making updates that don't appear to include updating the version of mypy or the expected hash for it.
There was a warning in the error message that suggested one of the possible reasons for this could be that a package had been hacked.

SimonBiggs:
Yeah I was surprised it was fixed by just running the update. Essentially
the hash didn't match what was expected on pypi. A man-in-the-middle attack
is one option. But a mistake/network error in storing the very new hash
value could also be the cause.

I find the Occam's razor for this sort of thing is the likelihood of
malicious intent < the likelihood of a mistake. The fact that it's fixed
had me put it down to a broken hash somewhere in the chain that was later
fixed.

PR #957, opened 2020-07-29
assign dtype=object as per warnings for numpy array creation of ragged arrays

sjswerdloff:
this might not be the desired approach if there is something deeper going on with the creation of the arrays or if there is significant impact in usage of the array with dtype=object assigned.
However, the warning during unit testing is clear that this code will fail to build/run in a near future version of numpy without some kind of change.
I just got tired of seeing the warnings while in a tight modify/unit test cycle

### DIFF
--- a/pymedphys/_experimental/fileformats/mephysto/api.py
+++ b/pymedphys/_experimental/fileformats/mephysto/api.py
@@ -48,8 +48,8 @@ def load_mephysto(filepath, output_to_file=False, output_directory=None, sort=Tr
     scan_depth = pull_mephysto_number("SCAN_DEPTH", file_contents)
 
     # Convert python lists into numpy arrays for easier use
-    distance = np.array(distance)
-    relative_dose = np.array(relative_dose)
+    distance = np.array(distance, dtype=object)
+    relative_dose = np.array(relative_dose, dtype=object)
     scan_curvetype = np.array(scan_curvetype)
     scan_depth = np.array(scan_depth)
 
--- a/pymedphys/tests/dicom/test_coords.py
+++ b/pymedphys/tests/dicom/test_coords.py
@@ -63,7 +63,7 @@ def run_xyz_function_tests(coord_system):
     for orient, dicom in test_ds_dict.items():
         test_xyz = coords.xyz_axes_from_dataset(dicom, coord_system)
 
-        expected_xyz[orient] = np.array(expected_xyz[orient])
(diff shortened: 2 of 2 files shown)


SimonBiggs:
> this might not be the desired approach if there is something deeper going on with the creation of the arrays or if there is significant impact in usage of the array with dtype=object assigned.

In fact, these cases are primarily a misuse of numpy. One learns quite a bit in ~4 years of coding and I wouldn't write that mephysto loading code in the same way should I write it today.

Nevertheless, both of these cases have some rudimentary tests, and this change doesn't break those tests. So good to go from me.

Thanks @sjswerdloff :)

sjswerdloff:
Post merge error:
 - Installing pydicom (2.0.0)
101

102
[EnvCommandError]
103
Command ['D:\\a\\pymedphys\\pymedphys\\.venv\\Scripts\\pip.exe', 'install', '--no-deps', 'pydicom==2.0.0'] errored with the following return code 1, and output: 
104
Collecting pydicom==2.0.0
105

106
  Downloading pydicom-2.0.0-py3-none-any.whl (35.4 MB)
107

108
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
109

110
    pydicom==2.0.0 from https://files.pythonhosted.org/packages/d3/56/342e1f8ce5afe63bf65c23d0b2c1cd5a05600caad1c211c39725d3a4cc56/pydicom-2.0.0-py3-none-any.whl#sha256=667c5bf9ca52f440e538c9d03ce86b04555c10d069472a5db53103fe40d310c0:
111

112
        Expected sha256 667c5bf9ca52f440e538c9d03ce86b04555c10d069472a5db53103fe40d310c0
113

114
             Got        90fa1ac365d99cf527453f255453b8a88297eb7cba45c4fcdc402fd0371b6b54
115

116

sjswerdloff:
I will review the mechanics of addressing this issue as described in a previous post-merge error and attempt to fix it.

sjswerdloff:
re-run of all jobs of Tests (windows, tests) resulted in Complete job (no errors).
I think that means it's safe to assume there was some kind of momentary network glitch.
There was no need to update the poetry.lock

SimonBiggs:
Hmm, maybe there's an issue with how `poetry` is handling network glitches. Ideally it should attempt a re-download or something of that sort. I'll have a bit of an investigation.

Cheers,
Simon

PR #959, opened 2020-07-31
Use strategy to anonymise

sjswerdloff:
This is a design proposal:
instead of hardcoding a specific anonymisation approach (albeit with a few control variables that offer variations on the theme), explicitly specify a strategy for doing so.
That way, if someone has a different strategy for anonymisation, they can provide it (an inversion of control style approach).
I have not made modifications to the calling stack to provide the mechanism for passing in an alternative strategy yet.
I just wanted to show the idea (and invoke it by default to ensure it got tested by the unit/automatic tests) with a minimum of code.
By itself, this is a very inefficient way of doing what was already being done (adds two levels of indirection), but with this approach, one gains the flexibility to implement pseudonymisation (as an alternative anonymisation strategy).

Note that this also identifies a TODO that is separate from the strategy design discussion, which is appropriate handling of VR of CS where "ANON" or "" might break interoperability, or worse, DICOM conformance.

### DIFF
--- a/pymedphys/_dicom/anonymisation_strategy.py
+++ b/pymedphys/_dicom/anonymisation_strategy.py
@@ -14,7 +14,7 @@
 
 import functools
 
-import pydicom
+from pymedphys._imports import pydicom
 
 from pymedphys._dicom.constants import PYMEDPHYS_ROOT_UID
 
--- a/pymedphys/_dicom/anonymisation_strategy.py
+++ b/pymedphys/_dicom/anonymisation_strategy.py
@@ -1,5 +1,5 @@
-# Copyright (C) 2020 Stuart Swerdloff, Matthew Jennings, Simon Biggs
-
+# Copyright (C) 2020 Stuart Swerdloff
+# Copyright (C) 2018 Matthew Jennings, Simon Biggs
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
--- a/pymedphys/_dicom/anonymisation_strategy.py
+++ b/pymedphys/_dicom/anonymisation_strategy.py
@@ -56,9 +56,12 @@ def _get_vr_anonymous_hardcode_replacement_value(current_value, value_representa
     return get_vr_anonymous_replacement_value_dict()[value_representation]
 
 
-ANONYMISATION_HARDCODE_DISPATCH = {
-    key: functools.partial(
-        _get_vr_anonymous_hardcode_replacement_value, value_representation=key
-    )
-    for key in get_vr_anonymous_replacement_value_dict().keys()
-}
+@functools.lru_cache(maxsize=1)
+def get_default_hardcode_dispatch():
+    ANONYMISATION_HARDCODE_DISPATCH = {
+        key: functools.partial(
+            _get_vr_anonymous_hardcode_replacement_value, value_representation=key
+        )
+        for key in get_vr_anonymous_replacement_value_dict().keys()
+    }
+    return ANONYMISATION_HARDCODE_DISPATCH
--- a/pymedphys/_dicom/anonymisation_strategy.py
+++ b/pymedphys/_dicom/anonymisation_strategy.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2020 Stuart Swerdloff
+# Copyright (C) 2020 Stuart Swerdloff, Simon Biggs
 # Copyright (C) 2018 Matthew Jennings, Simon Biggs
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -46,7 +46,7 @@ def get_vr_anonymous_replacement_value_dict():
     return VR_ANONYMOUS_REPLACEMENT_VALUE_DICT
 
 
-def _get_vr_anonymous_hardcode_replacement_value(value_representation, current_value):
+def _get_vr_anonymous_hardcode_replacement_value(current_value, value_representation):
     """A single dispatch function that is used for any VR with the current_value of the element ignored
     This is the default for the replacement strategy
     """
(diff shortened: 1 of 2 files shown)


pymedphys/_dicom/anonymisation_strategy.py
```diff
@@ -0,0 +1,151 @@
+# Copyright (C) 2020 Stuart Swerdloff, Matthew Jennings, Simon Biggs
```
SimonBiggs, line 1: Let's change this here to:

```python
# Copyright (C) 2020 Stuart Swerdloff
# Copyright (C) 2018 Matthew Jennings, Simon Biggs
```

To indicate that it is borrowing from work done in 2018, but the 2020 updates were yours.

pymedphys/_dicom/anonymisation_strategy.py
```diff
@@ -0,0 +1,151 @@
...
+        "OW": _get_OW_anonymous_hardcode_replacement_value,
+        "PN": _get_PN_anonymous_hardcode_replacement_value,
+        "SH": _get_SH_anonymous_hardcode_replacement_value,
+        "SQ": _get_SQ_anonymous_hardcode_replacement_value,
+        "ST": _get_ST_anonymous_hardcode_replacement_value,
+        "TM": _get_TM_anonymous_hardcode_replacement_value,
+        "UI": _get_UI_anonymous_hardcode_replacement_value,
+        "US": _get_US_anonymous_hardcode_replacement_value,
+    }
+)
```
SimonBiggs, line 150: To remove the duplicated logic that is found in lines 58-127 can we change lines 58-151 to something like the following:

```python
KEYS_TO_OVERRIDE = [
    "AE", "AS", "CS", "DA", "DS", "DT",
    "LO", "LT", "OB", "OB or OW", "OW",
    "PN", "SH", "SQ", "ST", "TM", "UI", 
    "US"
]

ANONYMISATION_HARDCODE_DISPATCH = {
    key: functools.partial(_get_vr_anonymous_hardcode_replacement_value, value_representation=key)
    for key in KEYS_TO_OVERRIDE 
}
```

pymedphys/_dicom/anonymise.py
```diff
@@ -26,6 +26,7 @@
 from pymedphys._imports import numpy as np
```
SimonBiggs, line 25: I think it's worth adding your copyright to the top of this file too.

pymedphys/_dicom/anonymise.py
```diff
@@ -642,9 +619,50 @@ def _filter_identifying_keywords(keywords_to_leave_unchanged):
     return keywords_filtered
 
 
-def get_anonymous_replacement_value(keyword):
-    """Get an appropriate dummy anonymisation value for a DICOM element
+def get_anonymous_replacement_value(
+    keyword,
+    current_value=None,
+    replacement_strategy=anon.ANONYMISATION_HARDCODE_DISPATCH,
```
SimonBiggs, line 623: I really like this design. It's really nice. It makes my mind happy.

pymedphys/_dicom/anonymise.py
```diff
@@ -642,9 +619,50 @@ def _filter_identifying_keywords(keywords_to_leave_unchanged):
...
     based on its value representation (VR)
+    Parameters
+    ----------
+    keyword: text string that is the pydicom name for the DICOM attribute/element
+
+    current_value: optional, the value that is currently assigned to the element
+
+    replacement_strategy: optional, a dispatch dictionary whose keys are the text representation
+    of the VR of the element, and whose values are function references that take the current value
+    of the element.
```
SimonBiggs, line 635: Might be worth making reference here to where the user can import a list of keys that need to be override (eg the `KEYS_TO_OVERRIDE` defined in a previous review comment).

Do you think, if a certain key is not defined we should raise an error? Or should it fall back to the default override? I can see both pros and cons to each approach.

pymedphys/_dicom/anonymise.py
SimonBiggs: On a separate point though, it is best to avoid mutable default parameters, see the following:

https://docs.python-guide.org/writing/gotchas/#what-you-should-do-instead

sjswerdloff: I'm happy to get rid of the mutable default parameters, but that was a useful temporary mechanism for conveying the concept as well as exercising the pathway.  So I will change that soon

pymedphys/_dicom/anonymise.py
sjswerdloff: they keys to the dispatch dictionary are the VR.  And there are VR in DICOM that are not in the dispatch dictionary, and that should *definitely* get checked (e.g. if vr not in dispatch_dict).  If that shows up, it's an error.  Because someone put in a keyword whose VR is not being addressed.

if you mean keywords to be replaced not including particular keywords that "we know should be included"... logging a warning seems appropriate, but not creating an error, and the log should identify the keyword but not the current value because of the potential to leak PHI in to the log.  If a user has specified a key to be ignored (not replaced), that is their prerogative.   I had to do that with PatientSex (because ANON breaks conformance for that attribute).

>import a list of keys that need to be overridden
I will have to hunt for that previous review comment.
It would be good to offer configuration for additional keywords to anonymise, but I don't want to try to do that here/yet.
I think UIDs need to be handled differently/separately, but let's have that discussion a bit later.

pymedphys/_dicom/anonymise.py
SimonBiggs: The trick is to just make the default be None, and then have a if replacement_strategy is None --> replacement_strategy = dict

Functionally it is exactly the same, it has just side stepped the gothya.

pymedphys/_dicom/anonymise.py
SimonBiggs: I think we can keep it simple by just going, if the user provides a replacement_strategy and there is a VR on the DICOM file that is not in the strategy, then raise an error. We can leave it at that. That's a clear contract.

Failure modes can be handled elsewhere if needed.

SimonBiggs: No stress about the list of keys. Potentially that is too hand holding.

pymedphys/_dicom/anonymisation_strategy.py
sjswerdloff: next push, as soon as I can see what is still failing

pymedphys/_dicom/anonymisation_strategy.py
sjswerdloff: yes.  but this approach will not likely be applicable to more complex strategies (e.g. pseudonymisation), or at least not fully applicable, and not without it's own twisting around to avoid vr specific logic.
it is applicable here.  I wasn't aware of partial functions in python before.  Learn something new every day :-)

pymedphys/_dicom/anonymise.py
```diff
@@ -662,5 +660,5 @@ def get_anonymous_replacement_value(
     if replacement_strategy is not None:
         replacement_value = replacement_strategy[vr](current_value)
     else:
-        replacement_value = anon.get_vr_anonymous_replacement_value_dict()[vr]
+        replacement_value = anon.ANONYMISATION_HARDCODE_DISPATCH[vr](current_value)
     return replacement_value
```
SimonBiggs, line 664: To remove the duplicated logic of `replacement_strategy[vr](current_value)` and `anon.ANONYMISATION_HARDCODE_DISPATCH[vr](current_value)` (look-up dict with `vr` and then call the function with `current_value`), can we change to the following:

```python
if replacement_strategy is not None:
    replacement_strategy = anon.ANONYMISATION_HARDCODE_DISPATCH


replacement_value = replacement_strategy[vr](current_value)
```

sjswerdloff: I get that in Python, “Object references are passed by value” (reassignment of the argument within the function doesn't affect the parameter that was passed in, while mutation of the contents of the argument within the function does affect the contents of the parameter passed in).  But this is incredibly similar in approach to having it as a default parameter.
The default dispatch dict is in theory mutable, but in practice it is not (or at least not intended to be mutated).
Mutation of anon.ANONYMISATION_HARDCODE_DISPATCH will effect subsequent runs, regardless of how it got mutated.  If we want to be 100% sure that it does not get mutated, things get complicated (there are frozen* and there is discussion of putting an immutable dict in Python 3.9).

pushed change (it's the right thing to do, given the current limitations in python not readily supporting immutable dictionaries)

pymedphys/_dicom/anonymise.py
```diff
@@ -1,4 +1,5 @@
 # Copyright (C) 2018 Matthew Jennings, Simon Biggs
+# Copyright (C) 2020 Stuart Swerdloff
```
SimonBiggs, line 2: Protocol here is to always have the latest additions higher up the document, even if they are small. If people want to see specific contributions they can look at the git blame :)

https://github.com/pymedphys/pymedphys/blame/894260669bc847de39baec9a45646735bc5cddb9/pymedphys/_dicom/anonymise.py

pymedphys/_dicom/anonymisation_strategy.py
```diff
@@ -56,97 +56,30 @@ def _get_vr_anonymous_hardcode_replacement_value(value_representation, current_v
...
+    "UI",
+    "US",
+]
+
+ANONYMISATION_HARDCODE_DISPATCH = {
+    key: functools.partial(
+        _get_vr_anonymous_hardcode_replacement_value, value_representation=key
+    )
+    for key in DISPATCH_KEYS_TO_OVERRIDE
+}
```
SimonBiggs, line 85: After your comments regarding having it not really worth detailing `DISPATCH_KEYS_TO_OVERRIDE` given that's just the VR, would it be better to do the following:

```python
def get_default_hardcode_dispatch():
    keys = get_vr_anonymous_replacement_value_dict().keys()

	ANONYMISATION_HARDCODE_DISPATCH = {
	    key: functools.partial(
	        _get_vr_anonymous_hardcode_replacement_value, value_representation=key
	    )
	    for key in DISPATCH_KEYS_TO_OVERRIDE
	}

	return ANONYMISATION_HARDCODE_DISPATCH
```

pymedphys/_dicom/anonymisation_strategy.py
SimonBiggs: > or at least not fully applicable, and not without it's own twisting around to avoid vr specific logic.

Yeah, certainly don't use it if it will require twisting.

> I wasn't aware of partial functions in python before. Learn something new every day :-)

Yeah, it certainly has a well defined place. It can be overused (like anything), but it can result in neat solutions. I am always amazed how much I learn through Open Source :).

pymedphys/_dicom/anonymise.py
SimonBiggs: > But this is incredibly similar in approach to having it as a default parameter.

On further thought, yeah, I don't think we've actually changed the problem. Because `anon.ANONYMISATION_HARDCODE_DISPATCH` is defined outside the function, if someone were to come along and mutate it, we would still have a bad time.

I thought the code I provided was actually logically different, but I do think at the end of the day it actually has all the same traps. The only time the pattern I described works is if the dictionary is initialised within the local scope of the function. Because we're pulling in a global constant from outside the scope of the function, we're actually falling into the very same trap.

The only thing saving us, is we're going to be responsible adults and not mutate that dictionary.

SimonBiggs: And there's me. I learnt something new :).

SimonBiggs:
Feel free to ignore the following fail:

https://github.com/pymedphys/pymedphys/pull/959/checks?check_run_id=930538482#step:34:96

I believe it is an upstream issue that I need to isolate and reproduce so that I can get it fixed.

pymedphys/_dicom/anonymisation_strategy.py
sjswerdloff: I've done half of this, without using keys as a temporary variable.
If we wrap the get_default_hardcode_dispatch with a lru_cache, then it avoids the dispatch table being built until it is first used.  If you want that done, I'm OK with it.  But if there is a problem, I'd like it to fail fast, and it will fail fast if it's a module level variable declaration (the first time someone imports the module... it will fail then if it's going to fail).

sjswerdloff: And now I've done the other half, include lru_cache.
I don't understand why it failed to find pydicom, but that was enough motivation for me to complete the approach you suggested and hope that runs the gauntlet

sjswerdloff: and used the temporary variable to avoid pylint whinging.

pymedphys/_dicom/anonymise.py
sjswerdloff: I think that by wrapping the dictionary generation with an lru_cache'd function with no parameters, it results in something that resembles a factory that produces the same dictionary always, so something like an immutable dictionary as long as people always use the function call to get to it.  I'd need to create a test to confirm that mutating the produced dictionary would in fact not affect the output of the lru_cache'd function.  An alternative is to *not* use lru_cache, but then the output dict would always be a new copy with a new reference.

we probably want to put up a sign that says "do not mutate" (or fold, or spindle) this dictionary.

pymedphys/_dicom/anonymisation_strategy.py
SimonBiggs: > I don't understand why it failed to find pydicom

Because by design, the PyMedPhys CLI and library needs to be able to start up and run even if none of its dependencies are installed. This is to support stripped down installations in weird and wonderful places such as [MicroPython](https://micropython.org/) and [PyOdide](https://hacks.mozilla.org/2019/04/pyodide-bringing-the-scientific-python-stack-to-the-browser/)

pymedphys/_dicom/anonymise.py
SimonBiggs: > I'd need to create a test to confirm that mutating the produced dictionary would in fact not affect the output of the lru_cache'd function.

Under the hood lru_cache creates a cache dictionary where it stores inputs as keys and outputs as values. If you mutate one of the outputs of an lru_cache it will mutate it everywhere.

In this case, I don't believe we need to take this any further. We have signaled that it is a constant with `UPPER_CASE`, I think the "We are all consenting adults" rules apply here:

https://python-guide-chinese.readthedocs.io/zh_CN/latest/writing/style.html#we-are-all-consenting-adults

SimonBiggs:
> switched to using temporary variable for list of VR keys to avoid pylint

I highly recommend [using a linter within your editing environment](https://code.visualstudio.com/docs/python/linting). Ideally CI should never need to tell you that pylint is grumpy.

pymedphys/_dicom/anonymisation_strategy.py
SimonBiggs: To give some more info here, before any dependencies are installed the following two commands a run to make sure that they at least run without error:

```bash
pymedphys --help
python -c "import pymedphys"
```

For more details, see:

https://github.com/pymedphys/pymedphys/blob/1086964c25f569f2171074f70f58076520c4f5b9/.github/workflows/on-push.yml#L119-L141

sjswerdloff:
I have (or at least had) multiple linters in my (VS) code environment, but I'll double check.
The complaint had to do with not iterating over the dictionary rather than getting the keys and then iterating over them.
What I think I need is to have the CI baked in to my own sandbox (maybe a k8 that gets provisioned depending on what project I'm working on and what CI it uses).  CPU is cheap.  Memory is cheap.  Bandwitdh is cheap.
Time from reviewers is too valuable to burn...

SimonBiggs:
I can also spin out the slow tests into their own one. If I'm reading this right the limit has now gone up to 20 concurrent jobs:

https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits

I think it used to be 8 ... I'll see if I can make that a priority.

PR #960, opened 2020-08-02
Disable pylint hint regarding dict iter

SimonBiggs:
@sjswerdloff, if there's something in pylint that we think shouldn't be flagged we can add it to the following list. What do you think about this change?

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -21,7 +21,8 @@ disable =
     ungrouped-imports,
     import-outside-toplevel,
     no-else-return,
-    unbalanced-tuple-unpacking
+    unbalanced-tuple-unpacking,
+    consider-iterating-dictionary
 
 
 [FORMAT]


sjswerdloff:
I think that something isn't working the way I think it should be working if I didn't get this warning in my own sandbox with the project settings having pylint enabled and the .pylintrc from pymedphys.
The GitHub Actions job found a pylint issue.  My environment should have as well.
so maybe I need to copy .pylintrc up to my home directory or enable pylint in my user settings (and not just the project settings) or otherwise find a way of having pylint operating while working on pymedphys

SimonBiggs:
> I think that something isn't working the way I think it should be working if I didn't get this warning in my own sandbox with the project settings having pylint enabled and the .pylintrc from pymedphys.

That is odd.

Yeah, not sure.

PR #961, opened 2020-08-02
Removed need for wrapping the strategy constants

SimonBiggs:
@sjswerdloff, the reason the strategy constants needed to be wrapped was because of the `"SQ": [pydicom.Dataset()]` line. Having to wrap them all up like that was getting to me.

Anyway, so I dived in and shuffled it a little. Let me know if you're okay with the changes.


pymedphys/_dicom/anonymise/strategy.py
```diff
@@ -0,0 +1,68 @@
...
+    for key in keys:
+        VR_TO_REPLACEMENT_MAP[key] = value
+
+
+def _get_vr_anonymous_hardcode_replacement_value(current_value, value_representation):
+    """A single dispatch function that is used for any VR with the current_value of the element ignored
+    This is the default for the replacement strategy
+    """
+    if current_value is None:
+        # TODO: Decide if we actually want to be doing this. I haven't
```
sjswerdloff, line 46: Personally, I don't like replacing an empty value with a hardcoded anonymisation value.  I don't think it makes the data any more anonymous and it risks interoperability issues or worse, DICOM conformance violation.
But one of the nice things about using the strategy (pattern) is substitution of an alternative strategy is straight forward.  So a different strategy is "hardcode but don't replace empty values", and empty could be either "no value" or "empty string".
I tend to avoid changing things unless I know they are wrong, instead I tend to offer an alternative.  One does end up with a lot of cruft after a while, but it avoids accidentally breaking things because someone had an expectation down the line that maybe they should not have.

pymedphys/_dicom/anonymise/strategy.py
```diff
@@ -0,0 +1,68 @@
...
+        # been the biggest fan of this outcome.
+        logging.debug("Replacing empty value with hardcoded anonymisation value")
+
+    if value_representation == "SQ":
+        # This is defined in here instead of within REPLACEMENT_TO_VR_MAP so that
+        # pydicom does not get called during library import. This is to support
```
sjswerdloff, line 52: The implication is that other strategies will need to (or should attempt to) take the same care to not have pydicom get called during library import.  It may get difficult to avoid calling on pydicom when the entire goal is to operate on DICOM data.  I guess the trick is to avoid having module level constants or variables that involve calling on pydicom (or any other particular import that one does not want to assume is in place for a user).

pymedphys/_dicom/anonymise/strategy.py
```diff
@@ -0,0 +1,68 @@
...
+        # installed.
+        return [pydicom.Dataset()]
+
+    return VR_TO_REPLACEMENT_MAP[value_representation]
+
+
+_keys = list(VR_TO_REPLACEMENT_MAP.keys())
+_keys.append("SQ")
+
+ANONYMISATION_HARDCODE_DISPATCH = {
```
sjswerdloff, line 63: need to wrap with lru cached function:
get_default_hardcode_dispatch()

pymedphys/_dicom/anonymise/strategy.py
SimonBiggs: > One does end up with a lot of cruft after a while, but it avoids accidentally breaking things because someone had an expectation down the line that maybe they should not have.

At the moment, given we are version 0.*, we are quite free to break expectations, especially to improve code quality and keep cruft out, if possible. In the future, we are still free to break expectations, it just has to come with deprecation warnings and the like.

I think for PyMedPhys to continue to stand under its own weight into the future, given the size of the development team, we simply cannot afford to let cruft hang around.

I really like the pragmatic programmer book, on page 27 onwards he describes "software entropy" and how some teams manage to successfully combat it.

https://www.cin.ufpe.br/~cavmj/104The%20Pragmatic%20Programmer,%20From%20Journeyman%20To%20Master%20-%20Andrew%20Hunt,%20David%20Thomas%20-%20Addison%20Wesley%20-%201999.pdf#page=27

Excerpt:

> While software development is immune from almost all physical laws, entropy hits us hard.
Entropy is a term from physics that refers to the amount of "disorder" in a system.
Unfortunately, the laws of thermodynamics guarantee that the entropy in the universe
tends toward a maximum. When disorder increases in software, programmers call it
"software rot."
>
> There are many factors that can contribute to software rot. The most important one seems
to be the psychology, or culture, at work on a project. Even if you are a team of one, your
project's psychology can be a very delicate thing. Despite the best laid plans and the best
people, a project can still experience ruin and decay during its lifetime. Yet there are other
projects that, despite enormous difficulties and constant setbacks, successfully fight
nature's tendency toward disorder and manage to come out pretty well.
>
> What makes the difference?


-----

> Personally, I don't like replacing an empty value with a hardcoded anonymisation value. I don't think it makes the data any more anonymous and it risks interoperability issues or worse, DICOM conformance violation.

I agree, it has stepped beyond its task, a function shouldn't go and do more than one expects it to. And it is confusing that they appear when they weren't there before.

I say we remove it and put a note in the changelog to warn people as such.

sjswerdloff:
@SimonBiggs does this now meet your current expectations (less Issue #963 )?
I'm investigating #963, but it's not a simple one line fix (in part due to automated tests, i.e. the implicit contract made explicit through testing).
If you are satisfied, feel free to merge.
I suggest we keep #963 separate from this.

pymedphys/_dicom/anonymise/strategy.py
```diff
@@ -60,9 +60,13 @@ def _get_vr_anonymous_hardcode_replacement_value(current_value, value_representa
...
 
-ANONYMISATION_HARDCODE_DISPATCH = {
-    key: functools.partial(
-        _get_vr_anonymous_hardcode_replacement_value, value_representation=key
-    )
-    for key in _keys
-}
+
+@functools.lru_cache(maxsize=1)
+def get_default_hardcode_dispatch():
```
SimonBiggs, line 65: I was actually hoping to remove the need for this wrapping entirely. I'll show you what I mean in the next commit.

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -691,7 +691,7 @@ def get_anonymous_replacement_value(
         #   elif ...
 
     if replacement_strategy is None:
-        replacement_strategy = strategy.get_default_hardcode_dispatch()
+        replacement_strategy = strategy.ANONYMISATION_HARDCODE_DISPATCH
```
SimonBiggs, line 694: Instead just by pulling it in here.

SimonBiggs:
@sjswerdloff I was actually hoping to not deal with lru_cache + function wrapping at all, given the need for that was removed by moving `[pydicom.Dataset()]` into the function.

I believe this is still logically equivalent though. Just without that extra chaff.

sjswerdloff:
> I believe this is still logically equivalent though. Just without that extra chaff.

It is now.
It was not before when I used it as a default argument value and it was flagged as a mutable object being used as a default argument value...
I believe it also implies that the dispatch table will be constructed before it is used, and I am now trying to make sure that the very bare minimum gets constructed before being used (to avoid forcing the need for external packages based on imports unless the functionality that requires it is being put in use).  I'm probably overdoing that for lack of deep understanding of what drives the package resolution.

sjswerdloff:
LGTM

PR #979, opened 2020-08-10
Allow alternative identifying keywords

sjswerdloff:
provide a mechanism for passing in an alternative list of identifying keywords.
The primary intended use is to allow passing in of a list that includes a variety of instance or instance reference UIDs so that said alternative list can be used for pseudonymisation.

### DIFF
--- a/pymedphys/_dicom/anonymise/__init__.py
+++ b/pymedphys/_dicom/anonymise/__init__.py
@@ -18,7 +18,7 @@ import json
 import logging
 import os.path
 import pprint
-from copy import deepcopy
+from copy import copy, deepcopy
 from glob import glob
 from os.path import abspath, basename, dirname, isdir, isfile
 from os.path import join as pjoin
@@ -42,10 +42,14 @@ IDENTIFYING_KEYWORDS_FILEPATH = pjoin(HERE, "identifying_keywords.json")
 
 
 @functools.lru_cache()
-def get_default_identifying_keywords():
+def _get_default_identifying_keywords():
     with open(IDENTIFYING_KEYWORDS_FILEPATH) as infile:
         IDENTIFYING_KEYWORDS = json.load(infile)
     return list(IDENTIFYING_KEYWORDS)
+
+
+def get_default_identifying_keywords():
+    return copy(_get_default_identifying_keywords())
 
 
 def label_dicom_filepath_as_anonymised(filepath):
--- a/pymedphys/_dicom/anonymise/__init__.py
+++ b/pymedphys/_dicom/anonymise/__init__.py
@@ -113,6 +113,12 @@ def anonymise_dataset(  # pylint: disable = inconsistent-return-statements
     copy_dataset : ``bool``, optional
         If ``True``, then a copy of ``ds`` is returned.
 
+    replacement_strategy: ``dict`` (keys are VR, value is dispatch function), optional
+        If left as the default value of ``None``, the hardcode replacement strategy is used.
+
+    alternative_identifying_keywords: ``list``, optional
+        If left as None, the default values for/list of identifying keywords are used
+
     Returns
     -------
     ds_anon : ``pydicom.dataset.Dataset``
@@ -243,6 +249,12 @@ def anonymise_file(
         set to ``False``, these tags are simply ignored. Pass ``False``
         with caution, since unrecognised tags may contain identifying
         information.
+
+    replacement_strategy: ``dict`` (keys are VR, value is dispatch function), optional
+        If left as the default value of ``None``, the hardcode replacement strategy is used.
+
+    alternative_identifying_keywords: ``list``, optional
+        If left as None, the default values for/list of identifying keywords are used
     """
     dicom_filepath = str(dicom_filepath)
 
@@ -348,6 +360,12 @@ def anonymise_directory(
         set to ``False``, these tags are simply ignored. Pass ``False``
         with caution, since unrecognised tags may contain identifying
         information.
+
+    replacement_strategy: ``dict`` (keys are VR, value is dispatch function), optional
+        If left as the default value of ``None``, the hardcode replacement strategy is used.
+
+    alternative_identifying_keywords: ``list``, optional
+        If left as None, the default values for/list of identifying keywords are used
     """
     dicom_dirpath = str(dicom_dirpath)
 
--- a/pymedphys/_dicom/anonymise/__init__.py
+++ b/pymedphys/_dicom/anonymise/__init__.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
+import functools
 import json
 import logging
 import os.path
@@ -40,8 +40,12 @@ HERE = dirname(abspath(__file__))
 
 IDENTIFYING_KEYWORDS_FILEPATH = pjoin(HERE, "identifying_keywords.json")
 
-with open(IDENTIFYING_KEYWORDS_FILEPATH) as infile:
-    IDENTIFYING_KEYWORDS = json.load(infile)
+
+@functools.lru_cache()
+def get_default_identifying_keywords():
+    with open(IDENTIFYING_KEYWORDS_FILEPATH) as infile:
+        IDENTIFYING_KEYWORDS = json.load(infile)
+    return list(IDENTIFYING_KEYWORDS)
 
 
 def label_dicom_filepath_as_anonymised(filepath):
@@ -64,7 +68,7 @@ def anonymise_dataset(  # pylint: disable = inconsistent-return-statements
     delete_unknown_tags=None,
     copy_dataset=True,
     replacement_strategy=None,
-    alternative_identifying_keywords=None,
+    identifying_keywords=None,
 ):
     r"""A simple tool to anonymise a DICOM dataset.
 
@@ -116,7 +120,7 @@ def anonymise_dataset(  # pylint: disable = inconsistent-return-statements
     replacement_strategy: ``dict`` (keys are VR, value is dispatch function), optional
         If left as the default value of ``None``, the hardcode replacement strategy is used.
 
-    alternative_identifying_keywords: ``list``, optional
+    identifying_keywords: ``list``, optional
         If left as None, the default values for/list of identifying keywords are used
 
     Returns
@@ -171,8 +175,7 @@ def anonymise_dataset(  # pylint: disable = inconsistent-return-statements
         ds_anon.remove_private_tags()
 
     keywords_to_anonymise = _filter_identifying_keywords(
-        keywords_to_leave_unchanged,
-        alternative_identifying_keywords=alternative_identifying_keywords,
+        keywords_to_leave_unchanged, identifying_keywords=identifying_keywords
     )
 
     ds_anon = _anonymise_tags(
@@ -196,7 +199,7 @@ def anonymise_file(
     delete_private_tags=True,
     delete_unknown_tags=None,
     replacement_strategy=None,
-    alternative_identifying_keywords=None,
+    identifying_keywords=None,
 ):
     r"""A simple tool to anonymise a DICOM file.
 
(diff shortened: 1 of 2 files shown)


pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -64,6 +64,7 @@ def anonymise_dataset(  # pylint: disable = inconsistent-return-statements
     delete_unknown_tags=None,
     copy_dataset=True,
     replacement_strategy=None,
+    alternative_identifying_keywords=None,
```
SimonBiggs, line 67: Can we just call this `identifying_keywords` throughout? And then, create an `lru_cached` function called `get_default_identifying_keywords` which then runs the following and returns the default keywords:

https://github.com/sjswerdloff/pymedphys/blob/cd15371571d29d88ad1db307c2ef8eb449311e77/pymedphys/_dicom/anonymise/__init__.py#L41-L44

We can then have `anonymised_dataset` be such that should it be `None`, then it sets `idetifying_keywords` to the default function.

What do you think?

pymedphys/tests/dicom/test_anonymise.py
```diff
@@ -169,6 +169,27 @@ def test_anonymised_dataset_with_empty_patient_sex():
     assert hardcode_replace_ds["PatientSex"].value is None
 
 
+@pytest.mark.slow
+@pytest.mark.pydicom
+def test_alternative_identifying_keywords():
+    default_keyword_list = list(IDENTIFYING_KEYWORDS)
+    alternative_keyword_list = list(default_keyword_list)
```
SimonBiggs, line 176: This does make a copy, but can we make it more explicit, as in, we're not trying to convert the variable to a list (what `list(object)` is for), but instead the point of this line is to make a copy:

```python
import copy

# ...

alternative_keyword_list = copy.copy(default_keyword_list)
```

pymedphys/tests/dicom/test_anonymise.py
```diff
@@ -169,6 +169,27 @@ def test_anonymised_dataset_with_empty_patient_sex():
     assert hardcode_replace_ds["PatientSex"].value is None
 
 
+@pytest.mark.slow
+@pytest.mark.pydicom
+def test_alternative_identifying_keywords():
+    default_keyword_list = list(IDENTIFYING_KEYWORDS)
```
SimonBiggs, line 175: Is this line needed in light of the following line and it's review comment?

pymedphys/tests/dicom/test_anonymise.py
```diff
@@ -169,6 +169,27 @@ def test_anonymised_dataset_with_empty_patient_sex():
...
+    alternative_keyword_list = list(default_keyword_list)
+    alternative_keyword_list.append("SOPInstanceUID")
+    test_file_path = get_treatmentrecord_test_file_path()
+    ds_test = pydicom.dcmread(test_file_path, force=True)
+    anon_private_filepath = anonymise_file(
+        test_file_path,
+        delete_private_tags=True,
+        alternative_identifying_keywords=alternative_keyword_list,
+    )
```
SimonBiggs, line 184: Sometimes these tests need to be able to be run in parallel. Therefore creating and deleting files amongst the test data can cause really weird errors down the line.

Can you make use of the tempfile library when creating files. Here's an example of what I mean:

https://github.com/pymedphys/pymedphys/blob/7959e19b256a3befb965817741002a3e6c6235d9/pymedphys/tests/trf/test_decode.py#L84-L89

When you use the tempfile library you won't actually have to delete the file at the end either.

pymedphys/tests/dicom/test_anonymise.py
```diff
@@ -169,6 +169,27 @@ def test_anonymised_dataset_with_empty_patient_sex():
...
+    ds_anon = pydicom.dcmread(anon_private_filepath, force=True)
+    try:
+        assert is_anonymised_file(anon_private_filepath, ignore_private_tags=False)
```
SimonBiggs, line 187: Is it worth having `is_anonymised_file` also accept `identifying_keywords`?

pymedphys/_dicom/anonymise/__init__.py
sjswerdloff: Will do.

pymedphys/tests/dicom/test_anonymise.py
sjswerdloff: OK.  But note that what I deleted was the newly created anonymised file whose SOP Instance UID was changed.

pymedphys/tests/dicom/test_anonymise.py
sjswerdloff: This implies considerable variability in what could be considered an anonymised file.
Which already has a (no offense) brittle definition.
So... yes, it's worthwhile, but I think it also needs to take in to account the strategy involved in anonymisation, which would then involve having an alternative (matching) strategy for determining if something was anonymised according to said particular strategy.
Or coming up with a very different definition of what "anonymised" really is.  For example by having the "non-anonymised" data in hand and then comparing to ensure that there is no identifying information.
Bottom line, I don't think it's quite the right time to refactor (redesign) is_anonymised_*().

pymedphys/tests/dicom/test_anonymise.py
sjswerdloff: Sure

pymedphys/tests/dicom/test_anonymise.py
sjswerdloff: No.  Once it's clear that one is making a copy to start with.
I took one approach, you have identified an alternative approach that expresses the semantics intended in a clearer fashion.  I'm happy to switch to something that reads more clearly.

pymedphys/tests/dicom/test_anonymise.py
SimonBiggs: Yup, the weird thing that can happen is, let's say this is being tested on the same machine for all of Python 3.6, 3.7, and 3.8. What can happen here is the Python 3.6 version might create the file, and then maybe the 3.7 version errors with a `FileExistsError`, or maybe it ignores that, uses the version that was created with Python 3.6, but then at the last moment while it's reading maybe Python 3.6 tries to delete it, and gets file lock/user access errors. All sorts of pain. And intermittent.

pymedphys/tests/dicom/test_anonymise.py
SimonBiggs: Yup, fair :). I'm also not opposed to its removal one day if it is too brittle.

Also, for any case within PyMedPhys, no offense needed, although we do pour ourselves and our ideas and take pride in work well done, at the end of the day it's a great thing to look at work once done, recognise it could have been done better, and change.

sjswerdloff: Well said.
The implementation for this (in PyMedPhys) is not brittle. 
What constitutes "anonymised" is very difficult to define.
There have been a variety of papers written (I participated in/co-authored one), and a DICOM supplement was created to help address the issues, so it's a tough nut to crack.
I think that without having a set of identifying information to measure against, determining if an arbitrary DICOM object contains identifying information is a formidable task (imagine if someone's name was actually "Anonymous" :-) 

I'm a big fan of continuous improvement.
I consider the perfect to be the enemy of the good.  Good, working code provides value.  Net Present Value.
Having achieved that, and learning along the way allows for improving code.
Also, the requirements for a body of code will almost certainly change over time (typically the requirements grow).
Trying to anticipate that too much and the results of taking that approach has spawned a multitude of software development (and general management) approaches.

So... yes, it's worthwhile to re-design is_anonymised_*().  But I'd like to review what the purpose is for that set of functions, and separate out the intended uses before doing further design, re-factoring, etc.
Even though I've just said trying to anticipate too much is not a good thing, spending a little time thinking about a problem before attacking it is worthwhile.
...
Potential purposes:
- One purpose is to ensure that a particular anonymisation strategy has been implemented correctly (including various variants/flags/semaphores).
- Another purpose might be to avoid applying a strategy/anonymising (at all), if the data is already "anonymised". 
- Another purpose might be to reject data that does not meet some kind of anonymisation criteria (e.g. ensuring that PHI has been protected before uploading/transmitting data outside the appropriate domain).
- Another (perhaps nefarious) purpose might be to rapidly filter through data to determine which might be susceptible to being de-identified (albeit with a more expensive/time consuming attack). Applying this in the form of a red team/blue team approach (e.g. a GAN) could be used for beneficial purposes to strengthen anonymisation techniques.

pymedphys/tests/dicom/test_anonymise.py
sjswerdloff: Understood and implemented.

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -667,16 +670,16 @@ def _anonymise_tags(
...
+    keywords_to_leave_unchanged, identifying_keywords=None
 ):
     r"""Removes DICOM keywords that the user desires to leave unchanged
     from the list of known DICOM identifying keywords and returns the
     resulting keyword list.
     """
-    if alternative_identifying_keywords is None:
-        keywords_filtered = list(IDENTIFYING_KEYWORDS)
+    if identifying_keywords is None:
+        keywords_filtered = get_default_identifying_keywords()
```
SimonBiggs, line 680: I believe a `copy` might be needed here.

SimonBiggs: See the following:

```python
In [1]: import functools

In [2]: @functools.lru_cache()
   ...: def a():
   ...:     return [1, 2, 3]
   ...:

In [3]: b = a()

In [4]: b
Out[4]: [1, 2, 3]

In [5]: b.append('oops')

In [6]: a()
Out[6]: [1, 2, 3, 'oops']
```

Essentially, running "keywords_filtered" once, could permanently remove them from the identifying list for that run through.

sjswerdloff: I agree that the list needs to be safeguarded against mutation.
I've committed an alternative approach to achieving that, one that I think is a reasonable trade-off in performance/memory utilisation and minimising the need for the consumer to be "careful".

SimonBiggs: Yup, awesome. I like that approach much better.

pymedphys/tests/dicom/test_anonymise.py
SimonBiggs: Yup, let's spin this off into a separate issue.

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -40,8 +40,16 @@
...
 IDENTIFYING_KEYWORDS_FILEPATH = pjoin(HERE, "identifying_keywords.json")
 
-with open(IDENTIFYING_KEYWORDS_FILEPATH) as infile:
-    IDENTIFYING_KEYWORDS = json.load(infile)
+
+@functools.lru_cache()
+def _get_default_identifying_keywords():
+    with open(IDENTIFYING_KEYWORDS_FILEPATH) as infile:
+        IDENTIFYING_KEYWORDS = json.load(infile)
+    return list(IDENTIFYING_KEYWORDS)
```
SimonBiggs, line 48: Or, would this be better as:

```suggestion
    return tuple(IDENTIFYING_KEYWORDS)
```

That way mutation simply wouldn't be possible?

sjswerdloff: but then we are back to the creation of the list having an implicit copy instead of explicit.
And the same issue is true for the (alternative) identifying keywords being passed in, so I needed to copy that list.

sjswerdloff: Ah...  I see the point.  Even within the module, one would not be able to mutate the iterable.
And instead of copying a list in the "public" method, one must create a list from the tuple.
That works for me.

SimonBiggs:
Yup :)

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -45,11 +45,11 @@
 def _get_default_identifying_keywords():
     with open(IDENTIFYING_KEYWORDS_FILEPATH) as infile:
         IDENTIFYING_KEYWORDS = json.load(infile)
-    return list(IDENTIFYING_KEYWORDS)
+    return tuple(IDENTIFYING_KEYWORDS)
 
 
 def get_default_identifying_keywords():
```
SimonBiggs, line 51: Is this wrapper even needed now?

PR #983, opened 2020-08-11
Matthew

matthewdeancooper:
Hi Simon, 
Probably easier to move our discussions to here instead of sms.

I have set the maximum number of simultaneous association requests to be 1 on the scp. Now, Monaco throws an export error straight away and will only allow exporting once the previous association request has been released. A fix that I imagine would work for each series being sent automatically (because of the time delay between imaging), however, it would be nice if we could find a way to queue manual requests.

EDIT: Just saw that you can comment on commits without first requiring a PR.


SimonBiggs:
I just wanted to add a note, than when DICOM sending to Monaco, you may
have to also send all the CTs. We'll see.

SimonBiggs:
Is pre-commit running on your machine? (https://github.com/pymedphys/pymedphys/pull/983/checks?check_run_id=970032404#step:7:48)

SimonBiggs:
A great next step to getting those tested would be to make a pymedphys CLI for them. And then call the CLI within the test.

matthewdeancooper:
@SimonBiggs  Something like this may provide a solution to exporting multiple series 

https://stackoverflow.com/questions/57818164/add-item-to-asyncio-queue-from-a-request-handler

It seems like the current code is working on multiple series using either the threading or multiprocess modules. Multi-processing may be the safest option between the two as each process is run on a different CPU core with its own interpreter. Modern CPUs have something like 6-12 cores as far as i know. Although, without some sort of queue, this is just pushing the limitation back a step and would not apply to GPU inference. A class as shown in the link above would allow the queue loop to be run independently of of the release handle.

Picking your brain here...

Cheers

SimonBiggs:
At the end of the day there is an upper limit of what you can accept, so at some point it will saturate. I would recommend going down the que route and just handling one request at a time for now. This makes a speed/robustness/development effort trade off. It runs slower, but it can handle by default any number of requests, as well as support both CPU and GPU configs out of the box. I believe at the end of the day some sort of que will be needed even if one day a multithreaded approach is undergone.

Also, tensorflow may already be using multiple CPU threads to achieve its inference. So this approach may not actually be any faster.

I highly recommend implementing a que, assuming that people will send multiple in a row, and do a FIFO approach.

SimonBiggs:
When there is a que, it can be really helpful for the user to have some form of status indicator. It might be helpful to have a little mini web server that details the current status of the que. That way the user can see where the processing is up to. I can help you do that part.

matthewdeancooper:
Having a read now.  :) 

Cheers

matthewdeancooper:
Ok so I think i have a prototype running correctly on both GPU and CPU.

I needed to change the matplotlib backend to Agg, as the default (TKinter) is not thread safe.

Additionally, I made use of multi-threading so that we could have the inference loop checking to see if the queue (via deque) was empty simultaneously while each association release adds its study path to the queue.  Ill commit after a little more testing :)

SimonBiggs:
By the way, when I said a "que", I was hoping that you would do it without
using multithreading at all. Before making a commit that includes
multithreading (or multiprocessing) maybe it's worth watching the following
PyCon keynote by an awesome Python programmer:

https://www.youtube.com/watch?v=9zinZmE3Ogk

... It certainly is possible without using multithreading and just a que
alone...

SimonBiggs:
And, this joke within the video is semi-serious:

[image: image.png]

:)...

SimonBiggs:
And, this joke within the video is semi-serious :):



-- https://pybay.com/site_media/slides/raymond2017-keynote/threading.html#david-baron-at-mozillas-san-francisco-office

matthewdeancooper:
Haha! Not even with stilts on!

PR #984, opened 2020-08-12
creating an extend-ct server prototype

SimonBiggs:


SimonBiggs:
@matthewdeancooper you might be interested in this. Here I use the same AE server to receive some DICOM files, do some processing on them, and then send the files back to the TPS.

https://github.com/pymedphys/pymedphys/blob/extend-ct-server-prototype/protyping/dicom/extend-ct/extend-ct-server.ipynb

PR #985, opened 2020-08-12
Experimental pseudonymisation strategy

sjswerdloff:
Puts strategy and additional identifying keywords (uids) in to experimental area, includes CLI


sjswerdloff:
Addresses Issue 828:  Provide an option for pseudonymisation

pymedphys/_experimental/pseudonymisation/strategy.py
```diff
@@ -0,0 +1,379 @@
...
+    -------
+    None.
+
+    """
+    vr = get_baseline_keyword_vr_dict()[keyword]
+    replacement_value = value
+    if vr in pseudonymisation_dispatch:
+        replacement_value = pseudonymisation_dispatch[vr](value)
+    else:
+        logging.warning("VR of %s not found in pseudonymisation dispatch table", vr)
```
SimonBiggs, line 55: In the principle of "Design with Contract" I feel like this here should raise an error. If the function that calls it is okay with it raising an error and it wants to ignore that, then the calling function can put it within a `try/except`.

>  Here, the emphasis is on "lazy" code: be strict in what you will accept before you begin, and promise as little as possible in return. Remember, if your contract indicates that you'll accept anything and promise the world in return, then you've got a lot of code to write!

-- Pragmatic Programmer

pymedphys/_experimental/pseudonymisation/strategy.py
```diff
@@ -0,0 +1,379 @@
...
+    -------
+    The base64 text representation of the hashed input value
+
+    """
+    try:
+        encoded_value = value.encode("ASCII")
+    except AttributeError:
+        encoded_value = str(value).encode("ASCII")
+
+    my_hash_func = hashlib.new("sha3_256")
```
SimonBiggs, line 81: Was the intent for the user to be able to revert these values if they had a private key? I thought the approach might involve a reversible encryption step.

pymedphys/_experimental/pseudonymisation/strategy.py
```diff
@@ -0,0 +1,379 @@
...
+            pseudo_numeric = 12 * numeric + increment
+            pseudo_unit = "M"
+        elif unit == "M":
+            pseudo_numeric = 4 * numeric + increment
+            pseudo_unit = "W"
+        elif unit == "W":
+            pseudo_numeric = 7 * numeric + increment
+            pseudo_unit = "D"
+
+    pseudo_age = str(pseudo_numeric).zfill(3) + pseudo_unit
```
SimonBiggs, line 150: Do we want this to be reversible?

Something that would be reversible, but also unable to be determined without a key (the large prime number) would be modulo arithmetic with a modulus that is prime and multiplying by a very large number that is also prime, and then offset so as to always be older than 20 say.

pymedphys/tests/experimental/pseudonymisation/test_pseudonymisation.py
```diff
@@ -0,0 +1,130 @@
...
+        assert ds_input["PatientID"].value != ds_pseudo["PatientID"].value
+        # make sure that we are not accidentally using the hardcode replacement approach
+        assert ds_pseudo["PatientID"].value not in ["", "Anonymous"]
+
+
+@pytest.mark.slow
+@pytest.mark.pydicom
+@pytest.mark.skipif(
+    "SUBPACKAGE" in os.environ, reason="Need to extract CLI out of subpackages"
+)
```
SimonBiggs, line 69: Yeah, I'm realising that what is making this hard to review, it is hard to know which bits have been changed. Copying the file into experimental and making changes, although it's okay, given the fact it's in experimental, it is hard to review.

pymedphys/_experimental/pseudonymisation/strategy.py
SimonBiggs: eg https://cryptography.io/en/latest/fernet/

pymedphys/_experimental/pseudonymisation/strategy.py
sjswerdloff: No.  As above, if someone needs the PHI, they should already have it with the original data.
If they need to find their way back to that original data see previous comment/response.

pymedphys/tests/experimental/pseudonymisation/test_pseudonymisation.py
sjswerdloff: The only bits that were really copied was the invocation of the methods that are in the "stable" (non-experimental) package.
In essence, the strategy and keywords are what has changed (they are new).

sjswerdloff:
An encryption strategy is a legitimate strategy.
The strategy I have defined is a hashing (one-way) strategy.
They have different purposes and different trade-offs.
Encryption requires a key.  Which means people have to *have* a key, and they have to have access to that key later on (to fulfil the goal of reversing the encryption).  That's a burden to anyone who is simply interested in shipping their data off elsewhere so it can be used (for good purposes) as part of some pool of data (data lake?) and once it's out of their hands (and they are reasonably confident that the PHI won't be accessible), they just don't care.

Now that the anonymisation package has been refactored, a developer can implement an encryption strategy and deploy it programmatically without having to fork/modify/submit to pymedphys, or they can readily submit it as an addition to the experimental area (either in pseudonymisation or somewhere else, but the task would probably be easier in pseudonymisation).

There are a variety of improvements that I can envision for the hashing strategy, including providing salt (preferably stable for a given user or facility so they don't make extra copies that they think are distinct from the same original data), and an alternative ORG_ROOT (as options within the current implementation).

If someone wants to implement an encryption strategy, I will be more than happy to act as a/the reviewer.
If the use case is compelling, I would be willing to put implementation of an encryption strategy in to my "To Do" list.
But I'm not an expert in encryption, key storage, etc.  And from my perspective, it's the infrastructure around the key management that's the hard part.

pymedphys/_experimental/pseudonymisation/__init__.py
```diff
@@ -1,8 +1,108 @@
...
+    if not args.keywords_to_leave_unchanged:
+        keywords_to_leave_unchanged = ()
+    else:
+        keywords_to_leave_unchanged = args.keywords_to_leave_unchanged
+
+    replacement_strategy = None
     if args.pseudo:
-        print("Was run with pseudo!")
+        logging.info("Was run with pseudo!")
+        identifying_keywords_for_pseudo = get_default_pseudonymisation_keywords()
```
sjswerdloff, line 72: new list of identifying keywords, which is basically the uids plus the identifying keywords already known to the anonymise module (the union of them, to avoid a problem if someone adds the uids back to the anonymise module).

SimonBiggs:
Yup, fair point. To ease the "storage" of the encryption the "key" can be a password that is then turned into an encryption key.

I think I had presumed that the encryption route was the one chose after the following comment:

https://github.com/pymedphys/pymedphys/issues/828#issuecomment-624345747

But that's all good :). Happy to have this in pymedphys :). It will be good to move it out of experimental once it's gone through some paces.

SimonBiggs:
@sjswerdloff did you want to have a celebratory beer one night?

sjswerdloff:
Definitely. 
I'll tidy up the bar.

SimonBiggs:
:)

SimonBiggs:
I would also argue it's overdue having your name on the contributors list on the front page. It would also be great if you can detail the changes within the changelog:

https://github.com/pymedphys/pymedphys/blob/master/CHANGELOG.md

SimonBiggs:
Likely also worth add a document to the CLI documentation:

https://github.com/pymedphys/pymedphys/tree/master/docs/ref/cli

SimonBiggs:
And exposing it via the API:

https://github.com/pymedphys/pymedphys/tree/master/pymedphys/experimental

and then adding appropriate documentation for that exposed API:

https://github.com/pymedphys/pymedphys/tree/master/docs/ref/lib/experimental

PR #987, opened 2020-08-12
Raise key error pseudonymisation

sjswerdloff:
addresses review comment regarding handling situation where the identifier has a vr that is not part of the strategy

### DIFF
--- a/pymedphys/_experimental/pseudonymisation/strategy.py
+++ b/pymedphys/_experimental/pseudonymisation/strategy.py
@@ -51,13 +51,7 @@ def get_pseudonymous_replacement_value(keyword, value):
 
     """
     vr = get_baseline_keyword_vr_dict()[keyword]
-    if vr in pseudonymisation_dispatch:
-        replacement_value = pseudonymisation_dispatch[vr](value)
-    else:
-        logging.error(
-            "VR of %s for %s not found in pseudonymisation dispatch table", vr, keyword
-        )
-        raise KeyError
+    replacement_value = pseudonymisation_dispatch[vr](value)
 
     return replacement_value
 
--- a/pymedphys/_experimental/pseudonymisation/strategy.py
+++ b/pymedphys/_experimental/pseudonymisation/strategy.py
@@ -15,7 +15,6 @@
 import base64
 import datetime
 import hashlib
-import logging
 import random
 from decimal import Decimal, DecimalTuple
 
--- a/pymedphys/_experimental/pseudonymisation/strategy.py
+++ b/pymedphys/_experimental/pseudonymisation/strategy.py
@@ -51,7 +51,6 @@ def get_pseudonymous_replacement_value(keyword, value):
 
     """
     vr = get_baseline_keyword_vr_dict()[keyword]
-    replacement_value = value
     if vr in pseudonymisation_dispatch:
         replacement_value = pseudonymisation_dispatch[vr](value)
     else:
--- a/pymedphys/_experimental/pseudonymisation/strategy.py
+++ b/pymedphys/_experimental/pseudonymisation/strategy.py
@@ -32,19 +32,22 @@ DICOM_DATETIME_FORMAT_STR = "%Y%m%d%H%M%S.%f"
 
 def get_pseudonymous_replacement_value(keyword, value):
     """
-    Get an appropriate value for a DICOM element based on its VR and
-    take in to account if it is an SOP Class UID.
-    Cache values of keywords that have already come in and re-use
-    the pseudonym for the *value* if it has already been used/replaced.
+    Get an appropriate value for a DICOM element based on its VR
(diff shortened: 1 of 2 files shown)


sjswerdloff:
??? This PR has no commits to streamlit, nor anonymise itself, only experimental/pseudonymisation.

=================================== FAILURES ===================================
____________ [pylint] pymedphys/_gui/streamlit/anonymise-monaco.py _____________
E: 35,10: Module 'streamlit' has no 'ScriptRunner' member (no-member)
E: 35,41: Module 'streamlit' has no 'ScriptRequestQueue' member (no-member)
E: 74,10: Module 'streamlit' has no 'ScriptRunner' member (no-member)
________________ [pylint] pymedphys/_gui/streamlit/electrons.py ________________
E: 24,10: Module 'streamlit' has no 'ScriptRunner' member (no-member)
___________________ [pylint] pymedphys/_streamlit/monaco.py ____________________
E: 32,14: Module 'streamlit' has no 'ScriptRunner' member (no-member)
E: 42,18: Module 'streamlit' has no 'ScriptRunner' member (no-member)

sjswerdloff:
While my (vs) code still doesn't seem to be catching pylint issues, I have discovered:
poetry run pymedphys --debug dev tests --pylint --pylint-jobs=1

so with a modest amount of discipline on my part, I hope to avoid getting linter errors in a pull request

SimonBiggs:
> While my (vs) code still doesn't seem to be catching pylint issues, I have discovered:
> poetry run pymedphys --debug dev tests --pylint --pylint-jobs=1
> 
> so with a modest amount of discipline on my part, I hope to avoid getting linter errors in future pull requests

All good :), it's only an issue with the `with_updates` CI:

image
image

If with updates fails, but the standard tests CI passes, then that is nothing to do with your code. It's an upstream issue. And therefore can be ignored for the purpose of the PR. It just means I need to sort something out before updating the dependencies.

pymedphys/_experimental/pseudonymisation/strategy.py
```diff
@@ -32,27 +32,32 @@
...
     vr = get_baseline_keyword_vr_dict()[keyword]
-    replacement_value = value
     if vr in pseudonymisation_dispatch:
         replacement_value = pseudonymisation_dispatch[vr](value)
     else:
-        logging.warning("VR of %s not found in pseudonymisation dispatch table", vr)
+        logging.error(
+            "VR of %s for %s not found in pseudonymisation dispatch table", vr, keyword
+        )
+        raise KeyError
```
SimonBiggs, line 60: So that the traceback is a more direct link can we do EAFP instead of LBYL:

```python
try:
	replacement_value = pseudonymisation_dispatch[vr](value)
except KeyError:
    logging.error(
        "VR of %s for %s not found in pseudonymisation dispatch table", vr, keyword
    )
    raise
```

SimonBiggs: Or better:

```python
try:
	dispatch = pseudonymisation_dispatch[vr]
except KeyError:
    logging.error(
        "VR of %s for %s not found in pseudonymisation dispatch table", vr, keyword
    )
    raise

replacement_value = dispatch(value)
```

That way, a KeyError in the function call won't be incorrectly caught.

SimonBiggs: But even better still, if you really do want it to be a warning, you can wrap `get_pseudonymous_replacement_value(keyword, value)` in a `try/except`, and then do a logging.warning there. As such, the logging.error potentially doesn't belong in this function, if you intend to turn it into a warning.

sjswerdloff: ok.  no try/catch.  if there's a problem, it's entirely up to the calling code to deal with it, and if not, the stack trace/dump should have the values not only of the vr that the dict didn't support, but also which keyword was the one responsible for it.  I'm used to the paradigm of leaving breadcrumbs, but if EAFP is the preferred approach in python, I'm not going to argue.
This motivates creation of a validation/helper function so that a programmer can test that their combination of keywords and strategy are compatible up front.  Especially if the keywords become user (rather than programmer) defined.

SimonBiggs: > it's entirely up to the calling code to deal with it

Yup, beautiful :) I believe that's the heart of the following:

> ... the emphasis is on "lazy" code: be strict in what you will accept before you begin, and promise as little as possible in return. Remember, if your contract indicates that you'll accept anything and promise the world in return, then you've got a lot of code to write!

> This motivates creation of a validation/helper function so that a programmer can test that their combination of keywords and strategy are compatible up front. Especially if the keywords become user (rather than programmer) defined.

I really like that concept.

sjswerdloff:
failing test is streamlit, not related to this PR.
Need this PR merged so I can expose the API per comment in #985

PR #995, opened 2020-08-14
Fixes #973 issue with pinnacle export CLI

### LINKED ISSUE #973: @pchlap Pinnacle experimental tools line command
Hi Phillip,
Thank you for the Pinnacle tar to DICOM tool. It is very nice.

However, being a python and pymedphys novice, I have a few problems when I try running it in command line.
I was able to use the -m option and export only dose, plans and structures, but I have not been able to pause the script, read the available plans, select the one I need and export a specific trial within it.

Is there a way I can do this?

Thanks in advance

Sotiri

pchlap:
This fixes the issue reported in #973 where the trial couldn't be set using the command line interface.

@SimonBiggs I don't have a test which covers this code, since it's part of the CLI. Would you expect to have a test for the CLI functions?

### DIFF
--- a/pymedphys/_experimental/pinnacle/pinnacle.py
+++ b/pymedphys/_experimental/pinnacle/pinnacle.py
@@ -272,12 +272,19 @@ class PinnacleExport:
                 # If no ImageHeader is available then we can ignore this
                 continue
 
-            self.logger.info(
-                "%s: %s %s",
-                image_header["modality"],
-                image_header["series_UID"],
-                image_header["SeriesDateTime"],
-            )
+            modality = "?"
+            if "modality" in image_header.keys():
+                modality = image_header["modality"]
+
+            series_uid = "?"
+            if "series_UID" in image_header.keys():
+                series_uid = image_header["series_UID"]
+
+            series_date_time = "?"
+            if "SeriesDateTime" in image_header.keys():
+                series_date_time = image_header["SeriesDateTime"]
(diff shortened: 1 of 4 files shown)


SimonBiggs:
> Would you expect to have a test for the CLI functions?

Yup, certainly, if that's okay.

Here is an example:

https://github.com/pymedphys/pymedphys/blob/ed39ad7850c57227cf9f2b15e31b18cfcf170660/pymedphys/tests/dicom/test_header_tweaks.py#L44-L54

Although there I removed the file after, it is much better to use a temp directory like the following:

https://github.com/pymedphys/pymedphys/blob/ed39ad7850c57227cf9f2b15e31b18cfcf170660/pymedphys/tests/experimental/pseudonymisation/test_pseudonymisation.py#L92-L114

SimonBiggs:
You can feel free to ignore the following test failure:

image

Unfortunately it is intermittent... I need to get to the bottom of that :(.

pchlap:
> > Would you expect to have a test for the CLI functions?
> 
> Yup, certainly, if that's okay.

Cool, thanks for the examples. I'll sort that out, hopefully I can get that done tomorrow morning :)

pchlap:
Hi @SimonBiggs ,

I've added some tests for the CLI, let me know what you think. If you're happy we can merge this and let it roll out in the next release. Thanks a lot!

PR #997, opened 2020-08-14
Document pseudonymisation api and tweak pyproject.toml to nudge streamlit pylint issue

sjswerdloff:
I'm not able to get the pseudonymisation_dispatch member to appear, so this isn't good to go.
But, it's a starting point for the documentation and it's a nudge on the streamlit pylint issue.

### DIFF
--- a/docs/ref/lib/experimental/index.rst
+++ b/docs/ref/lib/experimental/index.rst
@@ -17,3 +17,4 @@ Experimental Modules
     pinnacle
     profile
     sinogram
+    pseudonymisation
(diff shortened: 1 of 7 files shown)


sjswerdloff:
Interesting failures.
@SimonBiggs it doesn't seem that the change to pyproject.toml had an impact on _streamlit or streamlit.ScriptRunner related code.

SimonBiggs:
Instead of bumping streamlit up, instead I am holding it back in the following:

https://github.com/pymedphys/pymedphys/pull/998

Which should hopefully fix all the issues. I can then choose to explicitly upgrade it instead.

SimonBiggs:
So, streamlit was fixed by the above... but that weird cffi bug also crept in. Fixed that in the following:

https://github.com/pymedphys/pymedphys/pull/1000

sjswerdloff:
discard in favor of alternative PRs

PR #998, opened 2020-08-14
lock streamlit to 0.64.0

SimonBiggs:
Here I pinned streamlit to a historical version that worked. For the time being I think I will explicitly pin streamlit, and choose to upgrade it only purposefully.


pyproject.toml
```diff
@@ -67,26 +67,20 @@ pymssql = { version = "< 3.0.0", python = "< 3.8" }
 python_dateutil = "*"
 scikit-image = "*"
 
+# gui
+streamlit = "0.64.0"
```
SimonBiggs, line 71: Here is the primary change

PR #1000, opened 2020-08-14
run poetry update

SimonBiggs:


pyproject.toml
```diff
@@ -59,6 +59,9 @@ pydicom = ">=2.0.0"
 pynetdicom = "*"
 pylibjpeg-libjpeg = "*"
 
+# pins for flow on deps
+cffi = "1.14.0"
```
SimonBiggs, line 63: Here's the primary fix for this one... weird, the issue was on a 1.14.0 --> 1.14.1. I'll raise a case upstream.

PR #1002, opened 2020-08-14
Show effect of CFFI upgrade

SimonBiggs:
When upgrading from cffi 1.14.0 --> 1.14.1 it appears a new cffi dependency has been added to the code, but that dependency does not appear to have been declared within the pypi manifest. The extra dependency appears to be `pycparser`.

Here is a PR where the only change to the code base is upgrading from cffi 1.14.0 --> 1.14.1


SimonBiggs:
...ooo weird... it didn't break :/

PR #1009, opened 2020-08-17
document pseudonymisation module api, use autoattribute to document the module variable

sjswerdloff:
used annotate keyword to provide descriptive text instead of attempting
to use docstring or comment doc in the source for the module variable/attribute

### DIFF
--- a/docs/ref/lib/experimental/pseudonymisation.rst
+++ b/docs/ref/lib/experimental/pseudonymisation.rst
@@ -1,0 +1,43 @@
+#####################
+Pseudonymisation Tool
+#####################
+
+.. contents::
+    :local:
+    :backlinks: entry
+
+
+.. automodule:: pymedphys.experimental.pseudonymisation
+    :no-members:
+
+*******
+Summary
+*******
+
+    Pseudonymisation provides a list of identifying keywords that includes UIDs
+    and an anonymisation strategy that utilises SHA3_256 to hash text which is then
+    encoded as base64, or for UIDs, converted to an integer appended to the PyMedPhys Org Root.  Dates are shifted consistently and ages are jittered.
+    The list of identifying keywords and the anonymisation strategy are passed in to
+    the (non-experimental) anonymisation module/functions.
+
+
+
+***
+API
+***
+
+.. autofunction:: pymedphys.experimental.pseudonymisation.get_default_pseudonymisation_keywords
+.. autoattribute:: pymedphys.experimental.pseudonymisation.pseudonymisation_dispatch
+    :annotation: strategy, i.e. dictionary of VR and function references for anonymisation to achieve pseudonymisation
+
+*******
+Example
+*******
+::
+
+    import pymedphys.experimental.pseudonymisation as pseudonymisation_api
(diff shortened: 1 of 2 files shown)


docs/ref/lib/experimental/pseudonymisation.rst
```diff
@@ -0,0 +1,43 @@
...
+.. autofunction:: pymedphys.experimental.pseudonymisation.get_default_pseudonymisation_keywords
+.. autoattribute:: pymedphys.experimental.pseudonymisation.pseudonymisation_dispatch
+    :annotation: strategy, i.e. dictionary of VR and function references for anonymisation to achieve pseudonymisation
+
+*******
+Example
+*******
+::
+
+    import pymedphys.experimental.pseudonymisation as pseudonymisation_api
```
SimonBiggs, line 38: I've been thinking about this API here. I wonder if we should eventually have the API be:

```python
from pymedphys.dicom import strategy

result = anonymise_dataset(dataset, strategy=strategy.pseudonymisation)
```

To achieve this we can combine the dispatch and the identifying keywords into the same dictionary...

I'm happy to merge this, but I am keen to narrow down on a nice clean API.

SimonBiggs: I'm not sold on what I just presented there either. But I am really keen to make a really nice clean API

sjswerdloff: if you mean to condense the various "knobs and buttons" that are currently used to control what happens within the anonymisation, then yes, having a single object represent the set of values currently defined separately:
replace_values=True, keywords_to_leave_unchanged=(), delete_private_tags=True, delete_unknown_tags=None, copy_dataset=True, replacement_strategy=None, identifying_keywords=None
While it could be done as a dictionary, it might make sense to incorporate it in to a Class, so the class can be derived from, and in that class, have the "keys validate against the dispatch table" (potentially automatically at creation and when any particular control is modified, as well as explicit if a programmer wants to confirm the strategy is valid before sending it in)
This also helps to avoid people mutating the dictionary unintentionally (although just having a function that returns a copy rather than directly hitting at the pseudonymisation_dispatch would be appropriate and is simple).
But it won't reduce the work a programmer has to do.  It will simply shift it from the anonymisation call to the specification of the contents of the strategy.  The benefit is in making a clear separation of where the concern is addressed.
It might motivate creation of either a variety of child classes or a set of factory creation methods, but it will at least shift the signature from invocation of anonymisation over to creation of a strategy object.

SimonBiggs: Yup, I think an "anonymisation strategy"/config object which has validation methods on initialisation would be quite a tidy way to go.

Building it as a frozen dataclass under the hood might be a neat solution:
https://docs.python.org/3/library/dataclasses.html#frozen-instances

SimonBiggs: Let's use https://github.com/pymedphys/pymedphys/issues/1011 to track this.

PR #1010, opened 2020-08-17
add stuart to the list of contributors

sjswerdloff:


sjswerdloff:
> Awesome LGTM. Would you be happy with me making your name a bit more prominent when I get the chance?

If you like.  I've basically just scratched my own itch at this point.
But I do think that once the pseudonymisation API documentation is in and there is a minor release (0.x++), I'll be quite ready for that celebratory beer :-)

SimonBiggs:
> and there is a minor release (0.x++)

So that you can begin playing with it as it currently is, I have just cut a development release, which can be installed on machines using:

```
pip install pymedphys==0.32.0dev5
```

Also, for a minimal "anonymisation only" install, you can use:

```
pip install pymedphys==0.32.0dev5 --no-deps
pip install pydicom
```

PR #1029, opened 2020-08-28
Beginning of autocontouring project documentation

SimonBiggs:
@matthewdeancooper can you clone this branch and run:

```bash
# install docs dependencies
poetry install -E docs

# build docs
poetry run pymedphys dev docs
```

Which will build the docs, which you can then browse by using something like https://docs.python.org/3/library/http.server.html.

Then you can see what the impact is of updating this file. If you can then update the document at https://github.com/pymedphys/pymedphys/blob/autocontouring-project/docs/background/autocontouring.rst we might be able to get it merged today. ...I'd rather not need to work on this on Saturday if possible :).

Cheers,
Simon

### DIFF
--- a/docs/background/autocontouring.rst
+++ b/docs/background/autocontouring.rst
@@ -20,7 +20,7 @@ Video goes here
 
 Details
 -------
-George!
+George! [1]_
 
 
 
--- a/docs/background/autocontouring.rst
+++ b/docs/background/autocontouring.rst
@@ -1,0 +1,42 @@
+======================
+Autocontouring Project
+======================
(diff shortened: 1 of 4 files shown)


matthewdeancooper:
@SimonBiggs  Ready for review.

SimonBiggs:
Beautiful :)

SimonBiggs:
@matthewdeancooper we need to make that video embedding responsive (as in, happily work on a phone screen).

Here's where you can get the code to achieve that:

https://embedresponsively.com/

PR #1031, opened 2020-09-03
Auto segmentation lock bug

matthewdeancooper:
How does this look to you?  Moved some code in handle_accepted(event) so that is requires a lock. However, I could not reproduce the bug on my end.

### DIFF
--- a/exit
+++ b/exit
@@ -1,3 +1,0 @@
-  autocontouring-project[m
-* [32mmaster[m
-  matthew[m
--- a/exit
+++ b/exit
@@ -1,0 +1,3 @@
+  autocontouring-project[m
+* [32mmaster[m
+  matthew[m
--- a/protyping/auto-segmentation/storage_and_inference_scp.py
+++ b/protyping/auto-segmentation/storage_and_inference_scp.py
@@ -39,20 +39,20 @@ def handle_accepted(event):
 def handle_store(event):
     """Handle EVT_C_STORE events."""
 
-    ds = event.dataset
-    ds.file_meta = event.file_meta
-
-    # Path for the study
-    study_path = config.SCP_STORAGE_PATH + "/" + ds.StudyInstanceUID
-    os_helpers.make_directory(study_path)
-
-    # Add study path to storage dictionary
     with lock:
+        ds = event.dataset
+        ds.file_meta = event.file_meta
+
+        # Path for the study
+        study_path = config.SCP_STORAGE_PATH + "/" + ds.StudyInstanceUID
+        os_helpers.make_directory(study_path)
+
+        # Add study path to storage dictionary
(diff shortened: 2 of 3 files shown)


SimonBiggs:
For this, before trying to attempt a fix, can I recommend we do a data gathering exercise first? Could we instead add logging statements throughout your server that make it clear what it is doing. We can then pipe those logs to files which are organised by date.

Here's an example of logging:

https://github.com/pymedphys/pymedphys/blob/9a9fd5bc6a51c4cfa75ae5351d2c1b32915c843e/pymedphys/_icom/listener.py#L24

That way, you can have a look through the logfiles and have a pretty good idea of what happened that lead up to the issue. If there is not enough information, when the issue happens, then more logging may be warranted.

matthewdeancooper:
> For this, before trying to attempt a fix, can I recommend we do a data gathering exercise first? Could we instead add logging statements throughout your server that make it clear what it is doing. We can then pipe those logs to files which are organised by date.
> 
> Here's an example of logging:
> 
> https://github.com/pymedphys/pymedphys/blob/9a9fd5bc6a51c4cfa75ae5351d2c1b32915c843e/pymedphys/_icom/listener.py#L24
> 
> That way, you can have a look through the logfiles and have a pretty good idea of what happened that lead up to the issue. If there is not enough information, when the issue happens, then more logging may be warranted.

Yep good idea.

I'm assuming that if we have not heard any complaints directly then they have found a work around - possibly by exporting individually. Have you heard anything on your end other than noticing the assertion error from the interpreter?

SimonBiggs:
All I noticed was the assertion. Probably worth shooting off an email and
asking about it.

SimonBiggs:
Heya @matthewdeancooper,

I'm going to close this and we can replace it with a PR that includes further logging of the issue?

matthewdeancooper:
Yep sounds good. Btw, all seems to be running well on the deployment end for now. No issues reported, so the assert statement did its job and they must have just re-exported studies that were blocked.

SimonBiggs:
Did you send through an email?

matthewdeancooper:
Yep, got hold of Neil to make sure everything was functioning properly. We
are planning to do a whereby session in the near future to discuss any
limitations/updates for a future release :).

I'm about halfway through Andrew Ng's cs299 ML course - aiming to be in a
position where i can guide a future student on other algorithms should they
come up. Also, thesis will be up on the docs page before Friday.

SimonBiggs:
> Also, thesis will be up on the docs page before Friday.

Awesome :). It'd be great if I could put a direct link in the power point too :).

SimonBiggs:
> I'm about halfway through Andrew Ng's cs299 ML course - aiming to be in a position where i can guide a future student on other algorithms should they come up.

Good plan

PR #1036, opened 2020-09-07
PyMedPhys minimal deps by default

SimonBiggs:
@sjswerdloff there's a bit going on here. 

But essentially I have made it so that all dependencies are optional:
https://github.com/pymedphys/pymedphys/blob/f51535f5dbcf68ed3e3b27cd7effea959d6739b1/pyproject.toml#L32-L96

I have made the management of optional dependencies a little more automated based on the idea over at https://github.com/python-poetry/poetry/issues/1644#issuecomment-639789610. See:
https://github.com/pymedphys/pymedphys/blob/f51535f5dbcf68ed3e3b27cd7effea959d6739b1/scripts/propegate-extras.py#L25-L51

Which writes the following:
https://github.com/pymedphys/pymedphys/blob/f51535f5dbcf68ed3e3b27cd7effea959d6739b1/pyproject.toml#L98-L106

And that `[tool.poetry.extras]` section defines extras that `pip` will use.

And then, I added a test to make sure `pymedphys.experimental` will work without dependencies into the future:

https://github.com/pymedphys/pymedphys/blob/f51535f5dbcf68ed3e3b27cd7effea959d6739b1/.github/workflows/on-push.yml#L141-L147

And lastly, to fix the actual original issue, I either made all imports come from `pymedphys._imports` or moved the module out of `pymedphys.experimental`.


pymedphys/_experimental/cube.py
```diff
@@ -15,10 +15,7 @@
 
 from pymedphys._imports import mpl_toolkits
 from pymedphys._imports import numpy as np
-from pymedphys._imports import plt, pydicom
-
-from scipy.interpolate import splev, splprep
-from scipy.optimize import basinhopping
+from pymedphys._imports import plt, scipy
```
sjswerdloff, line 18: is the removal of pydicom from the import deliberate?

pymedphys/_experimental/cube.py
```diff
@@ -316,11 +313,7 @@ def contour_to_points(contours):
 
 
 def align_cube_to_structure(
-    structure_name: str,
-    dcm_struct: pydicom.dataset.FileDataset,
-    quiet=False,
-    niter=10,
-    x0=None,
+    structure_name: str, dcm_struct, quiet=False, niter=10, x0=None
```
sjswerdloff, line 316: is there a way to not lose the typing of dcm_struct?

pymedphys/_experimental/cube.py
SimonBiggs: Yup, since its only use was to provide the typing in your other comment.

scripts/propegate-extras.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (C) 2020 Simon Biggs
+
```
sjswerdloff, line 2: propagate (spelling)

SimonBiggs: Yup :) Cheers :)

pymedphys/_experimental/cube.py
SimonBiggs: There is a way. The function would need to be wrapped in a function factory. Essentially the function can't be defined at import time.

There might also be another way which is tidier. And that's to allow typing flags to not enact the optional import logic. That'd be a bit more work though. It's this approach that I was considering going down if more typing was to be used through the library. More thought here needed.

SimonBiggs: For now removing it is the easiest way forward.

sjswerdloff (changes requested):
not to be a nit-picker, but it's the spelling of propagate in a file name (which is an implicit interface because of the module name) that has me making a "must be addressed" request.
If the misspelling is from another project (e.g. poetry itself), I'll approve without change

pymedphys/_experimental/cube.py
sjswerdloff: ok by me.
losing the typing is not a huge issue, and the naming of the variable makes it relatively clear what type is expected

pymedphys/_experimental/cube.py
sjswerdloff: that's what i thought.  OK.

SimonBiggs:
> not to be a nit-picker, but it's the spelling of propagate in a file name (which is an implicit interface because of the module name) that has me making a "must be addressed" request.
> If the misspelling is from another project (e.g. poetry itself), I'll approve without change

The misspelling is mine. In this case it's not an interface as it is a "helper script" for which the sole purpose is to help in repo management. In its current location it won't actually end up in the distributed package.

Nevertheless, not a fan of sloppy spelling :) and I shall fix that up before merge.

sjswerdloff (approved):

Issue #1034, opened 2020-09-05
experimental pseudonymisation crashes when sequence is present in data that is listed as an identifying key word


sjswerdloff:
When an RT Structure Set contained RequestAttributesSequence pseudonymisation crashed with a KeyError ('SQ').
pseudonymisation recurses in to sequences that are not listed as identifiers.
But there is no strategy (function) for VR of type SQ
Possible fixes are:
1)  strategy to put an empty "value" in to the Sequence
2)  default identifying keywords for pseudonymisation to have *Sequence removed from the list (this is the work around I've implemented programmatically elsewhere):
>
        identifying_keywords_less_sequences = [
            x
            for x in pseudonymise.get_default_pseudonymisation_keywords()
            if not x.endswith("Sequence")
        ]

The result of #2 would be pseudonymisation of the *contents* of the identifying sequence (which might have value)
I'm open to suggestions on which approach to take.
I haven't tried creating a "replace with empty value" approach.

SimonBiggs:
I think the most natural option is that should the anonymisation method not have an entry for SQ then the approach should recurse into SQ and for each dataset within SQ apply the original anonymisation approach.

I believe this is similar to what you implemented already?

sjswerdloff:
yup.  that maps to solution number 2 above.
and that a "validate" feature discussed previously/elsewhere would indicate that an identifying key with VR of SQ is not compatible with the strategy (if someone is customizing their identifying keyword list).
Will get to it this coming week.
Thanks for the feedback.

SimonBiggs:
I don't think it'd hurt to let people define SQ in the strategy. Maybe I
need to give it more thought...

sjswerdloff:
I wasn't saying don't allow people to define SQ in their strategy, just that it could be avoided in the pseudonymisation strategy.
I've implemented both ways in my PR (#1038 )
The default identifying keys for pseudonymisation no longer contain Sequences.
The pseudonymisation strategy now does provide a SQ replacement function (replace with an empty sequence), but that should only get used if the programmer chooses to *add* a Sequence as an identifying key (which they could do one by one, or they could pull in the default keys for anonymisation, and do a set union (or comprehend just the items that end with "Sequence" and append to the list provided from pseudonymisation).
So the default behaviour for pseudonymisation will be "pseudonymise the contents of all sequences", while leaving open the option for a programmer to replace identifying sequences with an empty sequence (although that has DICOM compliance risks if they have type 1 or 2 elements).
[...]

PR #1038, opened 2020-09-08
fix issue #1034, pseudonymisation crashes on data containing sequences

sjswerdloff:
that were in the list of identifying keywords, e.g. RequestAttributesSequence
Remove keywords that are sequences from the default keyword list provided by pseudonymisation
 so that the contents get pseudonymised
Provide a dictionary entry in the strategy for SQ so that an empty sequence is provided
 if a programmer decides to put an identifier in that is a Sequence
Added error message to anonymise so that when there is a failure, the keyword and it's VR are identified, before raising the error again
TODO: provide a Validator so that keywords get validated against a strategy
Added a test that deals with both approaches
fixes issue #1034

### DIFF
--- a/pymedphys/_experimental/pseudonymisation/__init__.py
+++ b/pymedphys/_experimental/pseudonymisation/__init__.py
@@ -43,9 +43,18 @@ def get_default_identifying_uids():
 
 @functools.lru_cache()
 def _get_default_pseudonymisation_keywords():
-    anon_keyword_set = set(get_default_identifying_keywords())
-    psuedo_uid_set = set(get_default_identifying_uids())
-    return tuple(anon_keyword_set.union(psuedo_uid_set))
+    anon_keyword_list = get_default_identifying_keywords()
+    # The preferred approach is to pseudonymise the contents
+    # of sequences, rather than operate on the sequence itself
+    #
+    # Eliminating the keywords that are sequences fixes issue #1034
+    # for default usage
+    identifying_keywords_less_sequences = [
+        x for x in anon_keyword_list if not x.endswith("Sequence")
+    ]
+    anon_keyword_set = set(identifying_keywords_less_sequences)
+    pseudo_uid_set = set(get_default_identifying_uids())
+    return tuple(anon_keyword_set.union(pseudo_uid_set))
 
 
 def get_default_pseudonymisation_keywords():
--- a/pymedphys/tests/experimental/pseudonymisation/test_pseudonymisation.py
+++ b/pymedphys/tests/experimental/pseudonymisation/test_pseudonymisation.py
@@ -87,7 +87,7 @@ def _test_identifier_is_sequence_vr():
     logging.info("Using keyword with VR = UR")
 
     ds_input = pydicom.Dataset()
-
+    ds_input.PatientID = "ABC123"
     request_attributes_seq = dicom_dataset_from_dict(
         {
             "RequestAttributesSequence": [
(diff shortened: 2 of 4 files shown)


SimonBiggs:
@Matthew-Jennings I'd be super keen if you could cast your eye on this one and think through if there are any implications of removing all identifying keywords that are of the SQ type when undergoing anonymisation.

See https://github.com/pymedphys/pymedphys/issues/1034#issuecomment-688661726 for more context.

sjswerdloff:
> @Matthew-Jennings I'd be super keen if you could cast your eye on this one and think through if there are any implications of removing all identifying keywords that are of the SQ type when undergoing anonymisation.
> 
> See [#1034 (comment)](https://github.com/pymedphys/pymedphys/issues/1034#issuecomment-688661726) for more context.

not for "anonymisation"... only for pseudonymisation.
anonymisation hasn't been channged.

pymedphys/tests/experimental/pseudonymisation/test_pseudonymisation.py
```diff
@@ -73,6 +76,59 @@ def test_identifier_with_unknown_vr():
         )
 
 
+@pytest.mark.pydicom
+def _test_identifier_is_sequence_vr():
```
SimonBiggs, line 80: I'm not at a PC at the moment, but does this test run when it is prepended with an underscore?

pymedphys/_experimental/pseudonymisation/__init__.py
```diff
@@ -43,9 +43,18 @@ def get_default_identifying_uids():
 
 @functools.lru_cache()
 def _get_default_pseudonymisation_keywords():
-    anon_keyword_set = set(get_default_identifying_keywords())
-    psuedo_uid_set = set(get_default_identifying_uids())
-    return tuple(anon_keyword_set.union(psuedo_uid_set))
```
SimonBiggs, line 48: Silly me, I missed that this was explicitly only psudonymisation. A misread from flying through too quick.

SimonBiggs (approved):
I made the silly mistake of thinking that this was impacting both methods. But it is only impacting psudonymisation. Silly me. As long as that test is actually picked up by pytest with the _ at the beginning this is good to go. :)

SimonBiggs (changes requested):
Looks like that test file is only running the tests without the underscore:
https://github.com/pymedphys/pymedphys/runs/1084039094?check_suite_focus=true#step:27:116

So, just one change requested, can you remove the prepended underscore from those functions?

sjswerdloff:
Whoops, yes, I'll fix that test first thing tomorrow morning

SimonBiggs:
> Whoops, yes, I'll fix that test first thing tomorrow morning

Actually, I'm gonna merge and make the fix myself. I might not get a chance to forge a release soonish if I don't do it tonight. So I figure I'll merge this and make that release :).

pymedphys/_experimental/pseudonymisation/__init__.py
```diff
@@ -43,9 +43,18 @@ def get_default_identifying_uids():
...
+    anon_keyword_list = get_default_identifying_keywords()
+    # The preferred approach is to pseudonymise the contents
+    # of sequences, rather than operate on the sequence itself
+    #
+    # Eliminating the keywords that are sequences fixes issue #1034
+    # for default usage
+    identifying_keywords_less_sequences = [
+        x for x in anon_keyword_list if not x.endswith("Sequence")
+    ]
```
SimonBiggs, line 54: Here I think I'd prefer:

```python
identifying_keywords_less_sequences = [
    x for x in anon_keyword_list if pydicom.datadict.dictionary_VR(x) != 'SQ'
]
```

This is equivalent to yours I believe. But this has the benefit of directly testing against the parameter we care about, instead of relying on people to follow an appropriate naming convention.

PR #1044, opened 2020-09-10
Add a function for confirming whether a strategy supports the VR for each of the keywords

sjswerdloff:
fix broken test that was prefixed with an underscore and had mangled  construction of RequestAttributesSequence
include testing that the pseudonymisation does not support VR of UR regardless of what might be in the data by
using the is_valid_strategy_for_keywords function

### DIFF
--- a/pymedphys/_experimental/pseudonymisation/__init__.py
+++ b/pymedphys/_experimental/pseudonymisation/__init__.py
@@ -20,10 +20,11 @@ from os.path import join as pjoin
 from pymedphys._dicom.anonymise import (
     anonymise_directory,
     anonymise_file,
+    get_baseline_keyword_vr_dict,
     get_default_identifying_keywords,
 )
 
-from . import strategy
+from . import strategy  # pylint: disable=import-self
 
 HERE = dirname(abspath(__file__))
 
@@ -115,3 +116,22 @@ def anonymise_with_pseudo_cli(args):
         raise FileNotFoundError(
             "No file or directory was found at the supplied input path."
         )
+
+
+def is_valid_strategy_for_keywords(
+    identifying_keywords=None, replacement_strategy=None
+):
+    test_keywords = identifying_keywords
+    if test_keywords is None:
+        test_keywords = get_default_pseudonymisation_keywords()
+    test_strategy = replacement_strategy
+    if test_strategy is None:
+        test_strategy = strategy.pseudonymisation_dispatch
+
+    baseline_keyword_vr_dict = get_baseline_keyword_vr_dict()
+    for keyword in test_keywords:
+        vr = baseline_keyword_vr_dict[keyword]
+        # pydicom.datadict.dictionary_VR(pydicom.datadict.tag_for_keyword(keyword))
+        if vr not in test_strategy:
+            return False
+    return True
--- a/pymedphys/_experimental/pseudonymisation/__init__.py
+++ b/pymedphys/_experimental/pseudonymisation/__init__.py
@@ -24,7 +24,7 @@ from pymedphys._dicom.anonymise import (
     get_default_identifying_keywords,
 )
 
-from . import strategy  # pylint: disable=import-self
+from . import strategy
 
 HERE = dirname(abspath(__file__))
 
@@ -121,17 +121,16 @@ def anonymise_with_pseudo_cli(args):
 def is_valid_strategy_for_keywords(
     identifying_keywords=None, replacement_strategy=None
 ):
-    test_keywords = identifying_keywords
-    if test_keywords is None:
-        test_keywords = get_default_pseudonymisation_keywords()
-    test_strategy = replacement_strategy
-    if test_strategy is None:
-        test_strategy = strategy.pseudonymisation_dispatch
+    if identifying_keywords is None:
+        identifying_keywords = get_default_pseudonymisation_keywords()
+
+    if replacement_strategy is None:
+        replacement_strategy = strategy.pseudonymisation_dispatch
 
     baseline_keyword_vr_dict = get_baseline_keyword_vr_dict()
-    for keyword in test_keywords:
+    for keyword in identifying_keywords:
         vr = baseline_keyword_vr_dict[keyword]
         # pydicom.datadict.dictionary_VR(pydicom.datadict.tag_for_keyword(keyword))
-        if vr not in test_strategy:
+        if vr not in replacement_strategy:
             return False
     return True
--- a/pymedphys/tests/experimental/pseudonymisation/test_pseudonymisation.py
+++ b/pymedphys/tests/experimental/pseudonymisation/test_pseudonymisation.py
@@ -54,6 +54,12 @@ def test_identifier_with_unknown_vr():
     identifying_keywords_with_vr_unknown_to_strategy = ["CodingSchemeURL", "PatientID"]
     logging.info("Using keyword with VR = UR")
 
+    # test the new "is_valid_strategy_for_keywords", which should indicate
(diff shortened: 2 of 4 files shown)


sjswerdloff:
this provides a means for any strategy and set of identifying keywords to be tested, but it defaults to the keywords and strategy for pseudonymisation, and it's *in* pseudonymisation.
I'd be fine with moving it in to anonymisation (and making the defaults appropriate for anonymisation), and then make the same named function in pseudonymisation with no arguments allowed (and calling the anonymisation version with appropriate keywords and strategy)

I also put in a pylint: disable=import-self 
because that was generating a complaint in my IDE (vs code).
There is the same 
from . import strategy
in anonymise and that is generating a complaint as well.
I'd be happy to add a commit to this PR that puts in the pylint disable for import-self for that line.
But it does beg the question, how does one import a module from the same namespace without getting a pylint complaint (and without disabling the complaint)?  I did a lot of searching and didn't find anything, so... I just disabled the complaint.

pymedphys/tests/experimental/pseudonymisation/test_pseudonymisation.py
```diff
@@ -77,29 +83,31 @@ def test_identifier_with_unknown_vr():
...
 
-    ds_input = pydicom.Dataset()
-    ds_input.PatientID = "ABC123"
-    request_attributes_seq = dicom_dataset_from_dict(
+    ds_input = dicom_dataset_from_dict(
         {
+            "PatientID": "ABC123",
             "RequestAttributesSequence": [
                 {
                     "RequestedProcedureID": "Tumour Identification",
```
SimonBiggs, line 101: Just a preference nit-pick thing here. Can we define run:

```python
identifying_string = "Tumour Identification"
```

And then use that here:

```python
"RequestedProcedureID": identifying_string,
```

And then below change the following:

```python
    assert (
        ds_anon.RequestAttributesSequence[0].RequestedProcedureID
        != "Tumour Identification"
    )
```

to:

```python
    assert (
        ds_anon.RequestAttributesSequence[0].RequestedProcedureID
        != identifying_string
    )
```

Reason being, having a test potentially pass one day due to a typo raises a small alarm bell.

sjswerdloff:
and I should update the API documentation, since there is a new method that has been made available

pymedphys/_experimental/pseudonymisation/__init__.py
```diff
@@ -115,3 +116,22 @@ def anonymise_with_pseudo_cli(args):
...
+
+def is_valid_strategy_for_keywords(
+    identifying_keywords=None, replacement_strategy=None
+):
+    test_keywords = identifying_keywords
+    if test_keywords is None:
+        test_keywords = get_default_pseudonymisation_keywords()
+    test_strategy = replacement_strategy
+    if test_strategy is None:
+        test_strategy = strategy.pseudonymisation_dispatch
```
SimonBiggs, line 129: I think here it's equivalent to write:

```python
    if identifying_keywords is None:
        identifying_keywords = get_default_pseudonymisation_keywords()
    if replacement_strategy is None:
        replacement_strategy = strategy.pseudonymisation_dispatch
```

Then just use `identifying_keywords` and `replacement_strategy` below.

pymedphys/_experimental/pseudonymisation/__init__.py
```diff
@@ -20,10 +20,11 @@
 from pymedphys._dicom.anonymise import (
     anonymise_directory,
     anonymise_file,
+    get_baseline_keyword_vr_dict,
     get_default_identifying_keywords,
 )
 
-from . import strategy
+from . import strategy  # pylint: disable=import-self
```
SimonBiggs, line 27: I've never seen this pylint error before. Would this be able to be overcome by defining `is_valid_strategy_for_keywords` in a separate file?

pymedphys/tests/experimental/pseudonymisation/test_pseudonymisation.py
```diff
@@ -108,23 +116,27 @@ def _test_identifier_is_sequence_vr():
...
-        ds_anon.RequestAttributesSequence.RequestedProcedureID
+        ds_anon.RequestAttributesSequence[0].RequestedProcedureID
         != "Tumour Identification"
     )
     # and an element in the sequence that is not an identifier has been left as is
     assert (
-        ds_anon.RequestAttributesSequence.ScheduledProcedureStepID
+        ds_anon.RequestAttributesSequence[0].ScheduledProcedureStepID
         == "Tumour ID with Dual Energy"
     )
```
SimonBiggs, line 141: Yup, very nice :)

SimonBiggs (approved):
Awesome, looks good to me. Just a few comments throughout, but nothing that would cause an issue.

SimonBiggs:
> I'd be fine with moving it in to anonymisation (and making the defaults appropriate for anonymisation), and then make the same named function in pseudonymisation with no arguments allowed (and calling the anonymisation version with appropriate keywords and strategy)

I think in the near future it'd be worth moving all of this out of experimental. So I think that could be bundled in with that task?

pymedphys/_experimental/pseudonymisation/__init__.py
sjswerdloff: > I've never seen this pylint error before. Would this be able to be overcome by defining `is_valid_strategy_for_keywords` in a separate file?

it wasn't that function in particular, the same import is used in anonymise, presumably for accessing the default keywords or at least the default strategy for anonymisation.

SimonBiggs:
> I also put in a pylint: disable=import-self
> because that was generating a complaint in my IDE (vs code).
> There is the same
> from . import strategy

Ahh, I worked it out. In VSCode, you'll want to open the whole pymedphys git repository for pylint to work sensibly.

image

image

This then allows you to open that `__init__.py` file with no error:

image

pymedphys/_experimental/pseudonymisation/__init__.py
SimonBiggs: I managed to work it out :). Details over at https://github.com/pymedphys/pymedphys/pull/1044#issuecomment-690034690

sjswerdloff:
> I think in the near future it'd be worth moving all of this out of experimental. So I think that could be bundled in with that task?

Yup.  Sounds good.  But I'd like to add two commits to this PR:
1) do the same pylint: disable=import-self for anonymise/\_\_init\_\_.py
2) update the API documentation
I have come to appreciate the good code and documentation hygiene in pymedphys

I think the only way to avoid the import-self situation in an \_\_init\_\_.py is to 
from .module_in_this_directory import ( function1, function2, useful_constant )
but that means one has to be careful about not reusing the same function names in sibling modules in the directory, or
maybe there is a way to alias?

SimonBiggs:
> 1. do the same pylint: disable=import-self for anonymise/__init__.py

Does the comment at https://github.com/pymedphys/pymedphys/pull/1044#issuecomment-690034690 fix the issue and make this pylint override no longer needed?

> 2. update the API documentation
>    I have come to appreciate the good code and documentation hygiene in pymedphys

Yup, that would be awesome. I have higher aspirations than where it currently is too. The only way to get there is to keep on top of it and push it forward :). Thanks @sjswerdloff.

sjswerdloff:
> Does the comment at [#1044 (comment)](https://github.com/pymedphys/pymedphys/pull/1044#issuecomment-690034690) fix the issue and make this pylint override no longer needed?

Yes.  It appears so, even if I just open folder for experimental.  Good enough for me.

I will address the other comments and push those commits up momentarily

SimonBiggs (approved):
Yup, awesome. Thanks @sjswerdloff.

Issue #1047, opened 2020-09-12
VisibleDeprecationWarning problem with jaw


bianka-m:
Hi! I tried using the following code:

`imrt_plan_ds = pydicom.dcmread(file)
beam_sequence = imrt_plan_ds.BeamSequence

delivery = pymedphys.Delivery.from_dicom(imrt_plan_ds, fraction_number=1)`

However, the following warning appears when I run the program

`C:\Users\USER\PycharmProjects\Fluence_Map\venv\lib\site-packages\pymedphys\_dicom\delivery\core.py:170: **VisibleDeprecationWarning**: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  jaw = np.array(dicom_jaw)
Traceback (most recent call last):
  File "C:/Users/USER/PycharmProjects/Fluence_Map/fluence_map.py", line 103, in <module>
    delivery = pymedphys.Delivery.from_dicom(imrt_plan_ds, fraction_number=1)
  File "C:\Users\USER\PycharmProjects\Fluence_Map\venv\lib\site-packages\pymedphys\_dicom\delivery\core.py", line 75, in from_dicom
    delivery_data_by_beam_sequence.append(cls._from_dicom_beam(beam, meterset))
  File "C:\Users\USER\PycharmProjects\Fluence_Map\venv\lib\site-packages\pymedphys\_dicom\delivery\core.py", line 172, in _from_dicom_beam
    second_col = deepcopy(jaw[:, 1])
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
`

SimonBiggs:
Hi @bianka-m,

Would you be able to make sure that the DICOM file is anonymised and then upload it to this GitHub issue by dragging and dropping it onto the comment field?

To anonymise the DICOM file you can run in a command prompt/terminal `pymedphys dicom anonymise path/to/dicom/file.dcm`

And then, you can check that the anonymisation has worked by opening the new file with something like dicompyler (https://www.dicompyler.com/)

bianka-m:
Hi Simon! Here's the anonymised DICOM RTPlan file I'm using 

[NPCA_Anonymised.zip](https://github.com/pymedphys/pymedphys/files/5212087/NPCA_Anonymised.zip)

SimonBiggs:
[...]

PR #1048, opened 2020-09-12
Mark ASYMX as unsupported, fix some brittle DICOM assumptions

SimonBiggs:
See https://github.com/pymedphys/pymedphys/issues/1047 for details.

### DIFF
--- a/pymedphys/_data/urls.json
+++ b/pymedphys/_data/urls.json
@@ -30,5 +30,6 @@
   "rtplan-anonymisation.zip": "https://zenodo.org/record/3930533/files/rtplan-anonymisation.zip?download=1",
   "treatment-record-anonymisation.zip": "https://zenodo.org/record/3931186/files/treatmentrecord-anonymisation.zip?download=1",
   "negative-mu-density.trf": "https://zenodo.org/record/3973722/files/negative_mu_density.trf?download=1",
-  "trf-references-and-baselines.zip": "https://zenodo.org/record/3987725/files/trf-references-and-baselines.zip?download=1"
+  "trf-references-and-baselines.zip": "https://zenodo.org/record/3987725/files/trf-references-and-baselines.zip?download=1",
+  "dicom-to-delivery-issue-#1047.zip": "https://github.com/pymedphys/pymedphys/files/5212087/NPCA_Anonymised.zip"
 }
--- a/pymedphys/_dicom/delivery/core.py
+++ b/pymedphys/_dicom/delivery/core.py
@@ -148,7 +148,7 @@ class DeliveryDicom(DeliveryBase):
 
         if mlc_limiting_device.NumberOfLeafJawPairs != len(leaf_widths):
             raise ValueError(
-                "Expected number of leaf jaw pairs to be the same as "
+                "Expected number of leaf pairs to be the same as "
                 "the length of leaf widths"
             )
 
--- a/pymedphys/_dicom/delivery/core.py
+++ b/pymedphys/_dicom/delivery/core.py
@@ -166,8 +166,6 @@ class DeliveryDicom(DeliveryBase):
             control_points, "BeamLimitingDevicePositionSequence"
         )
 
-        print(beam_limiting_device_position_sequence[0])
-
         dicom_mlcs = rtplan.get_leaf_jaw_position_of_type(
             beam_limiting_device_position_sequence, "MLCX"
         )
--- a/pymedphys/_dicom/delivery/core.py
+++ b/pymedphys/_dicom/delivery/core.py
@@ -128,7 +128,8 @@ class DeliveryDicom(DeliveryBase):
         if rt_beam_limiting_device_types != set(["MLCX", "ASYMY"]):
             raise ValueError(
                 "Currently only DICOM files that contain "
-                "exactly MLCX and ASYMY are supported"
+                "exactly MLCX and ASYMY are supported. "
+                f"{rt_beam_limiting_device_types} is not supported."
             )
 
         mlc_sequence = [
--- a/pymedphys/_dicom/delivery/core.py
+++ b/pymedphys/_dicom/delivery/core.py
@@ -160,7 +160,7 @@ class DeliveryDicom(DeliveryBase):
             control_points, "BeamLimitingDevicePositionSequence"
         )
 
-        dicom_mlcs = rtplan.get_leaf_jaw_position_of_type(
+        dicom_mlcs = rtplan.get_leaf_jaw_positions_for_type(
             beam_limiting_device_position_sequences, "MLCX"
         )
 
@@ -173,7 +173,7 @@ class DeliveryDicom(DeliveryBase):
             for mlc in dicom_mlcs
         ]
 
-        dicom_jaw = rtplan.get_leaf_jaw_position_of_type(
+        dicom_jaw = rtplan.get_leaf_jaw_positions_for_type(
             beam_limiting_device_position_sequences, "ASYMY"
         )
 
--- a/pymedphys/_dicom/delivery/core.py
+++ b/pymedphys/_dicom/delivery/core.py
@@ -120,6 +120,17 @@ class DeliveryDicom(DeliveryBase):
             raise ValueError("Meterset should not ever be None")
 
         beam_limiting_device_sequence = beam.BeamLimitingDeviceSequence
+
+        rt_beam_limiting_device_types = {
+            item.RTBeamLimitingDeviceType for item in beam_limiting_device_sequence
+        }
+
+        if rt_beam_limiting_device_types != set(["MLCX", "ASYMY"]):
+            raise ValueError(
+                "Currently only DICOM files that contain "
+                "exactly MLCX and ASYMY are supported"
+            )
+
         mlc_sequence = [
             item
             for item in beam_limiting_device_sequence
@@ -147,18 +158,18 @@ class DeliveryDicom(DeliveryBase):
             )
 
         num_leaves = len(leaf_widths)
+
         control_points = beam.ControlPointSequence
 
-        beam_limiting_device_position_sequences = rtplan.get_cp_attribute_leaning_on_prior(
+        beam_limiting_device_position_sequence = rtplan.get_cp_attribute_leaning_on_prior(
             control_points, "BeamLimitingDevicePositionSequence"
         )
 
-        print(beam_limiting_device_position_sequences[0])
+        print(beam_limiting_device_position_sequence[0])
 
-        mlcs = [
-            sequence[-1].LeafJawPositions
-            for sequence in beam_limiting_device_position_sequences
-        ]
+        dicom_mlcs = rtplan.get_leaf_jaw_position_of_type(
+            beam_limiting_device_position_sequence, "MLCX"
+        )
 
         mlcs = [
             np.array(
@@ -166,13 +177,12 @@ class DeliveryDicom(DeliveryBase):
                     ::-1
                 ]
             ).T
-            for mlc in mlcs
+            for mlc in dicom_mlcs
         ]
 
-        dicom_jaw = [
-            sequence[0].LeafJawPositions
-            for sequence in beam_limiting_device_position_sequences
-        ]
+        dicom_jaw = rtplan.get_leaf_jaw_position_of_type(
+            beam_limiting_device_position_sequence, "ASYMY"
+        )
 
         jaw = np.array(dicom_jaw)
 
--- a/pymedphys/_dicom/delivery/core.py
+++ b/pymedphys/_dicom/delivery/core.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 from copy import deepcopy
 
 from pymedphys._imports import numpy as np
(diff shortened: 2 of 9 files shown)


pymedphys/_dicom/delivery/core.py
```diff
@@ -134,42 +119,68 @@ def _from_dicom_beam(cls, beam, meterset):
...
 
-        beam_limiting_device_position_sequences = get_cp_attribute_leaning_on_prior(
+        beam_limiting_device_position_sequence = rtplan.get_cp_attribute_leaning_on_prior(
             control_points, "BeamLimitingDevicePositionSequence"
         )
 
-        mlcs = [
-            sequence[-1].LeafJawPositions
-            for sequence in beam_limiting_device_position_sequences
-        ]
```
SimonBiggs, line 154: I am annoyed with past Simon that he chose to reference the "MLC" sequence entry by its index, not by its accompanying "MLCX" type. I can't believe I did that... Not sure what I was thinking...

pymedphys/_dicom/delivery/core.py
```diff
@@ -66,7 +49,7 @@ def from_dicom(cls, dicom_dataset, fraction_number=None):
                     f"   Fraction numbers to choose from are: {fraction_numbers}"
                 )
 
-        beam_sequence, metersets = get_fraction_group_beam_sequence_and_meterset(
+        beam_sequence, metersets = rtplan.get_fraction_group_beam_sequence_and_meterset(
```
SimonBiggs, line 52: I find it an order of magnitude easier to recognise an external function that has been defined in another file if it is instead imported and used this way. I am gradually converting pymedphys over to following this style.

It was recommended within the Google Style Guide.

pymedphys/_dicom/delivery/core.py
```diff
@@ -134,42 +119,68 @@ def _from_dicom_beam(cls, beam, meterset):
...
+        rt_beam_limiting_device_types = {
+            item.RTBeamLimitingDeviceType for item in beam_limiting_device_sequence
+        }
+
+        if rt_beam_limiting_device_types != set(["MLCX", "ASYMY"]):
+            raise ValueError(
+                "Currently only DICOM files that contain "
+                "exactly MLCX and ASYMY are supported. "
+                f"{rt_beam_limiting_device_types} is not supported."
+            )
```
SimonBiggs, line 133: So fundamentally the whole `Delivery` pathway assumes only one set of jaws. Honestly I think that means I should move the `Delivery` API out of stable and into `experimental` and flag it to be superseded by something more representative of what is needed.

pymedphys/tests/delivery/test_from_dicom.py
```diff
@@ -0,0 +1,34 @@
...
+@pytest.fixture
+def dataset():
+    path = pymedphys.zip_data_paths("dicom-to-delivery-issue-#1047.zip")[0]
+
+    return pydicom.dcmread(path)
+
+
+def test_converting_dicom_to_delivery(dataset):
+    with pytest.raises(ValueError, match=r".*not supported.*"):
+        pymedphys.Delivery.from_dicom(dataset)
```
SimonBiggs, line 34: This test was originally designed to replicate https://github.com/pymedphys/pymedphys/issues/1047. Once I had a bit of a look though I saw this was not a small fix. So for now, not supported.

SimonBiggs (commented):
So, this PR started out as bug fix. But then it became the realisation that I wasn't going to be able to fix said bug, and instead I just needed it to give a helpful error message.

...and then it was followed by the looking at my old code in disgust and needing to fix what I had seen.

pymedphys/_dicom/rtplan/core.py
```diff
@@ -203,6 +203,29 @@ def get_gantry_angles_from_dicom(dicom_dataset):
     return result
 
 
+def get_leaf_jaw_position_of_type(
+    beam_limiting_device_position_sequence, rt_beam_limiting_device_type
+):
```
SimonBiggs, line 208: I want to write a test for this function before merge.

SimonBiggs: Alright, ready for review :)

pymedphys/_data/urls.json
```diff
@@ -30,5 +30,6 @@
   "rtplan-anonymisation.zip": "https://zenodo.org/record/3930533/files/rtplan-anonymisation.zip?download=1",
   "treatment-record-anonymisation.zip": "https://zenodo.org/record/3931186/files/treatmentrecord-anonymisation.zip?download=1",
   "negative-mu-density.trf": "https://zenodo.org/record/3973722/files/negative_mu_density.trf?download=1",
-  "trf-references-and-baselines.zip": "https://zenodo.org/record/3987725/files/trf-references-and-baselines.zip?download=1"
+  "trf-references-and-baselines.zip": "https://zenodo.org/record/3987725/files/trf-references-and-baselines.zip?download=1",
+  "dicom-to-delivery-issue-#1047.zip": "https://github.com/pymedphys/pymedphys/files/5212087/NPCA_Anonymised.zip"
```
sjswerdloff, line 34: why not zenodo?

pymedphys/_dicom/rtplan/core.py
```diff
@@ -203,6 +203,28 @@ def get_gantry_angles_from_dicom(dicom_dataset):
     return result
 
 
+def get_leaf_jaw_position_of_type(
```
sjswerdloff, line 206: should at least be named get_leaf_jaw_positions_of_type (positions the plural of position)
and maybe get_leaf_jaw_positions_for_type (but that might be a distinction between the Queen's English and American English)

pymedphys/tests/delivery/test_exported_object.py
```diff
@@ -1,3 +1,18 @@
+# Copyright (C) 2019 Simon Biggs
```
sjswerdloff, line 1: If you added the copyright in 2020 but the body of work was done in 2019, maybe it should indicate either 2020, or 2019 and 2020?

pymedphys/tests/dicom/rtplan/test_leaf_jaw_pos.py
```diff
@@ -0,0 +1,56 @@
...
+
+from pymedphys._dicom import create, rtplan
+
+
+def test_get_leaf_jaw_position_of_type():
+
+    rt_beam_limiting_device_type = "ASYMY"
+    leaf_jaw_positions = [0, 0]
+
+    expected_positions = [leaf_jaw_positions] * 4
```
sjswerdloff, line 24: so this means expected_positions = [[0, 0], [0, 0], [0, 0], [0, 0] ] ?
or did you mean expected_positions = [0] * 4
For a single leaf pair Jaw (e.g. Y or ASYMY), there are two positions, e.g. [ -11, 10 ]
If you intend to have the wrong number of values set (as part of your test), then make that clear in the variable naming or at least comment.
Something isn't right here, I'm just not sure what your intent was.  But I think it's the content of the expected_positions.

pymedphys/tests/dicom/rtplan/test_leaf_jaw_pos.py
```diff
@@ -0,0 +1,56 @@
...
+                    ]
+                },
+                {},
+                {},
+                {},
+            ]
+        }
+    ).ControlPointSequence
+
+    beam_limiting_device_position_sequences = rtplan.get_cp_attribute_leaning_on_prior(
```
sjswerdloff, line 48: I'd need to go dig through to see exactly what get_cp_attribute_leaning_on_prior offers as it's contract, but if you are passing in a sequence with a nearly arbitrary number of items, I'm surprised you don't have to pass in the control point index (not the item index, but the control point index).  Maybe it defaults to zero, which would be reasonable.

pymedphys/tests/dicom/rtplan/test_leaf_jaw_pos.py
```diff
@@ -0,0 +1,56 @@
...
+        {
+            "ControlPointSequence": [
+                {
+                    "BeamLimitingDevicePositionSequence": [
+                        {
+                            "RTBeamLimitingDeviceType": rt_beam_limiting_device_type,
+                            "LeafJawPositions": leaf_jaw_positions,
+                        },
+                        {
+                            "RTBeamLimitingDeviceType": "Something Else!",
```
sjswerdloff, line 36: perhaps ASYMX would be appropriate.  I wouldn't deliberately put in something that fails DICOM conformance unless that is what I'm testing for.

sjswerdloff (changes requested):
I think test_leaf_jaw_pos.py is not right.  Please give that a hard look.
pluralization of the new function name is not a must have, but I encourage at least that.

pymedphys/_data/urls.json
SimonBiggs: The data was uploaded by a user over at https://github.com/pymedphys/pymedphys/issues/1047#issuecomment-691389682

...didn't feel right to re-upload to a different location given it wasn't my data... In the past I have asked users to upload to Zenodo themselves, but the barrier to entry is larger than "drag and drop onto a GitHub issue".

pymedphys/tests/dicom/rtplan/test_leaf_jaw_pos.py
SimonBiggs: I've seen that often in DICOM files if a leaf and/or jaw position doesn't change in one control point to the next then it actually just uses the result of the previous control point. As in, if the position isn't defined it is presumed to not move with respect to the previous control point.

https://github.com/pymedphys/pymedphys/blob/2cf5dacc50fd841c9a9ab07dd97a06fcb748b80e/pymedphys/_dicom/rtplan/core.py#L162-L178

I believe this may also shed light on https://github.com/pymedphys/pymedphys/pull/1048#discussion_r487453031

pymedphys/tests/dicom/rtplan/test_leaf_jaw_pos.py
SimonBiggs: > so this means expected_positions = [[0, 0], [0, 0], [0, 0], [0, 0] ] ?

Yup. Does https://github.com/pymedphys/pymedphys/pull/1048#discussion_r487467296 shed some light on this?

pymedphys/tests/dicom/rtplan/test_leaf_jaw_pos.py
SimonBiggs: Good call. I am happy to fall in line with that stylistic approach.

pymedphys/_data/urls.json
sjswerdloff: Fair enough.

pymedphys/_dicom/rtplan/core.py
SimonBiggs: Yup, I agree. I'm a little unsure on the of vs for, but I think you're right...

pymedphys/tests/delivery/test_exported_object.py
SimonBiggs: Yeah, I don't think it'd hurt here to have 2019-2020.

pymedphys/tests/dicom/rtplan/test_leaf_jaw_pos.py
SimonBiggs: Because I added in 3 extra control points:

https://github.com/pymedphys/pymedphys/blob/754eff8c49d1bfbaad4b1062d1a2f2b902e68a00/pymedphys/tests/dicom/rtplan/test_leaf_jaw_pos.py#L41-L43

I expected there to be four control points.

pymedphys/tests/dicom/rtplan/test_leaf_jaw_pos.py
SimonBiggs: > (not the item index, but the control point index)

I guess I had hoped... potentially inappropriately, that the order of the control point sequence is consistent. I'm not sure, is the control point order within `ControlPointSequence` strictly increasing?

SimonBiggs:
> I think test_leaf_jaw_pos.py is not right. Please give that a hard look.

Something that may be an issue is https://github.com/pymedphys/pymedphys/pull/1048#discussion_r487468133.

pymedphys/tests/dicom/rtplan/test_leaf_jaw_pos.py
sjswerdloff: Yes, there is a Run Length Encoding approach specified in DICOM.
But the standard also says that if the value *never* changes, you can use the RLE approach, and just put it in with the control point with Control Point Index == 0.  
And if it does *ever* change, you need to put in a value in each control point (even if it is all duplicates except for one).

What the code tells me is that the function returns a list of values, with the list being of length "Number of Control Points", and if the value is itself a list, you are going to receive a list of lists.
I encourage you to put in docstring that says something on the order of
"""returns a list of values, one list item per control point, unpacking any run length encoding
"""

Also, will the values returned be of the conceptual type most closely represented in Python, e.g.
a floating point value if the DICOM Element has a VR of DS, or will it be literally of the DICOM VR (so DS comes back as a string)?
If the attribute is a Sequence (VR is SQ), is the return a list of dataset or a list of iterables of the Sequence Item?

I think it will be very useful for it to be clear what the function returns for a variety of attributes.
While I'm not a fan of changing signatures, the name of the function suggested to me that it was trying to get a single value.  I'm not sure of the python terminology, but you are slicing across a dimension (the attribute axis?) (and getting back something that is of the dimension "dimension of attribute value + 1" (where the 1 comes from the control point/index axis).  For something like a Gantry Angle, the value is a scalar and has dimension 0.
If this was new code, I'd push for a name that more clearly reflects that.

sjswerdloff: > I'm not sure, is the control point order within `ControlPointSequence` strictly increasing?

It's a pretty safe bet.  But it's not strictly guaranteed by the standard that items in a sequence will be ordered monotonically increasing according to the index element that is used to formally identify the sequence item when it is being referenced.  I have not run in to data that came out of order for control points.  But that doesn't mean a bit of paranoia is unjustifiable.

sjswerdloff (approved):

Issue #1011, opened 2020-08-17
Refine and simplify the DICOM anonymise API


SimonBiggs:
> if you mean to condense the various "knobs and buttons" that are currently used to control what happens within the anonymisation, then yes, having a single object represent the set of values currently defined separately:
replace_values=True, keywords_to_leave_unchanged=(), delete_private_tags=True, delete_unknown_tags=None, copy_dataset=True, replacement_strategy=None, identifying_keywords=None
While it could be done as a dictionary, it might make sense to incorporate it in to a Class, so the class can be derived from, and in that class, have the "keys validate against the dispatch table" (potentially automatically at creation and when any particular control is modified, as well as explicit if a programmer wants to confirm the strategy is valid before sending it in)
This also helps to avoid people mutating the dictionary unintentionally (although just having a function that returns a copy rather than directly hitting at the pseudonymisation_dispatch would be appropriate and is simple).
But it won't reduce the work a programmer has to do.  It will simply shift it from the anonymisation call to the specification of the contents of the strategy.  The benefit is in making a clear separation of where the concern is addressed.
It might motivate creation of either a variety of child classes or a set of factory creation methods, but it will at least shift the signature from invocation of anonymisation over to creation of a strategy object.

_Originally posted by @sjswerdloff in https://github.com/pymedphys/pymedphys/pull/1009_

SimonBiggs:
> Yup, I think an "anonymisation strategy"/config object which has validation methods on initialisation would be quite a tidy way to go.
> 
> Building it as a frozen dataclass under the hood might be a neat solution:
> [docs.python.org/3/library/dataclasses.html#frozen-instances](https://docs.python.org/3/library/dataclasses.html#frozen-instances)
> 
> (It would be nice to be able to use [python.org/dev/peps/pep-0603](https://www.python.org/dev/peps/pep-0603/) though...)

Originally posted by @SimonBiggs in #1009

sjswerdloff:
frozenmap (python.org/dev/peps/pep-0603) would be an elegant and transparent way of doing this but then users are restricted to python 3.9+
dataclass would require Python 3.7+
Looks like Python 3.6 is still a rather popular version of 3.x
see:
https://dev.to/hugovk/python-version-share-over-time-6-1jb8
and
https://w3techs.com/technologies/details/pl-python/3

[...]

PR #1053, opened 2020-09-15
Refine anonymise api per issue #1011 

sjswerdloff:


pymedphys/tests/dicom/test_anonymise.py
```diff
@@ -171,6 +172,19 @@ def test_anonymised_dataset_with_empty_patient_sex():
     assert hardcode_replace_ds["PatientSex"].value is None
 
 
+@pytest.mark.dicom
+def test_anonymised_dataset_with_strategy():
+    replacement_strategy = get_copy_of_strategy()
+    ds = dicom_dataset_from_dict(
+        {"PatientName": "Smith^John", "PatientID": "ABC123", "PatientSex": "M"}
+    )
+    assert is_anonymised_dataset(ds)
```
sjswerdloff, line 181: Wait... this passed automated tests.
What?

sjswerdloff:
@SimonBiggs PyMedPhys cache failures?

sjswerdloff:
It's not finding immutables.
I used poetry update and poetry install and they seemed happy, so... I'm at a loss and could use some help.

sjswerdloff:
OK... so I think this means that the import needs to take a special route so that the pyproject.toml can have immutables specified as optional and a dicom group type thing.  I'll investigate in the next 24 hours (but not in the next 8 hours...)

pymedphys/_experimental/pseudonymisation/__init__.py
```diff
@@ -11,12 +11,15 @@
...
 # limitations under the License.
 
+
 import functools
 import json
 import logging
 from os.path import abspath, basename, dirname, isdir, isfile
 from os.path import join as pjoin
 
+from immutables import Map as frozenmap
```
SimonBiggs, line 21: Yup, key is running:

```
poetry add immutables
```

Then, within pyproject.toml change the version number to * and add optional like the others.

Then, within here we actually need to import it with `from pymedphys._imports import immutables`.

And lastly, add immutables to the following file:

https://github.com/pymedphys/pymedphys/blob/8b2b88f3659d7df3907dc17ff378987f2df2c493/pymedphys/_imports/imports.py#L6

pymedphys/_experimental/pseudonymisation/__init__.py
```diff
@@ -11,12 +11,16 @@
...
+
 import functools
 import json
 import logging
 from os.path import abspath, basename, dirname, isdir, isfile
 from os.path import join as pjoin
 
+from pymedphys._imports import immutables
+
+frozenmap = immutables.Map
```
SimonBiggs, line 23: Unfortunately, as useful as this is, for now I think it may be more trouble that it is worth. `pymedphys._imports` uses apipkg under the hood (https://pypi.org/project/apipkg/).

The key is, when an attribute on `immutables` is accessed that's when the lazy import "magic" is activated. I don't like using magic... but in this case, the benefit of optional imports justifies the magic I believe...

So, essentially, I imagine in the "clean" test, you'll now be getting an error at this line, as it is trying to import `immutables` when the attribute is accessed.

pyproject.toml
```diff
@@ -54,6 +55,7 @@ dbfread = {version = "*", optional = true}  # groups = ["user", "dev"]
 pydicom = {version = ">=2.0.0", optional = true}  # groups = ["user", "dev", "dicom"]
 pynetdicom = {version = "*", optional = true}  # groups = ["user", "dev", "dicom"]
 pylibjpeg-libjpeg = {version = "*", optional = true}  # groups = ["user", "dev", "dicom"]
+immutables = {version = "*", optional = true}  # groups = ["user", "dev", "dicom"]
```
SimonBiggs, line 58: When adding this, can you also run the following script:

https://github.com/pymedphys/pymedphys/blob/master/scripts/propagate-extras.py

sjswerdloff: when I went to run propagate-extras.py:
``python scripts/propagate-extras.py``
It complained about a missing package, so I did
poetry install -E dev
and I noticed the following:
- Updating numpy (1.19.1 -> 1.18.5)
while all other updates were more like:
 - Updating pylint (2.5.3 -> 2.6.0)
so... lower semantic version first, newer/higher semantic version second.

On the plus side, the poetry install -E dev fins and running propagate-extras.py is done, so... pushing

pymedphys/_experimental/pseudonymisation/__init__.py
sjswerdloff: I agree the optional import justifies the magic.
But learning magic is harder than learning physics, so... iterating on the incantations

sjswerdloff:
i don't know if the "1 unused import" is something to be concerned about, but everything passed, so I'm thinking it's an outcome of lazy import magic.

sjswerdloff:
@SimonBiggs I think this is good to go, less perhaps some symbol names...

SimonBiggs:
> i don't know if the "1 unused import" is something to be concerned about, but everything passed, so I'm thinking it's an outcome of lazy import magic.

Yeah, that's fine. It's just complaining about the `import immutables` within `pymedphys._imports.imports.py`

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -133,6 +134,26 @@ def anonymise_dataset(  # pylint: disable = inconsistent-return-statements
...
+        if "delete_unknown_tags" in replacement_strategy:
+            delete_unknown_tags = replacement_strategy["delete_unknown_tags"]
+        if "copy_dataset" in replacement_strategy:
+            copy_dataset = replacement_strategy["copy_dataset"]
+        # unlike the modifiers, if identifying keywords were passed in as parameters, use those
+        if (
+            identifying_keywords is not None
+            and "identifying_keywords" in replacement_strategy
+        ):
+            identifying_keywords = replacement_strategy["identifying_keywords"]
```
SimonBiggs, line 155: I think if we're going to go this route, I think we need to raise an error if `replace_values` and the other parameters are defined by the user as well as `replacement_strategy`. To achieve that, I believe we need to default these params to `None` and then raise a `ValueError` if `replacement_strategy` is not None and if any of the parameters that are inherited from `replacement_strategy` are also None.

SimonBiggs (commented):
Didn't get a chance to properly read, and I don't think I'll get a chance till tomorrow evening to read further. :) Made a comment though :)

pymedphys/_dicom/anonymise/__init__.py
SimonBiggs: I believe I didn't say the above correctly. But the heart of what I was trying to say is O don't want the user to be able to write:

```python
anonymise(replace_values=False, replacement_strategy=strategy)
```

And then have the code run off and override the parameter that the user passed. I was under the impression we were going to remove those parameters entirely and let replacement_strategy handle them.

Alternatively, if replacement strategy is left as None by the user, and if we want the current API to be able to carry on in unison, then it could be that all those parameters can stay in the function signature, it's just that they're not allowed to be defined by the user if replacement_strategy has also been defined.

sjswerdloff: > I was under the impression we were going to remove those parameters entirely and let replacement_strategy handle them.

Yes.  But I wanted to start by deprecating the current interface, and then follow with obsoleting it.
Further, the appropriate way to get the result of replace_values=False is to create a strategy that provides that approach, and eliminate the replace_values parameter entirely.  Although I can't recommend using such a strategy because it will likely set type 1 parameters to have a value of the empty string, which is not DICOM conformant (and very few applications will succeed with type 1 element non-conformance).

I'm working on putting a facade in front of the previous function calls with the new API (renaming the old ones by prefixing with and underscore to make them private).  That way it's just the CLI that has a directive to clear the values, and that will get the "clear values strategy" and *all* strategies have replace_values=True and any strategy that has a value of False for that will get treated as not valid.  And then I can delete the special case behaviour for that.

I've got some test cases to write.

sjswerdloff:
This just rips off the band-aid...
But the CLI is preserved.
I didn't try to get rid of the keywords_to_leave_unchanged, but that is mostly under the hood / hidden behind the public API.

sjswerdloff:
I'm at a loss for why the clear part of the job is failing.
Avoiding the import and from import in the same module would probably require that I change the get_copy_of_strategy() functions to have different names. e.g. get_pseudonymisation_strategy() and get_hardcode_strategy(), or alternatively, I can use the namespace (not do from anonymisation import (x, y, z))

sjswerdloff:
The deprecation package and decorator don't seem to provide a means for deprecating specific arguments to a function, just deprecating the entire function.

docs/ref/lib/experimental/pseudonymisation.rst
```diff
@@ -17,15 +17,24 @@ Summary
...
+    and also contain the identifying keywords that the strategy is intended to address.
+    Further, the strategies contain the boolean (or Tri-State) modifiers that control aspects of the anonymisation:
 
+  *  delete_private_tags
+  *  delete_unknown_tags
 
+    To modify the behaviour regarding private tags and unknown tags, it is necessary to construct a new dictionary like container,
+    potentially another immutable map with the new values for delete_private_tags and/or delete_unknown_tags
+
+    See PEP 603, or the PyPi immutables package for more information
```
SimonBiggs, line 31:
```suggestion
    See PEP 603, or the PyPI immutables package for more information
```

"**Py**thon **P**ackage **I**ndex"

docs/ref/lib/experimental/pseudonymisation.rst
```diff
@@ -17,15 +17,24 @@ Summary
     Pseudonymisation provides a list of identifying keywords that includes UIDs
     and an anonymisation strategy that utilises SHA3_256 to hash text which is then
     encoded as base64, or for UIDs, converted to an integer appended to the PyMedPhys Org Root.  Dates are shifted consistently and ages are jittered.
-    The list of identifying keywords and the anonymisation strategy are passed in to
-    the (non-experimental) anonymisation module/functions.
+    The anonymisation strategy is passed in to the (non-experimental) anonymisation module/functions.
+    Anonymisation strategies are immutable maps that contain dispatch tables for replacing the values of identifying elements,
```
SimonBiggs, line 21: Target audiences are programmers and non-programmers alike, only common denominator that can be assumed is that they are a Medical Physicist. Works like "immutable" "dispatch", "Tri-State" are all words I wouldn't expect a Medical Physicist to necessarily understand.

Might it be worth writing a doctest styled string (https://docs.python.org/3/library/doctest.html). For an example see:

https://github.com/pymedphys/pymedphys/blob/991dd077b6a6fa1e824d8490b5138025d492a0ad/pymedphys/_mudensity/mudensity.py#L88-L156

Which is then propagated into the docs over at:

https://docs.pymedphys.com/ref/lib/mudensity.html#pymedphys.mudensity.calculate

...also, of note, the following lines:

https://github.com/pymedphys/pymedphys/blob/2bf503be14227f5c460056428a65fdddc2ec01a5/.github/workflows/on-push.yml#L251-L256

Actually dives in and runs all of these examples to make sure that they work as written.

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -64,7 +65,47 @@ def create_filename_from_dataset(ds, dirpath=""):
     return pjoin(dirpath, "{}.{}.dcm".format(mode_prefix, ds.SOPInstanceUID))
 
 
-def anonymise_dataset(  # pylint: disable = inconsistent-return-statements
+def anonymise_dataset(ds, copy_dataset=True, replacement_strategy=None):
```
SimonBiggs, line 68:
```suggestion
def anonymise_dataset(ds, copy=True, strategy=None):
```

Also, can we potentially simplify the API down to just "strategy"?

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -64,7 +65,47 @@ def create_filename_from_dataset(ds, dirpath=""):
...
+    **We do not yet claim conformance to any DICOM Application Level
+    Confidentiality Profile**, but plan to be in a position to do so in the
+    not-to-distant future.
+
+    Parameters
+    ----------
+    ds : ``pydicom.dataset.Dataset``
+        The DICOM dataset to be anonymised.
+
+    copy : ``bool``, optional
```
SimonBiggs, line 85:
```suggestion
    copy : ``bool``, default ``True``
```

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -64,7 +65,47 @@ def create_filename_from_dataset(ds, dirpath=""):
...
+    ds_anon : ``pydicom.dataset.Dataset``
+        An anonymised version of the input DICOM dataset.
+    """
+
+    if replacement_strategy is None:
+        replacement_strategy = get_copy_of_strategy()
+
+    with replacement_strategy.mutate() as strategy_copy:
+        strategy_copy["copy_dataset"] = copy_dataset
+    replacement_strategy = strategy_copy.finish()
```
SimonBiggs, line 103:
```suggestion
        replacement_strategy = strategy_copy.finish()
```

...but also, I'd much prefer to simply pass the copy variable through to `_anonymise_dataset` instead of having an inconsistent approach between the two functions.

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -133,6 +175,27 @@ def anonymise_dataset(  # pylint: disable = inconsistent-return-statements
         An anonymised version of the input DICOM dataset.
     """
 
+    if replacement_strategy is not None:
```
SimonBiggs, line 178: I believe now this will never be None?

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -133,6 +175,27 @@ def anonymise_dataset(  # pylint: disable = inconsistent-return-statements
...
+        if "keywords_to_leave_unchanged" in replacement_strategy:
+            keywords_to_leave_unchanged = replacement_strategy[
+                "keywords_to_leave_unchanged"
+            ]
+        if "delete_private_tags" in replacement_strategy:
+            delete_private_tags = replacement_strategy["delete_private_tags"]
+        if "delete_unknown_tags" in replacement_strategy:
+            delete_unknown_tags = replacement_strategy["delete_unknown_tags"]
+        if "copy_dataset" in replacement_strategy:
+            copy_dataset = replacement_strategy["copy_dataset"]
```
SimonBiggs, line 190: As mentioned above, preference here to have this variable be defined as a function input, just to have approaches between the functions be more consistent and less cognitive load on someone reading the code :)

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -64,7 +65,47 @@ def create_filename_from_dataset(ds, dirpath=""):
...
+        strategy_copy["copy_dataset"] = copy_dataset
+    replacement_strategy = strategy_copy.finish()
+
+    return _anonymise_dataset(ds, replacement_strategy=replacement_strategy)
+
+
+def _anonymise_dataset(
     ds,
     replace_values=True,
     keywords_to_leave_unchanged=(),
```
SimonBiggs, line 111: Even if it is private, I think this here should default to None (as with every other parameter that can be overwritten). That way if `keywords_to_leave_unchanged` is not None and `keywords_to_leave_unchanged` is defined within the replacement strategy an error can be raised, as no matter what the user of the function provides to `keywords_to_leave_unchanged` it will be silently overwritten by the contents of `replacement_strategy`.

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -735,17 +903,90 @@ def get_anonymous_replacement_value(
...
+        strategy_copy["keywords_to_leave_unchanged"] = ()
+        strategy_copy["delete_private_tags"] = True
+        strategy_copy["delete_unknown_tags"] = None
+        strategy_copy["copy_dataset"] = True
+        strategy_copy["identifying_keywords"] = get_default_identifying_keywords()
+        strategy_map = strategy_copy.finish()
+    is_valid_strategy_for_keywords(
+        strategy_map["identifying_keywords"], replacement_strategy=strategy_map
+    )
+    return strategy_map
```
SimonBiggs, line 992: Given the size of this file, it's getting a unwieldy. We need to split out key concepts into their own file. It would make sense that these functions here would be split over about three files I would imagine.

It would make sense to me to have the top-level user exposed facade functions all in one file called `api.py`, any strategy based functions be within a `strategy.py`, and then maybe make a `core.py` for the rest of the code...

SimonBiggs (changes requested):
We've certainly managed to grow this one into what one would call a large pull request :). We'll have to work out ways to split these up into smaller pieces in the future.

I do have a range of change requests a few are API preferences of mine, some to do small refining of the docs, but the key ones really are still around my initial comment/request of making it so that any function (exposed or not) won't have within its contract/API variables that for some combination of them they may be silently ignored/overwritten.

pymedphys/_dicom/anonymise/__init__.py
sjswerdloff: If you want to modify DicomBase and test_base, yes.
This was left in/put back in to avoid having to dig down further in to other "unrelated" code.

pymedphys/_dicom/anonymise/__init__.py
sjswerdloff: Agreed.

pymedphys/_dicom/anonymise/__init__.py
sjswerdloff: I don't think it's correct to finish and overwrite the original frozenmap within the "with" statement.
For a single change, I could have used a different api that was a one liner, so I'll investigate doing that.

docs/ref/lib/experimental/pseudonymisation.rst
sjswerdloff: I will provide more explanation of terminology, and take a look at the references

pymedphys/_dicom/anonymise/__init__.py
sjswerdloff: I think it is None when invoked from DicomBase and tested in test_base

pymedphys/_dicom/anonymise/__init__.py
sjswerdloff: Other than copy_dataset, I think the other named parameters have been eliminated now.
I can go back in and eliminate the if statements and just assign as needed, or push the strategy further down.

pymedphys/_dicom/anonymise/__init__.py
sjswerdloff: OK.

sjswerdloff:
>We've certainly managed to grow this one into what one would call a large pull request :). We'll have to work out ways to split these up into smaller pieces in the future.

It became much larger in response to the comments requesting deprecation of the older API, which turned in to obsoleting of the older API.  Originally, I wanted to "add first, and make backward compatible", and follow with a separate PR that addressed deprecating or obsoleting.  

The pseudonymisation is dependent on anonymisation, so a non-backward compatible change in anonymisation requires modification to pseudonymisation.

SimonBiggs:
> It became much larger in response to the comments requesting deprecation of the older API, which turned in to obsoleting of the older API. Originally, I wanted to "add first, and make backward compatible", and follow with a separate PR that addressed deprecating or obsoleting.

Yup, certainly. I'll jump on reviewing this tonight.

pymedphys/_dicom/anonymise/__init__.py
SimonBiggs: What if this function is adjusted to have its signature be:

```python
def _anonymise_dataset(
    ds,
    replace_values=True,
    keywords_to_leave_unchanged=(),
    delete_private_tags=True,
    delete_unknown_tags=None,
    copy_dataset=True,
):
```

And then it is called within `anonymise_dataset` as:

```python
_anonymise_dataset(
    ds, replace_values=strategy["replace_values"],
    keywords_to_leave_unchanged=strategy["keywords_to_leave_unchanged"],
    delete_private_tags=strategy["delete_private_tags"],
    delete_unknown_tags=strategy["delete_unknown_tags"],
    copy_dataset=copy
)
```

That way the following lines won't be needed:

```python
    with replacement_strategy.mutate() as strategy_copy:
        strategy_copy["copy_dataset"] = copy_dataset
    replacement_strategy = strategy_copy.finish()
```

And none of the following logic would be needed:

```python
    if replacement_strategy is not None:
        if "replace_values" in replacement_strategy:
            replace_values = replacement_strategy["replace_values"]
        if "keywords_to_leave_unchanged" in replacement_strategy:
            keywords_to_leave_unchanged = replacement_strategy[
                "keywords_to_leave_unchanged"
            ]
        if "delete_private_tags" in replacement_strategy:
            delete_private_tags = replacement_strategy["delete_private_tags"]
        if "delete_unknown_tags" in replacement_strategy:
            delete_unknown_tags = replacement_strategy["delete_unknown_tags"]
        if "copy_dataset" in replacement_strategy:
            copy_dataset = replacement_strategy["copy_dataset"]

        if (
            identifying_keywords is None
            and "identifying_keywords" in replacement_strategy
        ):

            identifying_keywords = replacement_strategy["identifying_keywords"]
```

And tests that use the old API can just point to `_anonymise_dataset`, new API can point to `anonymise_dataset`.

What are your thoughts?

SimonBiggs: Scrap that, let's try again, downside of trying to quickly parse a lot of code, easy to make silly mistakes.

What if, for now, this function is left as it is

```python
def _anonymise_dataset(
    ds,
    replace_values=True,
    keywords_to_leave_unchanged=(),
    delete_private_tags=True,
    delete_unknown_tags=None,
    copy_dataset=True,
    replacement_strategy=None,
    identifying_keywords=None,
):
```

And then it is called within `anonymise_dataset` as:

```python
_anonymise_dataset(
    ds, replace_values=strategy["replace_values"],
    keywords_to_leave_unchanged=strategy["keywords_to_leave_unchanged"],
    delete_private_tags=strategy["delete_private_tags"],
    delete_unknown_tags=strategy["delete_unknown_tags"],
    copy_dataset=copy,
	replacement_strategy=strategy,
    identifying_keywords=replacement_strategy["identifying_keywords"],
)
```

I believe this would address my primary issue with this PR and simplify the logic.

What are your thoughts?

SimonBiggs:
> It is desirable to have smaller Pull Requests. I'm not happy about how large this became either.

Actually, I feel like this PR may grind to a halt. Can we pull out the easy parts (say including the immutables dependency) into other PRs so that the easy stuff can merge.

I think it might be helpful to have a single PR that splits up `anonymise/__init__.py` into three files, but doesn't actually change anything.

And then, once all the low hanging fruit is merged, lets then, and only then, let's tackle this API issue.

A good hint that we're in trouble is that when the most important part of the PR is being hidden by GitHub because it is saying the changes to that file are too large:

image

SimonBiggs:
The next PR after https://github.com/pymedphys/pymedphys/pull/1062 should probably be the organisation of all the functions within `__init__.py` out into other files. But, that PR would have clashed with this PR and made the changelog here neigh unreadable. So I figured I'd let you do that one so that I don't ruin this PR :).

SimonBiggs:
I find I have an issue with bikeshedding (https://en.wiktionary.org/wiki/bikeshedding) ... it's hard to not get bogged down in unimportant issues.

I need to have the following in my head:

> In general, reviewers should favor approving a CL once it is in a state where
it definitely improves the overall
code health of the system
being worked on, even if the CL isn't perfect.

-- https://github.com/google/eng-practices/blob/master/review/reviewer/standard.md#L31-L34

So let's get this PR merged, even if it is large. The overwritten parameters are my main bug bear. Everything else can be delayed into a separate PR after this merges. What are your thoughts on that?

sjswerdloff:
bikeshedding and this PR:
I've already made my way through to splitting up the code in to the separate files.
That's a refactoring that needed to be done because of file size.
You said that it was best to refactor code first before proceeding with changes, and I agree with that.
So...  it's not like the work that I did in this branch is going to get thrown away.  
Instead, it's going to get used as a model (if not copy/pasted from) for a better, cleaner, attempt, with smaller steps.

I just haven't closed this PR yet because I want it to stay readily visible to me until I'm far enough along on replacing it.

SimonBiggs:
> So... it's not like the work that I did in this branch is going to get thrown away.
> Instead, it's going to get used as a model (if not copy/pasted from) for a better, cleaner, attempt, with smaller steps.

Yup, fair enough. Make sure to let me know if I'm being too painful.

sjswerdloff:
Decided not only that the PR as a whole isn't desirable, but the general approach isn't desirable (to me, anyway)

PR #1060, opened 2020-09-23
Add immutables

SimonBiggs:


sjswerdloff (approved):

SimonBiggs:
Needed to run `poetry update`, running that now

sjswerdloff (approved):
was this updated poetry.lock driven by a run of propagate-extras.py or was it due to an update to pyproject.toml that didn't get followed by poetry update ?
There's a fair amount of magic involved, so it's good to know which incantations are needed and when...

SimonBiggs:
Yeah, it was me running `propagate-extras.py` but not following that with a `poetry update`. That way when the doctest ran the equivalent of `pip install pymedphys[user]` it didn't actually install immuatbles. Doctest shouldn't actually need immutables though. I really should just tell it to ignore that `_imports.imports.py` file.

SimonBiggs:
Yeah, there is quite a bit of magic in `poetry`. But it is certainly worth the trade off. `propagate-extras.py` however... that I need to make a bit cleaner.

PR #1062, opened 2020-09-23
Refactor to using a facade, no actual changes though

SimonBiggs:


pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -34,13 +34,22 @@
...
 
 IDENTIFYING_KEYWORDS_FILEPATH = pjoin(HERE, "identifying_keywords.json")
 
 
+def anonymise(ds, copy=True, strategy=None):
+    if strategy is None:
+        return _anonymise_dataset(ds, copy_dataset=copy)
+
+    return _anonymise_dataset(
+        ds, copy_dataset=copy, replacement_strategy=strategy, **strategy
```
sjswerdloff, line 49: this exposes each VR string as a variable (i.e. DS becomes a variable).
I get that this is a compact representation, but it leans towards implicit instead of explicit.

pymedphys/_dicom/anonymise/__init__.py
```diff
@@ -73,6 +82,7 @@ def anonymise_dataset(  # pylint: disable = inconsistent-return-statements
     copy_dataset=True,
     replacement_strategy=None,
     identifying_keywords=None,
+    **_,
```
sjswerdloff, line 85: presumably this prevents the remaining keys of the unpacked strategy dict/map from being exposed as variables/parameters.
I would document what this means because it's not a common usage.
A physicist who doesn't know what a dispatch table is will have difficulty understanding what **_  means, 
When I did a search, I got:
https://stackoverflow.com/questions/13726155/is-an-acceptable-way-to-ignore-arguments-in-python
and the approach (which was for *_ not **_) was considered arguable.
Beyond the "is it understandable to the average python programmer" (never mind physicist who isn't a programmer), there was also the issue that it would suppress errors if someone passed a named argument that wasn't going to be utilized.

SimonBiggs: Ahh, I needed to make myself clearer. The portion of PyMedPhys that has its target audience as Medical Physicists and not programmers is all of the language that would end up appearing on https://docs.pymedphys.com/.

So docstrings on publicly exposed functions that have their text written onto the documentation site, need to be able to be understood by a medical physicist. The naming of internal functions, and the naming of internal APIs, I believe that's okay. What do you think?

That `**_` will go away however should I be explicit regarding your other comment.

pymedphys/_dicom/anonymise/__init__.py
SimonBiggs: Yeah, this was more of a swiss army knife approach to allow the API to be refactored without actually changing anything and then replaced later. But I agree, this is implicit rather than explicit, and verbosity here doesn't hurt.

pymedphys/_experimental/pseudonymisation/__init__.py
```diff
@@ -18,8 +18,8 @@
 from os.path import join as pjoin
 
 from pymedphys._dicom.anonymise import (
-    anonymise_directory,
-    anonymise_file,
+    _anonymise_directory,
```
sjswerdloff, line 21: this is a bit disturbing, but maybe because of my having both an OO background and a .NET background
a namespace that is private, i.e. pymedphys._dicom suggests that it should only be used by someone who is programming within the python package.
But a function that is private, i.e. _anonymise_directory, suggests that it should only be used within the module/file, or at most, within the same namespace (files that are children of the same directory), and in tests of that module (although there is a philosophy that says unit tests should be of public methods/functions and not private methods/functions... not that I adhere to that but I respect the idea that a refactoring should not break unit tests, and altering private methods should not be considered a break in contract or a change in requirements.

I'm willing to go along but I'd like to see anonymise_directory and anonymise_file made part of the internal (public to a private module) API of _dicom.anonymise

SimonBiggs: Yup, I agree. I don't like what's happened here either. `_anonymise_directory` should mean internally that it should be kept within that specific namespace and not used across the library.

> I'm willing to go along but I'd like to see anonymise_directory and anonymise_file made part of the internal (public to a private module) API of _dicom.anonymise

Honestly let's close this PR. It was a quick and flick attempt to try and refactor... but it was done poorly. I'll close this one. But did you want to do something similar so that we can get the renaming and API facade out of the way as a refactor that doesn't actually change any logic. And then we can dive into the actual changes in the other PR.

sjswerdloff: I got started in splitting things up as you asked (core.py, api.py, strategies.py, and __init__.py), but it's not trivial because of dependencies

SimonBiggs: I'm happy to give it a shot of you'd like? Probably won't get a chance till tonight though. Sorry I jumped in here. I was trying to be helpful... But I was probably more of a hindrance...

PR #1065, opened 2020-09-24
changed strategies from dict to immutables.Map

sjswerdloff:
made the dict private
created a private function to wrap the conversion to immutables.Map so that the additional key/value pairs used to store other control parameters/values can be addressed in the wrapping function.

### DIFF
--- a/pymedphys/_dicom/anonymise/strategy.py
+++ b/pymedphys/_dicom/anonymise/strategy.py
@@ -15,7 +15,7 @@
 import functools
 import logging
 
-from pymedphys._imports import pydicom
+from pymedphys._imports import immutables, pydicom
 
 from pymedphys._dicom.constants import PYMEDPHYS_ROOT_UID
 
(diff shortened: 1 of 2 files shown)


sjswerdloff:
@SimonBiggs this failed due to import magic.
I take it this means that a module level variable can not rely on a delayed import?

sjswerdloff:
While I like the idea of using immutables, between having started with module level variables for the strategy rather than a function call, using immutables means having to make a breaking change to the API for the strategy itself, never mind the functions that use a strategy as a named/explicit/optional argument.

sjswerdloff:
I'm not OK with this situation.
I'd rather close the PR and rethink it.

SimonBiggs:
> I take it this means that a module level variable can not rely on a delayed import?

Yup, that's the key. The trick is, that file needs to be able to be imported, and everything at module level run, without having `immutables` installed.

PR #1067, opened 2020-09-25
change behaviour in anonymise_directory in a backward compatible fashion

sjswerdloff:
accumulate list of anonymised files / paths
  catch errors in file anonymisation and log which files failed to anonymise
  if any errors took place, after all cleanup, raise the first error
  if no errors took place, return the list of anonymised files / paths

also... updated docstring for anonymise_file and anonymise_directory to include return type and description

This provides for a reasonably consistent set of return values across the three main anonymise_xxx()
which I will make use of in the proposed facade methods

### DIFF
--- a/pymedphys/_dicom/anonymise/api.py
+++ b/pymedphys/_dicom/anonymise/api.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 
+import logging
 import os
 import pprint
 from copy import deepcopy
@@ -223,6 +224,11 @@ def anonymise_file(
 
     identifying_keywords: ``list``, optional
         If left as None, the default values for/list of identifying keywords are used
+
+    Returns
+    -------
+    ``str``
+    The file path of the anonymised file
     """
     dicom_filepath = str(dicom_filepath)
 
@@ -334,12 +340,18 @@ def anonymise_directory(
 
     identifying_keywords: ``list``, optional
         If left as None, the default values for/list of identifying keywords are used
+
+    Returns
+    -------
+    ``list`` of anonymised file paths
     """
     dicom_dirpath = str(dicom_dirpath)
 
     dicom_filepaths = glob(dicom_dirpath + "/**/*.dcm", recursive=True)
     failing_filepaths = []
-    # errors = []
+    successful_filepaths = []
+    anon_filepaths = []
+    errors = []
 
     for dicom_filepath in dicom_filepaths:
         if output_dirpath is not None:
@@ -347,19 +359,26 @@ def anonymise_directory(
             output_filepath = os.path.join(output_dirpath, relative_path)
         else:
             output_filepath = None
-
-        anonymise_file(
-            dicom_filepath,
-            output_filepath=output_filepath,
-            delete_original_file=delete_original_files,
-            anonymise_filename=anonymise_filenames,
-            replace_values=replace_values,
-            keywords_to_leave_unchanged=keywords_to_leave_unchanged,
-            delete_private_tags=delete_private_tags,
-            delete_unknown_tags=delete_unknown_tags,
-            replacement_strategy=replacement_strategy,
-            identifying_keywords=identifying_keywords,
-        )
+        try:
+            dicom_anon_filepath = anonymise_file(
+                dicom_filepath,
+                output_filepath=output_filepath,
+                delete_original_file=delete_original_files,
+                anonymise_filename=anonymise_filenames,
+                replace_values=replace_values,
+                keywords_to_leave_unchanged=keywords_to_leave_unchanged,
+                delete_private_tags=delete_private_tags,
+                delete_unknown_tags=delete_unknown_tags,
+                replacement_strategy=replacement_strategy,
+                identifying_keywords=identifying_keywords,
+            )
+            successful_filepaths.append(dicom_filepath)
+            anon_filepaths.append(dicom_anon_filepath)
+        except BaseException as error:
+            errors.append(error)
+            failing_filepaths.append(dicom_filepath)
+            logging.warning("Unable to anonymise %s", dicom_filepath)
+            logging.warning(str(error))
 
     # Separate loop provides the ability to raise Exceptions from the
     # unsuccessful deletion of the original DICOM files while preventing
@@ -368,6 +387,11 @@ def anonymise_directory(
         for dicom_filepath in dicom_filepaths:
             if not dicom_filepath in failing_filepaths:
                 remove_file(dicom_filepath)
+
+    if len(errors) > 0:
+        logging.info("Succeeded in anonymising: \n%s", "\n".join(successful_filepaths))
+        raise errors[0]
+    return anon_filepaths
 
 
 def anonymise_cli(args):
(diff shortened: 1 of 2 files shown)


pymedphys/_dicom/anonymise/api.py
```diff
@@ -334,32 +340,45 @@ def anonymise_directory(
...
+                replace_values=replace_values,
+                keywords_to_leave_unchanged=keywords_to_leave_unchanged,
+                delete_private_tags=delete_private_tags,
+                delete_unknown_tags=delete_unknown_tags,
+                replacement_strategy=replacement_strategy,
+                identifying_keywords=identifying_keywords,
+            )
+            successful_filepaths.append(dicom_filepath)
+            anon_filepaths.append(dicom_anon_filepath)
+        except (AttributeError, LookupError, TypeError, OSError, ValueError) as error:
```
SimonBiggs, line 377: I'm okay with using the BaseException / Exception here, and then just overriding the pylint error. Since the error is going to be reraised later, I believe pylint's complaint isn't relevant to this situation.

sjswerdloff: I think it's OK to just have a handful of known errors, and if it's a different error, the error rips up through the stack.
That ends up being no different than what was there previously (which was no try/error at all).

pymedphys/_dicom/anonymise/api.py
```diff
@@ -334,32 +340,45 @@ def anonymise_directory(
...
+            errors.append(error)
+            failing_filepaths.append(dicom_filepath)
```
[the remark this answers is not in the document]
sjswerdloff: I'm not sure which code being added back in you are referring to.
It appeared that someone had intended to do something similar previously (the # errors = [] and putting the conditional removal of existing files in a separate block), but there was no try/error block.
I copied entire functions at a time while splitting up anonymise/ moving things out of the __init__.py
Its a fair amount of history trawling to figure out if that try/error was there previously and if so, when it got removed, presumably by accident.  Or perhaps it was removed on purpose because of changed requirements (maybe it's better to fail up fast instead of continuing).
I think having the original author review and discuss what the thinking is regarding fail fast vs. complete as much as possible would be very valuable here.

pymedphys/_dicom/anonymise/api.py
sjswerdloff: >I'm okay with using the BaseException / Exception here

Is that a request/requirement for me to revert and override the pylint error?
I'm fine either way, but if it's just "it would be OK", then I'd rather move forward.
If it's "this must change", I'll go ahead and do that.

pymedphys/_dicom/anonymise/api.py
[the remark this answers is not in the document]
SimonBiggs: The "blame" UI in GitHub is awesome for going through the history:
https://github.com/pymedphys/pymedphys/blame/87042b829b8ed33082396f99acfd51a06a99bbd7/pymedphys/_dicom/anonymise/__init__.py#L380-L410

Here are the pertinent commits:

* https://github.com/pymedphys/pymedphys/commit/b05327a28ec1024a7ff0c67388848af7b455c282
* https://github.com/pymedphys/pymedphys/commit/4f7eb9747a5f358a2b3872a991df006b31777e68

pymedphys/_dicom/anonymise/api.py
SimonBiggs: No, certainly not a must change. I just noticed you went from one to the other, and I wondered if it was to make pylint happy.

SimonBiggs (approved):

SimonBiggs:
LGTM, I'm happy for @Matthew-Jennings to chime in if he's keen, otherwise happy for this to merge.

sjswerdloff:
>
>The "blame" UI in GitHub is awesome for going through the history:
>https://github.com/pymedphys/pymedphys/blame/87042b829b8ed33082396f99acfd51a06a99bbd7/pymedphys/_dicom/anonymise/__init__.py#L380-L410
>
>Here are the pertinent commits:
>
> b05327a
> 4f7eb97
>

Hey, @Matthew-Jennings , This shows you tried "push on through" and then dropped back to "fail-fast".
Can you comment on what drove the return to "fail-fast"?
I saw the intent but didn't see the two commits that @SimonBiggs pointed out above, so I just went ahead an implemented against what I thought was the intent.

sjswerdloff:
Provided configuration for fail_fast and defaulting to True, so now this really is backward compatible.
Unless the programmer/user specifies they do not want fail_fast, the outcome will be what it was before this PR in the face of a failure.  Said programmer will have to dig through the target directory to find files that have been anonymised if there was a failure.
The primary purpose of the PR was to get the list of anonymised files back (assuming complete success).

sjswerdloff:
@SimonBiggs I don't unnderstand the (ubuntu, updates) failure.
It is the one set of tests that is not Required.
Should I go ahead and merge?

SimonBiggs:
> @SimonBiggs I don't understand the (ubuntu, updates) failure.
> It is the one set of tests that is not Required.
> Should I go ahead and merge?

Yup, updates failures are not required. That's a flag that should we now run "poetry update" tests will break. Its okay for that test to be failing for a merge to master (also not that "required" is not written next to the test).

It would need to be fixed before the next release. It essentially means a dependency change has broken something...

SimonBiggs:
> Provided configuration for fail_fast and defaulting to True, so now this really is backward compatible.
> Unless the programmer/user specifies they do not want fail_fast, the outcome will be what it was before this PR in the face of a failure. Said programmer will have to dig through the target directory to find files that have been anonymised if there was a failure.
> The primary purpose of the PR was to get the list of anonymised files back (assuming complete success).

Yup, I'm okay with this.

SimonBiggs:
LGTM

Matthew-Jennings:
> > The "blame" UI in GitHub is awesome for going through the history:
> > https://github.com/pymedphys/pymedphys/blame/87042b829b8ed33082396f99acfd51a06a99bbd7/pymedphys/_dicom/anonymise/__init__.py#L380-L410
> > Here are the pertinent commits:
> > [b05327a](https://github.com/pymedphys/pymedphys/commit/b05327a28ec1024a7ff0c67388848af7b455c282)
> > [4f7eb97](https://github.com/pymedphys/pymedphys/commit/4f7eb9747a5f358a2b3872a991df006b31777e68)
> 
> Hey, @Matthew-Jennings , This shows you tried "push on through" and then dropped back to "fail-fast".
> Can you comment on what drove the return to "fail-fast"?
> I saw the intent but didn't see the two commits that @SimonBiggs pointed out above, so I just went ahead and implemented against what I thought was the intent.

Hi @sjswerdloff!

I'm sorry for the very late reply. Figured it might still be worth answering this question in case it warrants discussion:

I don't remember for sure, but I seem to recall that I concluded that it would be too risky to rely on the user to check for individual failed anonymisations. E.g., consider the case of the CLI being run by a scheduled batch file. Thinking on it more now, the fast-fail solution might still be silent when run from a batch file, so even that point probably doesn't hold.

In any case, I think it's worth trying to design the code to be robust against blind trust by users (even if I didn't actually do that!)

Issue #580, opened 2019-11-19
Need to migrate away from pymssql


SimonBiggs:
pymssql is being discontinued. Need to migrate away from this dependency.

SimonBiggs:
See https://github.com/pymssql/pymssql/issues/668

SimonBiggs:
Likely going to use https://turbodbc.readthedocs.io/en/latest/pages/introduction.html

SimonBiggs:
It seems pymssql is being kept on life support!

https://github.com/pymssql/pymssql/issues/668#issuecomment-670995344

Looks like we can stick with it for now.

PR #1071, opened 2020-10-01
Bump CI to Python 3.8, make pymssql install for all python versions

SimonBiggs:
Now that pymssql has been getting some love, it now supports Python 3.8:

https://github.com/pymedphys/pymedphys/issues/580#issuecomment-688815486

As such, I am migrating things up to Python 3.8 initially on my development machine, I have also bumped the CI to run Python 3.8, and the documentation needs updating. I will make a note of what doc updates are needed in another issue.


pyproject.toml
```diff
@@ -58,7 +58,7 @@ pylibjpeg-libjpeg = {version = "*", optional = true}  # groups = ["user", "dev",
 immutables = {version = "*", optional = true}  # groups = ["user", "dev", "dicom"]
 
 cython = {version = "*", optional = true}  # groups = ["user", "dev"]
-pymssql = {version = "< 3.0.0", python = "< 3.8", optional = true}  # groups = ["user", "dev"]
+pymssql = {version = "*", optional = true}  # groups = ["user", "dev"]
```
sjswerdloff, line 61: doesn't this imply that if someone already has pymssql installed, any version will be considered OK?
wouldn't it be better to require at least the current version?
Or is it enough to assume that if someone is starting fresh (e.g. the CI) they will get the latest and not have a "too old" version of Python?

SimonBiggs: Best practices here is to have as loose as possible restrictions on a library package, only ever apply a version restriction if it is certainly known that a given version range won't work. For application deployment, pin all versions to a fixed exact version that has been tested.

Given we're on the library end, not the application end, it is better for us to be as open as possible so that we won't cause downstream "dependency hell".

SimonBiggs:
Hopefully CI passes now. The only issue with the bump to Python 3.8 seemed to be docs on Windows. For now I've left Windows at Python 3.7 and flagged the issue over at.

pyproject.toml
SimonBiggs: I should add here, I am choosing to pin Streamlit, as their API is still in development.

.pre-commit-config.yaml
```diff
@@ -26,4 +26,3 @@ repos:
     rev: 19.3b0
     hooks:
     - id: black
-      language_version: python3.7
```
sjswerdloff, line 29: is this loss of language_versin intentional?

pymedphys/tests/dicom/test_header_tweaks.py
```diff
@@ -45,7 +45,7 @@ def compare_dicom_cli(command, original, expected):
     pydicom.write_file(ORIGINAL_DICOM_FILENAME, original)
 
     try:
-        subprocess.check_call(command)
+        subprocess.check_call(" ".join(command), shell=True)
```
sjswerdloff, line 48: not that its a problem, but what motivated this?

pyproject.toml
sjswerdloff: With streamlit being as unstable as it is (not a dig, they are making great progress and have a very powerful model), and streamlit being what one might expect a physicist or developer in a clinical/research setting to want to use... this is going to get sticky.  It should be clearly documented for those using pymedphys as a development library/package

sjswerdloff (approved):
But I think the pinning of streamlit and what that implies should get documented

pymedphys/tests/dicom/test_header_tweaks.py
SimonBiggs: Only an issue for Windows Python. Subprocess on windows when not using "shell=True" will sometimes have issues with being confined within a virtual environment. In this case it wasn't using the poetry version of Python to call pymedphys.

.pre-commit-config.yaml
SimonBiggs: Yup, when swapping to Python 3.8 I don't want black to still be forced into using Python 3.7

pyproject.toml
SimonBiggs: > It should be clearly documented for those using pymedphys as a development library/package

Yup fair point. At the moment there actually isn't even any mention of the GUI or Streamlit in the docs. We essentially need a whole "GUI" section in the documentation which contains advice for users, advice for GUI creators, and some guidance on testing practices...

There's a lot to document in that space...

sjswerdloff: No time like the present to create a nearly empty section (in another Issue/PR).
I'll own the sub-section on pseudonymisation GUI...

SimonBiggs: > No time like the present to create a nearly empty section (in another Issue/PR).

Yup :) good call. Will do.

PR #1076, opened 2020-10-02
Demo nn

SimonBiggs:
@pchlap this one here is all experimental or prototyping code.

I was attempting to make a "mini-lung" dataset and train it with nnUNet. Anyway, given the location of all of this code the review from your end is more just a make sure that I'm not accidently merging in anything outside of prototyping or experimental.

So this can be a "tick and flick" review, if you so choose.

...If you'd like to have a sneak peak at my (failed) attempts and provide feedback that'd be awesome :) but not needed for the merge/review.


pchlap (approved):
Hi @SimonBiggs 
I've had a quick look through here. I've got some code which creates a UNet using pytorch which could be trained using the mini-dataset. I see you were trying something like that with tensorflow?

As for the nnUNet,I'd be keen to have a go at setting setting it up to train much more quickly for demo purposes. I'll try to have a closer look at that at some stage.

SimonBiggs:
Thanks @pchlap,

Yeah, the goal here is to have a partially toy example that is directly relevant to medical physics (auto segmentation of CT is ideal). The ideal would be something that users could be taught about, and then train, run inference, and validate, all while on a CPU bound laptop. If all of that could fit in ~3 hour workshop it would be an ideal practical teaching tool.

My thought is that making the image dimensions on the order of ~64x64 should be able to support this order of magnitude of training time. Potentially having the whole training dataset fit in the user's RAM at once, gaining all the speed-ups from having a large batch size (smoothing out the loss function and generalising across the whole training dataset within every iteration -> less time getting stuck in local minima). And as a result, it really might be feasible to have a project based, medical physics centric, machine learning workshop that is a reasonable amount of time.

PR #1081, opened 2020-10-08
Add baseline dicom dict back into repo

SimonBiggs:
So that internet connection isn't needed to run.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
     -   id: check-yaml
     -   id: check-toml
     -   id: check-added-large-files
-        args: ['--maxkb', '300']
+        args: ['--maxkb', '500']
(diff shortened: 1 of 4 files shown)


sjswerdloff (approved):

sjswerdloff:
In addition to places that have limited bandwidth or spotty internet, there are countries where internet access from facility systems (clinical or similar) are forbidden for security reasons (e.g. South Korea) or deeply frowned upon (e.g. Japan).

SimonBiggs:
> In addition to places that have limited bandwidth or spotty internet, there are countries where internet access from facility systems (clinical or similar) are forbidden for security reasons (e.g. South Korea) or deeply frowned upon (e.g. Japan).

Yup, very fair.

SimonBiggs:
Thanks for this @sjswerdloff :)

PR #1087, opened 2020-10-08
Demonstrate nbconvert breaking docs build

SimonBiggs:


poetry.lock
```diff
@@ -990,7 +990,7 @@ description = "Converting Jupyter Notebooks"
 name = "nbconvert"
 optional = true
 python-versions = ">=3.6"
-version = "6.0.6"
+version = "6.0.7"
```
SimonBiggs, line 993: This here is the cause of the break. It seems that going from version `6.0.6` --> `6.0.7` breaks the docs build.

SimonBiggs: Reported the issue at https://github.com/jupyter/nbconvert/issues/1442.

For now will leave nbconvert pinned to 6.0.6

PR #1100, opened 2020-10-21
streamlit based gui for pseudonymisation

sjswerdloff:
Puts pseudonymised files in a zip for download.
if there is more than 50 MByte worth of data, multiple zips are created.


pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+
+    """
+    filename = os.path.basename(zip_to_download)
+    with open(zip_to_download, "rb") as f:
+        zip_bytes = f.read()
+    link_to_zipbuffer_download(filename, zip_bytes)
+
+
+def build_pseudonymised_file_name(ds_input: pydicom.dataset.Dataset):
+    """[summary] Provide full path to the intended destination of the pseudonymised file
```
SimonBiggs, line 86: *Nitpick*

```suggestion
    """Provide full path to the intended destination of the pseudonymised file
```

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+
+    Parameters
+    ----------
+    ds_input : pydicom.dataset.Dataset
+        the non-pseudonymised DICOM data
+
+    Returns
+    -------
+    str
```
SimonBiggs, line 95: *Nitpick*

```suggestion
    temp_anon_filepath : str
```

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+        full path to the individual pseudonymised file, which can be read later
```
SimonBiggs, line 96: *Nitpick*

```suggestion
        Full path to the individual pseudonymised file, which can be read later
```

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+    link_to_zipbuffer_download(filename, zip_bytes)
+
+
+def build_pseudonymised_file_name(ds_input: pydicom.dataset.Dataset):
+    """[summary] Provide full path to the intended destination of the pseudonymised file
+
+    Parameters
+    ----------
+    ds_input : pydicom.dataset.Dataset
+        the non-pseudonymised DICOM data
```
SimonBiggs, line 91: *Nitpick*

```suggestion
        The non-pseudonymised DICOM data
```

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import base64
+import datetime
+import io
+import os
+import pathlib
+from os.path import join as pjoin
```
SimonBiggs, line 23: *Nitpick*

Are we able to just use pathlib instead of `os.path` also?

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+        for addition to a zip
+    """
+    pseudo_sop_instance_uid = pseudonymisation_api.pseudonymisation_dispatch["UI"](
+        ds_input.SOPInstanceUID
+    )
+
+    sop_class_uid: pydicom.dataelem.DataElement = ds_input.SOPClassUID
+
+    mode_prefix = DICOM_SOP_CLASS_NAMES_MODE_PREFIXES[
+        sop_class_uid.name  # pylint: disable = no-member
```
SimonBiggs, line 106: It's interesting that pylint couldn't find this. I wonder if there's something we can do to help pylint here. Maybe a type hint might bump it in the right direction...

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+
+def _zip_pseudo_fifty_mbytes(
+    file_buffer_list: list, zipfile_path: str, zip_bytes_io=None
+):
+    """Pseudonymises the contents of the file_buffer_list (list of DICOM files)
+    and places the pseudonymised files in to a zip.
+
+    Parameters
+    ----------
+    file_buffer_list: ``list`` of file buffers from streamlit file_uploader
```
SimonBiggs, line 121:
```suggestion
    file_buffer_list : list
        List of DICOM file buffers from streamlit file_uploader to pseudonymise
```

Portion after ":"  is reserved for type definition. See the parameters section of https://numpydoc.readthedocs.io/en/latest/format.html.

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+        DICOM files to pseudonymise
+    zipfile_path: ``str``
```
SimonBiggs, line 123:
```suggestion
    zipfile_path : str
```

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+        Location to write the zip file so that it can be downloaded.
+        Basename provides default name to use for downloading
+    zip_bytes_io: io.BytesIO (optional)
```
SimonBiggs, line 126:
```suggestion
    zip_bytes_io : io.BytesIO, optional
```

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+    The 50MByte limit is imposed by the href download link limit.
+
+    The compression used is DEFLATE, but the uncompressed contents are kept at just under 50 MBytes
+    If a single original uncompressed file is > 50MByte, and does not compress to under 50 MByte
+    that file will fail to be made available for download, and may cause the entire pseudonymisation
+    attempt to fail.
+
+    Parameters
+    ----------
+    file_buffer_list : ``list``
```
SimonBiggs, line 193:
```suggestion
    file_buffer_list : list
```

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+        DICOM files that were uploaded using streamlit.file_uploader
+
+    Returns
+    -------
+    ``list``
```
SimonBiggs, line 198:
```suggestion
    zip_path_list : list
```

See returns section of https://numpydoc.readthedocs.io/en/latest/format.html

SimonBiggs (commented):
Awesome @sjswerdloff. Thanks for this. Makes me want to get a simple installable binary out there sooner rather than later.

A couple of notes, is that this relies on the new streamlit version. On this branch when I run

```bash
cd pymedphys/_gui/streamlit
poetry run streamlit run pseudonymise.py
```

I get errors due to the streamlit version on PyMedPhys master. To update streamlit you'll need to run:

```
poetry add streamlit==0.69.2
```

Given I suspect you want to make sure this works into the future, I would highly recommend an end to end test. Here's an example end to end test using a DICOM file in the following repo:

https://github.com/pymedphys/streamlit-template/blob/master/cypress/integration/dicom.spec.js

That's for the old file uploader however. Some ideas for writing end to end tests for the new file uploader version can be found within the streamlit repo:

https://github.com/streamlit/streamlit/blob/develop/e2e/specs/st_file_uploader.spec.js

If you want to see the end to end test for the mudensity tool, see the following:

https://github.com/pymedphys/pymedphys/blob/master/pymedphys/tests/e2e/cypress/integration/streamlit/mudensity-compare.js

To run that test above in an interactive fashion, install yarn (https://classic.yarnpkg.com/en/docs/install/). Then run the streamlit server:

```
cd pymedphys/_gui/streamlit
poetry run streamlit run mudensity-compare.py
```

Then within another terminal do the following:

```bash
cd pymedphys/tests/e2e
yarn
yarn run cypress open
```

Then click "run all specs":

image

Interestingly, when watching those tests I have there for the mudensity tool, I haven't got a test for the DICOM file upload portion in there, and I can see that that part is actually broken with the new streamlit version. Anyway, that's a test for me to write when I get back to work, and something I'll need to do to update to the new streamlit version.

Let me know if you need anymore pointers. We (I?) really need to gather together some developer guidelines for some of these concepts here.

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+    if size != 0:
+        index_to_fifty_mbyte_increment.append(file_count)
+
+    return index_to_fifty_mbyte_increment
+
+
+st.set_option("deprecation.showfileUploaderEncoding", False)
+
+uploaded_file_buffer_list = st.file_uploader(
+    "Files to pseudonymise, refresh page after downloading zip(s)",
```
SimonBiggs, line 285: From a user interface stand point, I believe it is possible to trigger that download automatically once the file is ready, and then, once it has been downloaded a rerun of the streamlit application can be automatically triggered without needing to have the user refresh the page themselves.

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+        default filename to use
+    zip_bytes : bytes
+        the bytes from an IOBytes in which a ZipFile wrote its contents, or the bytes
+        read directly from a zip file on disk
+    """
+    b64 = base64.b64encode(zip_bytes).decode()
+    href = f"<a href=\"data:file/zip;base64,{b64}\" download='{filename}'>\
+        Click to download {filename}\
+    </a>"
+    st.sidebar.markdown(href, unsafe_allow_html=True)
```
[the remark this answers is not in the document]
sjswerdloff: >How amenable are you to wrapping the "download a file" and "download a directory as a zip" code into a streamlit custom component?
I'm amenable to having it done and having that contributed.  I'm not excited about the prospect of doing that work myself.  Too many detours for me...

sjswerdloff:
accepted/committed directly actionable suggestions, will follow up on the remainder in a separate commit or set of commits.
I'll call out when I think I've addressed **all** suggestions.

pymedphys/_gui/streamlit/pseudonymise.py
sjswerdloff: These are both things I wanted to do, but didn't find the programmable refresh/rerun (that you contributed to streamlit) until after creating the PR.  So, yes, I want to address this, I think it would provide a significant improvement to the UX.
I'm not sure how one triggers the download automatically.

pymedphys/_gui/streamlit/pseudonymise.py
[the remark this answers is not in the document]
SimonBiggs: > I'm amenable to having it done and having that contributed. I'm not excited about the prospect of doing that work myself. Too many detours for me...

Yup, all good. I'll see if I can jump on that one evening soon. Not as a part of this PR though.

pymedphys/_gui/streamlit/pseudonymise.py
SimonBiggs: Regarding the automatic download, does the following do the trick?

https://stackoverflow.com/a/156715

Worth also putting in display:None as in the following comment:

https://stackoverflow.com/questions/156686/how-to-start-automatic-download-of-a-file-in-internet-explorer/156715#comment27313065_156715

Given its always going to be a zip here, that should do the trick. Haven't tested it though.

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -118,12 +118,13 @@ def _zip_pseudo_fifty_mbytes(
 
     Parameters
     ----------
-    file_buffer_list: ``list`` of file buffers from streamlit file_uploader
+    file_buffer_list : list
+        List of DICOM file buffers from streamlit file_uploader to pseudonymise
         DICOM files to pseudonymise
```
SimonBiggs, line 123:
```suggestion
```

pymedphys/_gui/streamlit/pseudonymise.py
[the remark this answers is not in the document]
sjswerdloff: >Yup, all good. I'll see if I can jump on that one evening soon. Not as a part of this PR though.
Much appreciated.

pymedphys/_gui/streamlit/pseudonymise.py
SimonBiggs: Better than the IFrame option actually, would be to open the zip URL in a new tab using `window.open`

https://www.codexworld.com/how-to/open-url-in-new-browser-window-tab-javascript/

That way you can refresh the page and it won't be a problem if the zip hasn't finished its "start download" process as it will begin its downloading in a new tab (which should then promptly disappear).

SimonBiggs: If you'd like, I can handle all that in the custom component PR, separate to this PR.

sjswerdloff: > If you'd like, I can handle all that in the custom component PR, separate to this PR.
Yes, please, with sugar on top :-)

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,297 @@
...
+# HACK This only works when we've installed streamlit with pipenv, so the
+# permissions during install are the same as the running process
+STREAMLIT_STATIC_PATH = pathlib.Path(st.__path__[0]) / "static"
+if not STREAMLIT_STATIC_PATH.is_dir():
+    STREAMLIT_STATIC_PATH.mkdir()
+# We create a downloads directory within the streamlit static asset directory
+# and we write output files to it
+DOWNLOADS_PATH = STREAMLIT_STATIC_PATH / "downloads"
+if not DOWNLOADS_PATH.is_dir():
+    DOWNLOADS_PATH.mkdir()
```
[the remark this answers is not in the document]
sjswerdloff: done. consolidated STREAMLIT_STATIC_PATH as well

sjswerdloff:
@SimonBiggs 
>Let me know if you need anymore pointers.

We have sample DICOM data (in zenodo) and I could use some of that (dummy-ct-and-struct.zip is perfect, uncompressed it is more than 50 MB, so it will result in multiple zips).
But cypress doesn't seem very good with downloading and reading large files.  I think I somehow started with v4.10 and I upgraded to 5.4.0 (your test runs the same in both versions of cypress).
The examples show binary download.  That doesn't work.
Modifying the request to use base64 does work, some times (intermittently... absolutely 100% same code), 
and when it does, one then has to decode (e.g. b64 -d < my_long_path/my.zip.b64 > my_long_path/my.zip) using exec().
And then unzip using exec(unzip -d my_long_path my_long_path/my.zip)
The download is very slow compared to just using the browser.
The readFile(downloadedFilename, 'binary', { timeout: 15000 })
 function times out.  So... fifteen seconds isn't enough time to read a 40 MByte file.  From a Solid State Drive that pumps data at 500 MByte/second (measured...).

I can look at doing this with a single DICOM file to avoid the size/scale/timeout issues.
But the cypress testing framework has been a very unsatisfying experience in that the example I used didn't work as is for something very straight forward, and it's performance is abysmal for the real world test case that I am starting with.

SimonBiggs:
> Modifying the request to use base64 does work

For large files I believe that writing >50MB worth of base64 HTML on the webpage will hurt the browser in most cases I believe, so I suspect it may not be cypress related.

> But cypress doesn't seem very good with downloading and reading large files. I think I somehow started with v4.10 and I upgraded to 5.4.0 (your test runs the same in both versions of cypress).

Let me have a look and see if I can work something out. (Won't get a chance until tomorrow night).

SimonBiggs:
> But cypress doesn't seem very good with downloading and reading large files.

Just an idea, might be able to change the download directory to something useful using the following:

https://docs.cypress.io/api/plugins/browser-launch-api.html#Change-download-directory

May also need to change the `prompt_for_download` setting to false:

```javascript
options.preferences.default['download'] = { 
  default_directory: downloadDirectory,
  prompt_for_download: false
}
```

Then, from there, can call a CLI command that checks to see if the file has been appropriately anonymised:

https://docs.cypress.io/api/commands/exec.html#Usage

SimonBiggs (approved):
Given Cypress is causing a decent amount of pain, I'd be happy to set up the Cypress testing in the same PR as when I create the download zip component.

Feel free to merge this one in when you're ready. I'll address creating the streamlit component and writing the Cypress tests :).

sjswerdloff:
I pushed up the first/rough draft of the cypress based test before I saw your comment.
It includes commented out code for downloading from zenodo and unzipping.
It assumes two of the DICOM files will get copied to fixture rather than be in download.
But even then a number of method/function calls in the example code fail ("upload is not a function")

I'll revert on Monday if the checks fail and then squash/merge, and leave the cypress e2e test for the future PR.
Life on the bleeding edge...

pymedphys/_gui/streamlit/pseudonymise.py
```diff
@@ -0,0 +1,284 @@
...
+
+from pymedphys._dicom.anonymise import anonymise_dataset
+from pymedphys._dicom.constants.core import DICOM_SOP_CLASS_NAMES_MODE_PREFIXES
+from pymedphys._dicom.utilities import remove_file
+from pymedphys.experimental import pseudonymisation as pseudonymisation_api
+
+# We create a downloads directory within the streamlit static asset directory
+# and we write output files to it, similar to approach in streamlit forum
+DOWNLOADS_PATH = pathlib.Path(st.__path__[0]).joinpath("static", "downloads")
+DOWNLOADS_PATH.mkdir(parents=True, exist_ok=True)
```
SimonBiggs, line 37: I believe this here has actually combined two approaches. When going the b64 encoding route, don't actually need to store the data into Streamlit's static directory. In fact, don't actually need to store the data to disk at all.

Instead, can just open an in memory file object using `io.BytesIO`:

```python
in_memory_temp_file = io.BytesIO()
pydicom.dcmwrite(in_memory_temp_file, ds_input)
myzip.write(in_memory_temp_file, anon_filename)
```

SimonBiggs (changes requested):
Only change needed is to swap to using in memory files for the temp DICOM saves instead of using the potentially write protected streamlit static directory.

SimonBiggs:
@sjswerdloff so that I can edit this, I am going to merge into a pymedphys branch, `add-gui-for-pseudonymisation`.

PR #1107, opened 2020-10-29
Poetry and docs tooling

SimonBiggs:
Key additions here are the following two development CLI commands:

```
pymedphys dev propagate
pymedphys dev docs --live
```

Propagate will update the package version number, update the extras within pyproject.toml and create a requirements.txt.

The live option boots a docs live reload server that will update whenever the documentation is edited.


pymedphys/_dev/propagate.py
```diff
@@ -0,0 +1,104 @@
...
+
+    version_contents = black.format_str(version_contents, mode=black.FileMode())
+
+    with open(VERSION_PATH, "w") as f:
+        f.write(version_contents)
+
+
+def propagate_requirements():
+    subprocess.check_call(
+        "poetry export -E dev -f requirements.txt --output requirements.txt", shell=True
```
sjswerdloff, line 72: dev is probably the most complete, but is requirements.txt then going to be based on [dev] ?

SimonBiggs: Yeah, that'll mean that the requirements.txt file produced will be equivalent to a dev install. I can see why it may be better to opt for user instead...

pymedphys/_version.py
```diff
@@ -1,4 +1,5 @@
-# pylint: disable=invalid-name, missing-docstring
+# DO NOT EDIT
+# This file is autogenerated by `pymedphys dev propagate`
 
 version_info = [0, 34, 0, "dev2"]
-__version__ = "0.34.0dev2"
+__version__ = "0.34.0.dev2"
```
sjswerdloff, line 5: is this proper semver ?  I thought that a dash in the 3rd dot-field for indicating pre-release


```
Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows:

Identifiers consisting of only digits are compared numerically.

Identifiers with letters or hyphens are compared lexically in ASCII sort order.

Numeric identifiers always have lower precedence than non-numeric identifiers.

A larger set of pre-release fields has a higher precedence than a smaller set, if all of the preceding identifiers are equal.

Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.
```

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pymedphys"
-version = "0.34.0dev2"
+version = "0.34.0.dev2"
```
sjswerdloff, line 3: is this proper semver?

pymedphys/_version.py
SimonBiggs: Yup, fair call. Here's an example in the wild:

https://github.com/python-semver/python-semver/blob/ddf2b304ebc0bf4a8083a13372c908c8dc6b8dff/semver.py#L28

.pre-commit-config.yaml
```diff
@@ -11,7 +11,6 @@ repos:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
         exclude: ^.*demographic\.000000$
-    -   id: requirements-txt-fixer
```
sjswerdloff, line 14: intentional deletion of requirements-txt-fixed?

sjswerdloff (changes requested):
confirm semver for pre-release.
I'm pretty sure that the norm is to use a dash in the third dot/field and only put in an additional dot if there are sub-increments of pre-release.

.pre-commit-config.yaml
SimonBiggs: Yup, poetry uses its own format, that fixer doesn't like it. Figured I'm happy to go with what poetry spits out by default.

pymedphys/_dev/propagate.py
sjswerdloff: I think user would be better.
If you are a pymedphys developer, you are going to use poetry (you pretty much have to) and not use the requirements.txt
But if you are a user, you'd really be a lot happier to just use the requirements.txt (and presumably pip would use that to address installing dependencies while it's installing pymedphys?)

SimonBiggs: Yup, that makes sense. To install with pip a setup.py is also needed. I'll use add a setup.py creation to this propagate CLI also:

https://github.com/python-poetry/poetry/issues/761#issuecomment-521124268

SimonBiggs: In the end the setup.py produced by dephell was nasty. So I extracted poetry's version instead.

setup.py
```diff
@@ -0,0 +1,238 @@
...
+    "tests": [
+        "pytest",
+        "pytest-sugar",
+        "hypothesis",
+        "psutil",
+        "pylint",
+        "pytest-pylint",
+    ],
+    "user": [
+        "tqdm",
```
sjswerdloff, line 190: As an fyi, I find I have to do
sudo apt-get install tqdm
otherwise running the tests fail for the lack of tqdm

sjswerdloff (approved):

setup.py
SimonBiggs: Cheers for the heads up. 

I suspect that may be due to Python path issues. I actually make sure that absolutely nothing is installed within the root Python to make it clear when my python path is being mangled.

PR #1108, opened 2020-10-29
Include GUI placeholder documentation

SimonBiggs:
I'm merging this one into the poetry doc and tooling branch, as it built upon that branch.


sjswerdloff (approved):
Nice.  Now it's on me to put something in to the area neatly set aside for pseudonymisation gui

PR #1116, opened 2020-11-01
remove kiwisolver version 1.3.0 option

SimonBiggs:


sjswerdloff:
why?  presumably broken API or bug in kiwisolver 1.3.0 ?

sjswerdloff (approved):
Do you expect 1.3.1 to fix things?
If an API issue, this is only a temporary fix, and < 1.3.0 is necessary.
If it's a bug, it's reasonable to hope, especially if the maintainers of kiwisolve have identified that they are going to fix this.

SimonBiggs:
> Do you expect 1.3.1 to fix things?

Yup, I do. 1.3.0 was "yanked". See https://pypi.org/project/kiwisolver/1.3.0/. Only 1.3.0 needed to be blacklisted.

PR #1118, opened 2020-11-02
[0] MU density GUI refactor only

SimonBiggs:
No changes are undergone in the PR. This is a refactor only PR.


pymedphys/_gui/streamlit/mudensity/_config.py
```diff
@@ -0,0 +1,180 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
```
sjswerdloff, line 1: copyright?

sjswerdloff (approved):

pymedphys/_gui/streamlit/mudensity/_config.py
SimonBiggs: Yup, that was missed in copying it across. Fixed it up, in the PR stacked on top of this one.

...hopefully all these "PR" stackings was actually helpful and not super confusing...
keen for you feedback...

PR #1111, opened 2020-10-30
[1] Make trf compare in mudensity gui not need Mosaiq access

SimonBiggs:
@sjswerdloff (as always no urgency to review during designated down time :) )

This is to make it so that TRF comparison can be done without Mosaiq access.

PR #1120, opened 2020-11-02
[3] Request users for more data regarding RT DICOM --> Delivery cases

SimonBiggs:
[EDIT]

Previously this PR relaxed the DICOM constraint in RT DICOM --> Delivery. Now however it has changed to linking to an issue and requesting more data from users.

---

Re query over at https://groups.google.com/d/msgid/pymedphys/CADSx4S2NCAzB9TLKht2iFzh_myFpKZFOq%3Dk%2Bo6EAsTzP4W4XzA%40mail.gmail.com?utm_medium=email&utm_source=footer

> Dear all,
>
> I also tried this version to compare trf and rtplan and this version works great!
>
> I also like to run this version for MR-Linac Unity files comparison, but then I got an error.
>
> Is it possible to fix this?
>
> 
> image
>
> I added some trf and rtplan files to test it.
>
> Best Regards,
> Martijn Kusters

The following files were provided for testing:
[unity-trf-and-dicom.zip](https://github.com/pymedphys/pymedphys/files/5474408/unity-trf-and-dicom.zip)

It would be great to create a test including these files. I'll add that as an issue and schedule it for a separate PR.


SimonBiggs:
My response on the mailing list:

------

Hi Martijn,

Glad to hear it works great :). Regarding those MRI Unity files I have loosened the restriction on MLCX + ASYMY vs MLCY + ASYMX, but there's going to need to be quite the swathe of "sanity checks" to make sure something weird hasn't resulted due to that.

Would you be able to install the following:

```
pip install pymedphys[user]==0.34.0-dev8
```

And then have a play around and make sure that it really is doing what it should be. Important things to check would be an off axis rectangular field, paying particular attention to the coordinates being reported by the software. Keen for you to put it through its paces and report back. Also, if there are a few simple RT plan DICOM files with short deliveries and their corresponding TRFs over a range of cases I would highly appreciate being able to include that data within the pymedphys testing suite.

Cheers,
Simon

pymedphys/_dicom/delivery/core.py
```diff
@@ -161,20 +163,28 @@ def _from_dicom_beam(cls, beam, meterset):
...
+                    [
+                        -np.array(mlc[0:num_leaves][::-1]),
+                        np.array(mlc[num_leaves::][::-1]),
+                    ][::-1]
+                ).T
+                for mlc in dicom_mlcs
+            ]
+        )
+
+        # Nasty hack, patching the fact that the leaf banks are swapped
```
sjswerdloff, line 181: this is high risk.
where is the unit test for this?

pymedphys/_dicom/rtplan/core.py
```diff
@@ -212,7 +212,7 @@ def get_leaf_jaw_positions_for_type(
         matching_type = [
             item
             for item in sequence
-            if item.RTBeamLimitingDeviceType == rt_beam_limiting_device_type
+            if item.RTBeamLimitingDeviceType.startswith(rt_beam_limiting_device_type)
```
sjswerdloff, line 215: I have concerns about using strings that are not a valid defined type for the rt_beam_limiting_device_type for partial matching.
Better to pass in a list of acceptable types to be matched.
You already have something that raises an error if the list of matching_types is != 1.
so something like rt_beam_limiting_device_type_list in item.RTBeamLimitingDeviceType.

pymedphys/_dicom/delivery/core.py
SimonBiggs: Yup, I agree. I need to make a test based upon the following data:

[unity-trf-and-dicom.zip](https://github.com/pymedphys/pymedphys/files/5474408/unity-trf-and-dicom.zip)

pymedphys/_dicom/delivery/core.py
```diff
@@ -161,20 +163,28 @@ def _from_dicom_beam(cls, beam, meterset):
...
+        # around for MLCY vs MLCX
+        if "MLCY" in rt_beam_limiting_device_types:
+            mlcs[:, :, [0, 1]] = mlcs[:, :, [1, 0]]
 
         dicom_jaw = rtplan.get_leaf_jaw_positions_for_type(
-            beam_limiting_device_position_sequences, "ASYMY"
+            beam_limiting_device_position_sequences, "ASYM"
```
sjswerdloff, line 187: I have concerns about using a value that isn't a defined or enumerated CS in the DICOM Standard .

pymedphys/_dicom/rtplan/core.py
SimonBiggs: Yeah, I was hesitant about this. You concern here cements that.

sjswerdloff (changes requested):
At least provide conversation that indicates where the unit test is for what identified as high risk code.
Prefer finding alternative to using a value that isn't a DICOM defined or enumerated value for partial matching.  Maybe just change the name of the parameter as a worst case approach so that one doesn't infer that it is actually a valid Beam Limiting Device type.

pymedphys/_dicom/delivery/core.py
SimonBiggs: I'd hoped to hear back with some results from Martijn's testing, and potentially some simple DICOM data and TRF data that would be ideal for testing with.

See email above:

> Keen for you to put it through its paces and report back. Also, if there are a few simple RT plan DICOM files with short deliveries and their corresponding TRFs over a range of cases I would highly appreciate being able to include that data within the pymedphys testing suite.

Unfortunately haven't heard back yet. My "Delivery Object" coordinate system uses as its reference the TRF file format. So in this case, essentially the key is that it is verified that the delivery object representation created from the DICOM file is equivalent to that produced by the TRF file.

pymedphys/_dicom/delivery/core.py
SimonBiggs: Yup, this is hand in hand with the comment above:

https://github.com/pymedphys/pymedphys/pull/1120#discussion_r517044617

Will fix this.

pymedphys/_dicom/delivery/core.py
SimonBiggs: That's all good and fine. I wouldn't have it any other way. I was planning tomorrow at work to build a test and mark it slow based on the data above. If we get more refined data I can speed up the test by pulling something in.

SimonBiggs: Any further thoughts on the following PR?

https://github.com/pymedphys/pymedphys/pull/1111

pymedphys/_dicom/delivery/core.py
sjswerdloff: Issue specified in https://github.com/pymedphys/pymedphys/issues/1139

pymedphys/_dicom/rtplan/core.py
sjswerdloff: issue specified in https://github.com/pymedphys/pymedphys/issues/1139

SimonBiggs:
My plan here is to actually remove this change for now. Given this issue here of making DICOM --> delivery more flexible is primarily (only) for other sites and not our own, I need to alot time for this one outside of work.

pymedphys/_dicom/delivery/core.py
```diff
@@ -125,11 +125,22 @@ def _from_dicom_beam(cls, beam, meterset):
             item.RTBeamLimitingDeviceType for item in beam_limiting_device_sequence
         }
 
-        if rt_beam_limiting_device_types != set(["MLCX", "ASYMY"]):
+        supported_configurations = [{"MLCX", "ASYMY"}]
+
+        if not rt_beam_limiting_device_types in supported_configurations:
             raise ValueError(
                 "Currently only DICOM files that contain "
                 "exactly MLCX and ASYMY are supported. "
```
sjswerdloff, line 133: prefer pretty printing the supported_configurations

```
for bld_config in supported_configurations:
    print(" and ".join(bld_config))
```

or something equivalent

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pymedphys"
-version = "0.34.0-dev6"
+version = "0.34.0-dev8"
```
sjswerdloff, line 3: are we past dev8 now?
up to dev10 maybe?

setup.py
```diff
@@ -223,7 +224,7 @@
 
 setup_kwargs = {
     "name": "pymedphys",
-    "version": "0.34.0.dev6",
+    "version": "0.34.0.dev8",
```
sjswerdloff, line 227: dev10?

sjswerdloff (changes requested):
we are past dev8, minimum dev10 required.
please also consider iterating on supported_configurations for error message

pyproject.toml
SimonBiggs: It was changing up the stack... so once this stack merges it'll say dev9, but yes, should a new dev release happen it'd need to be dev10

pymedphys/_dicom/delivery/core.py
SimonBiggs: Good call. DRY > WET :)

SimonBiggs:
Alright :)

I hope I have now addressed your concerns here :)

sjswerdloff:
concerns addressed (with an approach that looks like it could be re-used in the future).
I'll approve.
The tests are broken because of the content of the regex in the test, but GitHub will enforce that part.

sjswerdloff (approved):

SimonBiggs:
Thanks @sjswerdloff for these :). I asked a lot of you reviewing all of these. I really appreciate it.

PR #1124, opened 2020-11-03
[4-10] Approved GUI work -- merge to master awaiting [3]

SimonBiggs:
This PR contains the sum of the stacked PRs of what was previously PRs labelled 4-10.

Work remaining upward from here:

* work remaining re improvements for PR https://github.com/pymedphys/pymedphys/pull/1120.
* merge in the remaining applications that are used within our centre
* create a config application
* make mu density default to the file upload case instead of the demo case
* make the cypress test for mu density first use the config application to change it into "demo" mode
* nest the pymedphys package under a lib directory so that pymedphys can be called from the root of the repo
* add in remaining configuration so that the multi-app runs on http://share.streamlit.io/


pymedphys/_gui/streamlit/mudensity/main.py
```diff
@@ -18,36 +18,31 @@
 # pylint: disable = too-many-lines, redefined-outer-name
```
sjswerdloff, line 18: can you get rid of this pylint disablement now (or has this already happened in the future?)

sjswerdloff (approved):

pymedphys/_gui/streamlit/mudensity/main.py
SimonBiggs: Yup, was able to get rid of it :). Happened in the PR on top of this one:

https://github.com/pymedphys/pymedphys/pull/1125/files

PR #1125, opened 2020-11-03
[5] Remove pylint ignores

SimonBiggs:


pymedphys/_gui/streamlit/mudensity/main.py
```diff
@@ -692,11 +691,13 @@ def main():
...
     if advanced_mode:
-        """
-        ### Image record
+        st.write(
+            """
+            ### Image record
 
-        Path to save the image of the results for posterity
-        """
+            Path to save the image of the results for posterity
```
sjswerdloff, line 698: This wording is not English as a Second language friendly.
If we internationalize and localize, it's fine.
Not critical but something to keep in mind

sjswerdloff (approved):

pymedphys/_gui/streamlit/mudensity/main.py
SimonBiggs: Good call. I hadn't thought of that.

PR #1126, opened 2020-11-03
[6] Multi app with index

SimonBiggs:
Alright! Finally, a multi app approach that *seems* to be robust.

...so now apps can be directly linked to with http://localhost:8502/?app=mudensity, there is an index, and apps can navigate back to the index, url updates, all that jazz...

Test failures fixed downstream in https://github.com/pymedphys/pymedphys/pull/1130


pymedphys/_streamlit/state.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+    else:
+        session_infos = (
+            Server.get_current()._session_info_by_id.values()  # pylint: disable = protected-access
+        )
+
+    for session_info in session_infos:
+        s = session_info.session
+        if (
+            # Streamlit < 0.54.0
```
sjswerdloff, line 90: won't a number of streamlit features not exist or fail at these earlier versions?
Are you sure you don't want to put a lower bound on streamlit and just not support below that?

sjswerdloff (approved):
It's not clear if this is a PR still in play or was superceded.
If not still in play, please close/obsolete.
Also... please consider whether older versions of streamlit should be supported or not, and consider stripping out code for unsupported older versions.

pymedphys/_streamlit/state.py
SimonBiggs: Key here is actually that this code here is a "vendor":

https://github.com/pymedphys/pymedphys/blob/1e95d1c566b71411e2c8fe841129bab445feac8d/pymedphys/_streamlit/state.py#L1

From https://gist.github.com/tvst/036da038ab3e999a64497f42de966a92

I'm erring on the side of sticking close to the original here.

SimonBiggs:
> It's not clear if this is a PR still in play or was superceded.
> If not still in play, please close/obsolete.

These are all still in play. They're actually stacked ontop of each other instead into master:

image

That way the changelogs are incremental.

PR #1128, opened 2020-11-03
[7] Merge master (containing psuedonymise gui)

SimonBiggs:
This is just merging master. Nothing else major happening here. Done in it's own PR to make the change log of surrounding PRs easier to read.

Test failures fixed downstream in https://github.com/pymedphys/pymedphys/pull/1130


pymedphys/tests/e2e/cypress/integration/streamlit/pseudonymisation.spec.js
```diff
@@ -0,0 +1,256 @@
+/**
+ * @license
```
sjswerdloff, line 2: We should open an issue for cleaning out the cruft and updating the copyright.
While this started from a file from streamlit, it's got precious little left of that.
I will leave the license where it is, but above it indicate that it started from streamlit but was heavily modified, and include both Simon and Stuart (you and me if 3rd person sounds weird)

sjswerdloff (approved):

PR #1143, opened 2020-11-05
[Merge?] Wlutz GUI work

SimonBiggs:
Add an experimental iView database explorer
And an initial Winston Lutz utility

---

Both of these added applications are classified under the "experimental" category within the index. There are not any e2e tests of these tools thus far. In fact, they don't yet even work.

Nevertheless, I don't want this PR to grow too huge. I was hoping to do an incremental merge in its current "draft"/"experimental" state. What are your thoughts @sjswerdloff?

---

Alright, so to test out this PR locally, first download the following zip:

https://zenodo.org/record/4263606/files/demo-iviewdb.zip?download=1

Then extract the contents of that zip to `~/demo-iviewdb`.

Then, make a config.toml file at `~/.pymedphys/config.toml` that looks like the following:

```toml
version = 1

[[site]]
name = "default-site-name"

    [site.export-directories]
    iviewdb = '~/demo-iviewdb'
```

Then, within the root of the pymedphys repo run:

```
poetry run pymedphys gui
```

Then, navigate ones browser to http://localhost:8501/?app=wlutz for the wlutz app, or http://localhost:8501/?app=iviewdb for the iView database explorer.


lib/pymedphys/_streamlit/apps/iviewdb.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def main():
+    st.title("iView Database Explorer")
+
+    _, database_directory = misc.get_site_and_directory("Database Site", "iviewdb")
+
+    database_paths = {
+        path.stem.lower(): path
+        for path in database_directory.glob("*.dbf")
```
pchlap, line 36: This returned an empty dict for me (unix based, Mac OS) because the extension chars are in caps. When I changed it to *.DBF it worked.

lib/pymedphys/_streamlit/apps/iviewdb.py
```diff
@@ -0,0 +1,82 @@
...
+    field_names = dbf_record.field_names
+
+    selected_fields = st.multiselect(
+        "Field names to view", field_names, default=field_names
+    )
+
+    if selected_fields:
+        full_pandas_table = pd.DataFrame(iter(dbf_record))
+        st.write(full_pandas_table[selected_fields])
+
```
pchlap, line 59: Getting KeyError here for user and station table, not immediately clear why, but could be simply because they are empty?

lib/pymedphys/_streamlit/apps/wlutz/_dbf.py
```diff
@@ -0,0 +1,136 @@
...
+
+    for database_key, merge_key in [
+        ("port", "PORT_DBID"),
+        ("trtmnt", "TRT_DBID"),
+        ("patient", "PAT_DBID"),
+    ]:
+        dbf_to_be_merged = load_dbf(database_directory, refresh_cache, database_key)
+        merged = merged.merge(dbf_to_be_merged, left_on=merge_key, right_on=merge_key)
+
+    timestamps_string = (
```
pchlap, line 76: I have adopted f-strings (which I discovered thanks to pymedphys). Is this something we want to use consistently throughout?

lib/pymedphys/_streamlit/apps/wlutz/_dbf.py
```diff
@@ -0,0 +1,136 @@
...
+    return [dbf_to_pandas_without_cache(path)]
+
+
+def dbf_to_pandas(path, refresh_cache=False):
+    result = dbf_to_pandas_with_cache(path)
+
+    if refresh_cache:
+        result[0] = dbf_to_pandas_without_cache(path)
+
+    return result[0]
```
pchlap, line 119: I assume the use of the list here is to make it compatible with st.cache? This is a little confusing first time looking at it but makes sense if that an st.cache requirement.

lib/pymedphys/_streamlit/apps/wlutz/_frames.py
```diff
@@ -0,0 +1,137 @@
...
+@st.cache()
+def calc_xml_filepaths(table):
+    return (
+        "patient_"
+        + table["patient_id"].astype("str")
+        + "/MV_IMAGES/img_"
+        + table["DICOM_UID"].astype("str")
+        + "/_Frame.xml"
+    )
+
```
pchlap, line 99: Again, I would personally favour an f-string..

pchlap (commented):
Good stuff @SimonBiggs I have given the iView DB browser a go and provided a few comments in the code. For the Wlutz part I have read through the code and provided a couple of high level comments. But I don't know enough about it to have more constructive comments unfortunately.

lib/pymedphys/_streamlit/apps/wlutz/_dbf.py
SimonBiggs: I would also certainly favour an f string :). In this case though the variables aren't actually strings, but instead `pandas.Series` objects. So, essentially an array of strings, so here I am using that pandas `+` to do element-by-element concatenation.

lib/pymedphys/_streamlit/apps/wlutz/_dbf.py
SimonBiggs: Yeah, it is confusing. I should think of a clearer way to do it. Essentially I wanted to create a mutable output so that I can mutate the cache. This provides the functionality for the user to "re-read" the database. When a streamlit function is cached it only runs once and then stores the results. This is an issue though if the database updates. I may just have to remove the cache altogether in the end, or cache a different part of the pipeline instead. Having the database not update and be cached may just be a bad idea in and of itself.

I'll leave this here for now, but will certainly consider removing the cache altogether as I iterate.

lib/pymedphys/_streamlit/apps/iviewdb.py
SimonBiggs: Yeah, me too. I didn't actually need to use those tables, so for now I just left it as is.

lib/pymedphys/_streamlit/apps/iviewdb.py
SimonBiggs: Ahh, yes, good call. Thanks. :).

SimonBiggs:
> Good stuff @SimonBiggs I have given the iView DB browser a go and provided a few comments in the code. For the Wlutz part I have read through the code and provided a couple of high level comments. But I don't know enough about it to have more constructive comments unfortunately.

All good, was helpful :). Thanks for casting your eyes over it :).

lib/pymedphys/_streamlit/apps/wlutz/_dbf.py
pchlap: Ah I see, cool! :)

lib/pymedphys/_streamlit/apps/iviewdb.py
```diff
@@ -0,0 +1,82 @@
...
+
+def main():
+    st.title("iView Database Explorer")
+
+    _, database_directory = misc.get_site_and_directory("Database Site", "iviewdb")
+
+    database_paths = {
+        path.stem.lower(): path
+        for path in database_directory.glob("*.DBF")
```
sjswerdloff, line 36: can this be done as a comprehension?
having the if statement on a second line makes it a bit awkward to grasp.

SimonBiggs: This is a comprehension.

lib/pymedphys/_streamlit/apps/mudensity/_icom.py
```diff
@@ -41,6 +41,8 @@ def load_icom_streams(icom_paths):
     return icom_streams
 
 
+# TODO: Split this up to search by site
```
sjswerdloff, line 44: if it warrants a TODO, it warrants a docstring

lib/pymedphys/_streamlit/apps/wlutz/_dbf.py
```diff
@@ -0,0 +1,136 @@
...
+    },
+    "patient": {
+        "filename": "PATIENT.dbf",
+        "columns_to_keep": ["DBID", "ID", "LAST_NAME", "FIRST_NAME"],
+        "column_rename_map": {"DBID": "PAT_DBID", "ID": "patient_id"},
+    },
+}
+
+
+def loading_and_merging_dbfs(database_directory, refresh_cache):
```
sjswerdloff, line 56: load_and_merge_dbfs would be a better name.
needs docstring (it's a public method even if it's in a private module, whoever maintains the module should know what it is)

lib/pymedphys/_streamlit/apps/wlutz/_dbf.py
```diff
@@ -0,0 +1,136 @@
...
+        merged["IMG_DATE"].astype("str")
+        + "T"
+        + merged["IMG_TIME"].astype("str")
+        + "000"
+    )
+
+    merged["datetime"] = pd.to_datetime(timestamps_string, format="%Y%m%dT%H%M%S%f")
+    merged["time"] = merged["datetime"].dt.time
+
+    merged = merged[
```
sjswerdloff, line 85: is this creating a list of a subset of a dict ?
If so, would a list of namedtuple provide better clarity in particular for the consumer?

lib/pymedphys/_streamlit/apps/wlutz/_dbf.py
sjswerdloff: premature optimisation is not good.
caching something that you know a-priori is likely to change "underneath" and with no visible warning is a bad idea.
Is the performance hit of going to the db that horrific?
If so, and if you think that db changes that are relevant to the running application could happen with some frequency, reconsider the architecture.  "Stuff changes frequently" tends to motivate an event driven architecture.

lib/pymedphys/_streamlit/apps/wlutz/_filtering.py
```diff
@@ -0,0 +1,62 @@
...
+# limitations under the License.
+
+
+import datetime
+
+from pymedphys._imports import numpy as np
+from pymedphys._imports import streamlit as st
+
+
+def filtering_image_sets(to_be_filtered):
```
sjswerdloff, line 22: filter_image_sets might be a better name.
I'm sensing a pattern of using a name that implies an ongoing activity rather than an action.
Is there a purpose for that, or is this copying some approach from "found code"?

protyping/wlutz/wlutz-with-pylinac.ipynb
```diff
@@ -0,0 +1,160 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "path = pathlib.Path(\n",
+    "    r'\\\\tunnel-nbcc-pdc\\Physics\\NBCC-DataExchange\\iViewDB\\patient_MVISO\\MV_IMAGES\\img_1.3.46.423632.301843.1579156836.86\\00022.1.3.46.423632.301843.1579156836.86.jpg'\n",
```
sjswerdloff: hardcoded path.
where is this going to work correctly, and where is it going to fail?
This looks like it's intended to be an example (or a unit test).
suggest extracting the "works only in my house" part so that it can be readily understood for what it is and modified readily by someone who needs it to work in their house.

sjswerdloff (changes requested):
docstrings

protyping/wlutz/wlutz-with-pylinac.ipynb
SimonBiggs: This one's a prototyping notebook. (in the prototyping directory). It's purely a "I'm trying to learn something for myself". So anything goes here (as long as the information isn't sensitive)

lib/pymedphys/_streamlit/apps/wlutz/_dbf.py
SimonBiggs: This is a `pandas.DataFrame` chosen for both its database like merge functionality, it's brilliant handling of tables, and its interoperability with streamlit :)

lib/pymedphys/_streamlit/apps/wlutz/_dbf.py
SimonBiggs: During my usage this was live running on one monitor while I was coding on the other. It certainly wasn't premature as it was slowing down my ability to see results down the line. I do agree however that caching this one is a bad idea. And it is likely worth considering having a watchdog based listener trigger a rerun in this case.

For now, the caching is helping me iterate on code down the line. So for now I'll leave it as is.

SimonBiggs:
@sjswerdloff for reference, here is a relevant discussion/note over on pylinac:

https://github.com/jrkerns/pylinac/issues/277#issuecomment-724427590

sjswerdloff:
> @sjswerdloff for reference, here is a relevant discussion/note over on pylinac:
> 
> [jrkerns/pylinac#277 (comment)](https://github.com/jrkerns/pylinac/issues/277#issuecomment-724427590)

So...  this is inherited code, forked/copied (not a social fork whose only purpose is to enable contributing from someone who is not "in the project") until someone looks at it and decides to manually copy stuff back in to the other project?
We break it, we own it.
It was a struggle to understand a fair amount of what was going on, and if the code you referenced is as light on docstring as this subset, they should be happy to copy/merge the docstring back in.

It's a delight to review code that is elegant and has enough documentation that I feel I've learned something.
Even when there's plenty of stuff that could use improvement, and especially if I catch something worth fixing immediately.

It's a slog to go through 25 files of terse undocumented code in one go.

sjswerdloff (approved):

SimonBiggs:
> It's a slog to go through 25 files of terse undocumented code in one go.

Yup, that's fair. I really, really don't want to make any of these reviews a slog :(. In the future, before even moving code out of draft I'll make sure it is in the following state:

> It's a delight to review code that is elegant and has enough documentation that I feel I've learned something.

Otherwise we'll loose you, and others, and PyMedPhys will be all the poorer.

Sorry for this one @sjswerdloff.

SimonBiggs:
> So... this is inherited code, forked/copied (not a social fork whose only purpose is to enable contributing from someone who is not "in the project") until someone looks at it and decides to manually copy stuff back in to the other project?

Only some of the code is a fork, a lot of it is my own. I need to own the lack of docstrings, that's my fault, not that of @jrkerns. Upon reflection, the slog you describe is also a large factor of what made @Matthew-Jennings step down.

I need to step back and make sure I make this sort of thing never a slog. So I need to own that. And if docstrings is the path to it, then I better sharpen my game.

In reality, many of the docstrings you've grown to love within PyMedPhys was the work of @Matthew-Jennings. Here is a clear area I need to improve, and the sustainability of PyMedPhys depends upon it, it seems.

lib/pymedphys/_streamlit/apps/wlutz/_dbf.py
SimonBiggs: > premature optimisation is not good.
> caching something that you know a-priori is likely to change "underneath" and with no visible warning is a bad idea.
> Is the performance hit of going to the db that horrific?

To explain myself here a bit better. The database is actually at a separate site, through an SSH tunnel. It's not that it's changing regularly, in fact when a previous date is selected for data viewing, it is not going to be changing at all. For now, so that I didn't have to wait on the internet handshake each time I changed a line of code I found caching this read was worthwhile.

lib/pymedphys/_streamlit/apps/wlutz/_filtering.py
SimonBiggs: > I'm sensing a pattern of using a name that implies an ongoing activity rather than an action.

Yup, good call. It was subconscious and I suspect it may be because as I was writing it, it was "ongoing" per-se, as the user (me) was able to update the parameters within the streamlit app and the filtering was "ongoing".

But nevertheless the action of this function when called is not ongoing.

sjswerdloff (commented):
with the docstrings added, I've learned a fair amount about the system involved (both the ELEKTA system and the original pylinac code), as well as get a much better idea of how this all hangs together. 
Nice.

SimonBiggs:
> with the docstrings added, I've learned a fair amount about the system involved (both the ELEKTA system and the original pylinac code), as well as get a much better idea of how this all hangs together.
> Nice.

Thanks @sjswerdloff, it even caused me to slightly simplify my API. And I can imagine it makes it far nicer to read. Software should be designed to be read by humans more than machines. This is a key part of it. Thanks for pushing me to be a better software developer :). Thanks @sjswerdloff :).

SimonBiggs:
Also, please don't feel like you need to dive into this soon. I have over spent your reviewing juice... need an opportunity to let it refill :).

SimonBiggs:
@sjswerdloff would you prefer I make a new PR separate from this branch to keep the changelog clean? Or would yo prefer I keep committing into this PR?

sjswerdloff:
> @sjswerdloff would you prefer I make a new PR separate from this branch to keep the changelog clean? Or would you prefer I keep committing into this PR?

I prefer getting this merged as soon as it is functional. 
I think creating a new PR might result in a repeat of review, and at this point, that's the expensive part <cough>.

If anything can wait until after this functionality gets merged, put it in a separate PR.

SimonBiggs:
Kk, I'll merge this if that's okay?

PR #1151, opened 2020-11-06
Move pymedphys to lib/pymedphys, implement root level streamlit app.py

SimonBiggs:
A range of changes to support hosting the PyMedPhys GUI on https://share.streamlit.io.

Here is the resulting hosted application:

https://share.streamlit.io/pymedphys/pymedphys/to-lib/app.py

---

Also, given this URL will likely be shared quite widely, now is the time to decide whether or not we would like to change our "trunk" branch name from "master" --> "main". All new repositories are using this new naming convention. I think it's better to make this change now, before sharing a URL like this. My preference is to make the change and to make it once this PR merges.

---

A key change is that now there is an `app.py` within the root of pymedphys that looks like the following:

https://github.com/pymedphys/pymedphys/blob/8827b4c93e5fd502db43617732d2aa66402b5ea6/app.py#L16-L35

To be able to actually have such a file a directory called "pymedphys" can't be in the root of the repo.


sjswerdloff:
pseudonymise is breaking the app because of pointing to a download directory
I think that can be corrected relatively easily by removing the part attempting to make a directory there.
And might as well remove the code that assumes files instead of BytesIO for intermediate steps.

sjswerdloff:
agree with the choice to switch to main and to do so sooner rather than later.

SimonBiggs:
> pseudonymise is breaking the app because of pointing to a download directory
> I think that can be corrected relatively easily by removing the part attempting to make a directory there.
> And might as well remove the code that assumes files instead of BytesIO for intermediate steps.

Yup, I think that should be undergone in a separate PR though. Key aim here is to move to lib/pymedphys and make sure everything works the same as before.

SimonBiggs:
> agree with the choice to switch to main and to do so sooner rather than later.

Yup, will do :)

SimonBiggs:
Sorry for all the pings :/

.github/workflows/on-push.yml
```diff
@@ -290,7 +290,7 @@ jobs:
     - name: Run slow tests
       if: matrix.task == 'slow'
       run: |
-        poetry run pymedphys dev tests -v --run-only-slow
+        poetry run pymedphys dev tests -v --slow
```
sjswerdloff, line 293: Where can I find that the changed name for the option is implemented?
I'm struggling to find it in this PR.

sjswerdloff (approved):

.github/workflows/on-push.yml
SimonBiggs: Yeah, I added thos configuration options a little while back. I didn't remove the alternative, just added an easier flag. It was bugging me during usage, I kept mixing up "run-only-slow" with "only-run-slow".

Here's where the options are configured:

https://github.com/pymedphys/pymedphys/blob/15a79cb3036e66ad111dc8a66825245ad5714eb9/pymedphys/conftest.py#L9-L34

SimonBiggs:
@sjswerdloff as a note, after this merge, you'll need to reinstall the development version of pymedphys with:

```
poetry install -E dev
```

PR #1155, opened 2020-11-07
The remnants from "Allow user to turn off `device_strict` checking"

SimonBiggs:
PR changed direction, see https://github.com/pymedphys/pymedphys/pull/1155#issuecomment-725877849

----

So in the end, I opted to allow the user to flick a switch and turn "device_strict" off. Keen for your thoughts.


pymedphys/_dicom/delivery/core.py
```diff
@@ -12,21 +12,29 @@
...
+from pymedphys._dicom.delivery import utilities
 from pymedphys._utilities.transforms import convert_IEC_angle_to_bipolar
 
-from .. import rtplan
-from . import utilities
+dicom_path_or_dataset = Union[os.PathLike, "pydicom.Dataset"]
 
 
-def load_dicom_file(filepath):
+class NotSupportedError(ValueError):
```
sjswerdloff, line 31: It's a nit, but "NotSupported" in other languages is used to denote that something hasn't been implemented yet, or in some edge cases, that a child class considers a particular method inappropriate even though the parent supports it.

So... this is ValueNotSupported(ValueError), and it is distinguished from ValueError based on what criteria? (which would be good to have in a docstring). e.g. "While the Value is valid in the sense of being part of a Defined or Enumerated type, it is not acceptable in the current circumstances"

pymedphys/_dicom/delivery/core.py
```diff
@@ -37,15 +45,72 @@ def _pretty_print(string):
...
+        )
+
+
 class DeliveryDicom(DeliveryBase):
     @classmethod
-    def from_dicom(cls, dicom_dataset, fraction_number=None):
+    def from_dicom(
+        cls, rtplan: dicom_path_or_dataset, fraction_number=None, device_strict=True
+    ):
+        if isinstance(rtplan, pydicom.Dataset):
```
sjswerdloff, line 103: this seems like an incredibly useful few lines of code (pathlike | object one would deserialise from a file having said path).  While I don't think it's appropriate at the moment to extract it, I believe we should keep our eyes open for other places where it would be very useful to see just what level of generalization of the concept is appropriate.

pymedphys/_dicom/delivery/core.py
```diff
@@ -123,53 +195,14 @@ def _from_dicom_file(cls, filepath, fraction_number):
         return cls.from_dicom(load_dicom_file(filepath), fraction_number)
 
     @classmethod
-    def _from_dicom_beam(cls, beam, meterset):
+    def _from_dicom_beam(cls, beam, meterset, device_strict=True):
         if meterset is None:
             raise ValueError("Meterset should not ever be None")
```
sjswerdloff, line 200: Might be an American thing, but "Meterset should never be None"

pymedphys/_streamlit/apps/mudensity/_dicom.py
```diff
@@ -145,13 +147,43 @@ def dicom_input_method(  # pylint: disable = too-many-return-statements
...
+            """You have the option to set `device_strict=False` as long
+            as you are willing to accept that this is removing a safety
+            check from this software and will likely result in ignoring
+            one or more of the collimation devices that are within the
+            provided DICOM file.
+            """
+        )
+
+        if st.button("Set `device_strict=False` for this session"):
+            session_state.device_strict = False
```
sjswerdloff, line 178: This is where you absolutely must have a logging of someone choosing this value, with whatever identity is available and the timestamp.  Preferably in a separate audit, but at least in a log.
If identify is not available intrinsically, the user should have to identify themselves and have that in the audit/log (it's better than nothing, and demonstrates that you at least considered the situation, and if they leave the identify field blank, don't accept a setting to False for device_strict).

This is exactly the kind of thing that ends in tears

sjswerdloff (changes requested):
When the user chooses to set device_strict to False, that decision needs to be entered in to an audit (or at least a log) with their identity and a timestamp.
Not negotiable. 
audit entry should include information on which data is being operated on, or alternatively for any data for which device_strict is set to false, there should be an error level log (always logged) indicating that for "this" data, device_strict was set to false.  One could constrain that to only log if it actually failed strict validation and then that failure was ignored.

sjswerdloff:
I think that as a first pass using a general override is OK, assuming appropriate audit.
But make no mistake about it, it's an override, and all of the risk mitigation that one needs to apply for an override must be utilised.
The approach leaves room for separate additional beam limiting device combinations, potentially through site configuration.

pymedphys/_dicom/delivery/core.py
SimonBiggs: Yeah, Python has "NotImplementedError" which I was avoiding using. Here, the custom error class was so that I could explicitly catch this error and no other within the GUI. See where I have caught that error within streamlit:

https://github.com/pymedphys/pymedphys/blob/5f569a1be141cdd30cef94224c7e00b80568a743/pymedphys/_streamlit/apps/mudensity/_dicom.py#L159-L181

More than happy for a different name. In every sense of the word it is still a ValueError (which is why I am deriving from it), but I don't want a different ValueError to begin down the pathway designated within the GUI in that code linked in above.

pymedphys/_dicom/delivery/core.py
SimonBiggs: Yup I agree.

pymedphys/_streamlit/apps/mudensity/_dicom.py
SimonBiggs: Yup, fair. I wonder if MU Density should have a "user" field right at the beginning, that user can then be included within the report produced at the end, and if any override is applied, that can also be included within the PDF report.

I'll work on that in a separate PR and then once that's done come back to this PR and see if that can be used here.

SimonBiggs:
Hi @sjswerdloff,

So I had a bit more of a think about this PR. I don't like the approach I underwent here. It sets a bad precedent, and there are better solutions. I would rather no solution than this one.

However, there were quite a few little tid-bits that where iterated on in this PR, and I'd like to still merge those. To make it a little easier to review I have tried to add in docstrings when I have touched a function, also, when functions are meant to be private in the files I have tried to indicate as such.

I figure, if I make a habit of writing a docstring whenever I touch a function/class/method for a PR over time the docstrings available will improve. Slowly, but they'll improve.

So with that, would you be able to re-review this one for me?

lib/pymedphys/_dicom/delivery/core.py
```diff
@@ -37,15 +41,106 @@ def _pretty_print(string):
...
-    def from_dicom(cls, dicom_dataset, fraction_number=None):
+    def from_dicom(cls, rtplan: dicom_path_or_dataset, fraction_number=None):
+        """Create a ``pymedphys.Delivery`` object from an RT Plan DICOM
+        dataset.
+
+        Parameters
+        ----------
+        rtplan : pydicom.Dataset or pathlib.Path
+            An RT Plan DICOM dataset, or the filepath to such a dataset.
+        fraction_number : 'all' or int, optional
```
sjswerdloff, line 117: So it's actually FractionGroupNumber, and not fraction number...
I suggest renaming the variable.

lib/pymedphys/_dicom/delivery/core.py
```diff
@@ -37,15 +41,106 @@ def _pretty_print(string):
...
+            rtplan_filepath = cast(os.PathLike, rtplan)
+            rtplan_dataset = _load_dicom_file(rtplan_filepath)
 
         if str(fraction_number).lower() == "all":
-            return cls._load_all_fractions(dicom_dataset)
+            return cls._load_all_fractions(rtplan_dataset)
 
         if fraction_number is None:
-            fractions = dicom_dataset.FractionGroupSequence
+            fractions = rtplan_dataset.FractionGroupSequence
```
sjswerdloff, line 143: fraction_groups

lib/pymedphys/_dicom/delivery/core.py
```diff
@@ -37,15 +41,106 @@ def _pretty_print(string):
...
             fraction_numbers = [fraction.FractionGroupNumber for fraction in fractions]
```
sjswerdloff, line 144: fraction_group_numbers

lib/pymedphys/_streamlit/utilities/session.py
```diff
@@ -21,10 +21,17 @@
...
 class SessionState:
-    def __init__(self, **kwargs):
-        for key, val in kwargs.items():
+    def __init__(self, state):
+        for key, val in state.items():
             setattr(self, key, val)
 
+    def update(self, state):
+        for key, val in state.items():
+            try:
```
sjswerdloff, line 30: isn't there a has_key so that you can use a conditional instead of an exception?

sjswerdloff (changes requested):
see detailed comments

sjswerdloff:
I'd prefer to not review the same code multiple times.
But I'll look at that as the opportunity to set higher expectations.
If you find name changes in signatures create too much ripple, at least copy to a local variable with a more appropriate name.
But I consider variable names in signatures of incredibly high importance because misinterpreting what to pass in over an interface is hard to test for and can have grievous consequences (mars lander, anyone?)

SimonBiggs:
Hehe, I was trying to work out how to buy you a beer... turns out this stuff isn't made anymore... but it would've been perfect :beer: 

https://untappd.com/b/moa-brewing-company-manuka-honey-ale/3170663

lib/pymedphys/_streamlit/utilities/session.py
SimonBiggs: For here, EAFP instead of LBYL tends to be the Python idiom. See https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

SimonBiggs:
> I'd prefer to not review the same code multiple times.

Sorry about that. Might be worth revisiting the review schedule/approach/magnitude/method. Keen for your thoughts, suggestions.

> But I'll look at that as the opportunity to set higher expectations.
> If you find name changes in signatures create too much ripple, at least copy to a local variable with a more appropriate name.
> But I consider variable names in signatures of incredibly high importance because misinterpreting what to pass in over an interface is hard to test for and can have grievous consequences (mars lander, anyone?)

Thus far the `fraction_number` parameter within "pymedphys.Delivery.from_dicom" has not been widely broadcast. I would be surprised if anyone is using it. As such, given the `0.x.x` version number, I think this is an ideal opportunity to correct this misconception of mine which translated into signature naming. As such I have made that change and implemented all your review comments except the request to change EAFP to LBYL.

sjswerdloff (approved):

lib/pymedphys/_streamlit/utilities/session.py
sjswerdloff: I stand corrected.
Misplaced habit.

SimonBiggs: The docs have some neat comments on EAFP regarding multi-threaded environments:

https://docs.python.org/3.5/glossary.html#term-lbyl

PR #1156, opened 2020-11-08
Pin images to commit hash not branch name

SimonBiggs:
So that the image links around the web aren't dependent on a branch name.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -144,10 +144,6 @@ sections = [
     "MEDPHYS", "FIRSTPARTY", "LOCALFOLDER"
 ]
 
-[tool.dephell.main]
-from = {format = "poetry", path = "pyproject.toml"}
-to = {format = "setuppy", path = "setup.py"}
-
 [build-system]
 requires = ["poetry >= 1.0.9", "setuptools >= 42.0.0", "wheel >= 0.29.0"]
 build-backend = "poetry.masonry.api"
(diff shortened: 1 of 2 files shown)


sjswerdloff:
The implication being that one someone commits a new image, they have to update the README.
Not a bad thing, but something to keep in mind

pyproject.toml
```diff
@@ -144,10 +144,6 @@ sections = [
     "MEDPHYS", "FIRSTPARTY", "LOCALFOLDER"
 ]
 
-[tool.dephell.main]
```
sjswerdloff, line 147: This doesn't seem related to pinning the images.
While it's not an issue, it wasn't clear from the top level commit line

sjswerdloff (approved):

SimonBiggs:
> The implication being that one someone commits a new image, they have to update the README.
> Not a bad thing, but something to keep in mind

Yup, better than before though where the image in the PRs readme wouldn't actually appear changed until it merged.

pyproject.toml
SimonBiggs: Yeah good cool. I was doing a CTRL+F for main and master and that config popped up. It's old config that isn't doing anything. Figured a fly by removal wasn't an issue in this case.

sjswerdloff: It's not an issue.
But it could have been if it were accidental.
So I thought I would mention it without preventing progress

SimonBiggs: Yup good call.

PR #1157, opened 2020-11-08
Fix docs building on Netlify

SimonBiggs:
Docs was broken on Netlify with the move to `lib/pymedphys`. It seems to be going to town trying to use a system installed pymedphys.

Anyway, for a while now I haven't really been able to move the docs folder because all the images on the web were linked back to `master/docs/logos`. But now that master is becoming static for a bit, and we're using main, I can move the docs around.

Also, I have been wanting to deploy the documentation with the package itself. One day, the hope would be to have a link to the documentation from within the GUI, where the documentation was the locally hosted version that matches the user's installed version.

So, given netlify is trying to build docs from a production install... maybe, instead of trying to fight it, instead I am going to go with it, and implement what I have been wanting to do for a while. Make the documentation be packaged within the library itself.


SimonBiggs:
I also aimed to address the following issue in this PR also:

https://github.com/pymedphys/pymedphys/issues/1158

lib/pymedphys/docs/conf.py
```diff
@@ -39,7 +39,7 @@ def setup(app):
 # The short X.Y version
 # Currently being used to create mybinder link -- potentially need to rethink
 # this logic.
-version = "main"
+version = "fix-netlify"
```
SimonBiggs, line 42: Need to revert this before merging

```suggestion
version = "main"
```

lib/pymedphys/docs/conf.py
```diff
@@ -119,9 +119,10 @@ def setup(app):
 
         This page was generated from `{{ docname }}`__.
```
SimonBiggs, line 120:
```suggestion
        This page was generated from `/lib/pymedphys/{{ docname }}`__.
```

lib/pymedphys/cli/dev.py
```diff
@@ -15,6 +15,7 @@ def add_docs_parser(dev_subparsers):
     parser = dev_subparsers.add_parser("docs")
 
     parser.add_argument("--live", help="Make the docs live reload", action="store_true")
+    parser.add_argument("--output", help="Custom output directory")
```
sjswerdloff, line 18: will this be obvious to the user that it is the custom output directory of the docs? (vs. something else)

SimonBiggs, line 18:
```suggestion
    parser.add_argument("--output", help="Custom output directory for the built docs.")
```

SimonBiggs: Yup, I can see how that could be ambiguous. Just made a small addition. I might need to jump on a PC and re-blackify it though.

sjswerdloff (approved):

PR #1161, opened 2020-11-09
Initial code for a Dicom connect module

pchlap:
This pull request contains the code for a Dicom connect module. I have been using this code for many different purposes. Whats in here so far is the DicomListener class, which can store Dicom files which are sent to it, and also responds to C-Echo.

@SimonBiggs Maybe a little more code that you had initially hoped for, but I hope that it is understandable as is. I wanted to have something that works. This code works as a Dicom listener.

Simon, if you'd like to give a once over as far as the structure of the code goes. If you have any feedback there it would be easier to shift things around before getting in too deep.

Still to do here is some documentation. Later (in future pull requests) I will add the code for Dicom query and retrieve.

I hope this sort of thing would be useful for others too. I use this all the time from a number of scripts so it would be great to have this code cleaned up and well tested. Once this is looking good to you Simon perhaps we can identify another contributer to give it a review.

Looking forward to any and all feedback on this!


SimonBiggs:
Awesome :) thanks @pchlap I'll have a look at this tonight. Also, many hands make light work for reviewing, might you have a chance to look over the following PR:

https://github.com/pymedphys/pymedphys/pull/1143

...it's some code for two experimental apps, they're still a work in progress, but extra eyes on the code is helpful :).

pchlap:
> Awesome :) thanks @pchlap I'll have a look at this tonight. Also, many hands make light work for reviewing, might you have a chance to look over the following PR:
> 
> #1143
> 
> ...it's some code for two experimental apps, they're still a work in progress, but extra eyes on the code is helpful :).

Happy to. Will do that tonight or tomorrow :)

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+class DicomListener(DicomConnectBase):
+    def __init__(
+        self, storage_directory=tempfile.mkdtemp(), on_released_callback=None, **kwargs
+    ):
+        super(DicomListener, self).__init__(**kwargs)
+
+        # The directory where incoming data will be stored
+        self.storage_directory = storage_directory
```
sjswerdloff, line 43: this could be reduced to 
self.storage_directory = Path(storage_directory)
if storage_directory is already a Path, it still comes out as a Path

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+        mode_prefixes = {
+            "CT Image Storage": "CT",
+            "Enhanced CT Image Storage": "CTE",
+            "MR Image Storage": "MR",
+            "Enhanced MR Image Storage": "MRE",
+            "Positron Emission Tomography Image Storage": "PT",
+            "Enhanced PET Image Storage": "PTE",
+            "RT Image Storage": "RI",
+            "RT Dose Storage": "RD",
+            "RT Plan Storage": "RP",
```
sjswerdloff, line 111: RT Ion Plan?
RT Beams Treatment Record (this is used in conventional, so it is relevant to typical departments)
RT Ion Beams Treatment Record

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+
+        if self.on_released_callback:
+            self.on_released_callback(self.association_directory)
+
+    def on_c_store(self, event):
+
+        dataset = event.dataset
+
+        # TODO: Should these be place in a config file?
+        mode_prefixes = {
```
sjswerdloff, line 102: These are offered elsewhere in PyMedPhys
see _dicom.constants.core.DICOM_SOP_CLASS_NAMES_MODE_PREFIXES
If something is missing from that list, it would be good to add it there.

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+            "Nuclear Medicine Image Storage": "NM",
+            "Secondary Capture Image Storage": "SC",
+        }
+
+        try:
+            mode_prefix = mode_prefixes[dataset.SOPClassUID.name]
+        except KeyError:
+            mode_prefix = "UN"
+
+        suid = dataset.SeriesInstanceUID
```
sjswerdloff, line 127: suid has a very specific meaning in Unix (the "set uid bit").
may I request that you name it series_uid or something else, but not suid ?

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+        series_dir = self.storage_directory / suid
```
sjswerdloff, line 128: I'd need to defer to you and Simon, but I'm used to a Path having a subdirectory added using
pathlib.Path.joinpath(...)

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+        series_dir.mkdir(exist_ok=True)
+
+        filename = "{0!s}.{1!s}".format(mode_prefix, dataset.SOPInstanceUID)
+        filepath = series_dir / filename
+
+        if filepath.exists:
+            # TODO Currently just logging this, since I'm unsure of the desired
+            # behaviour if the file has already been recieved.
+            logging.debug("DICOM file already exists, overwriting")
```
sjswerdloff, line 137: Strongly recommend not overwriting with the same SOP Instance UID.
The implication is that someone could have made a decision based on the content of an SOP Instance and then have that change underneath them after the fact.
It's not an error if the data are the same.  It is an error or at least a problem if the data is not consistent.

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+        meta = Dataset()
+        meta.MediaStorageSOPClassUID = dataset.SOPClassUID
+        meta.MediaStorageSOPInstanceUID = dataset.SOPInstanceUID
+        meta.ImplementationClassUID = PYNETDICOM_IMPLEMENTATION_UID
+        meta.TransferSyntaxUID = context.transfer_syntax
+
+        # The following is not mandatory, set for convenience
+        meta.ImplementationVersionName = PYNETDICOM_IMPLEMENTATION_VERSION
+        file_ds = FileDataset(
+            filepath.as_posix(), {}, file_meta=meta, preamble=b"\0" * 128
```
sjswerdloff, line 149: is filepath.as_posix() portable (i.e. work on MS Windows)?

sjswerdloff: Including the preamble means that the data is susceptible to the DICOM Preamble Vulnerability.
see
https://www.medicalconnections.co.uk/kb/DICOM-Preamble-Vulnerability 
It's OK to use it, but one must be cautious about leaving the files available to a user via the file system because malware can find the files and infect them.

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+        status_ds = Dataset()
+        status_ds.Status = 0x0000
+
+        try:
+            # We use `write_like_original=False` to ensure that a compliant
+            #   File Meta Information Header is written
+            file_ds.save_as(filepath.as_posix(), write_like_original=False)
+            status_ds.Status = 0x0000  # Success
+        except IOError:
+            logging.warning("Could not write file to specified directory:")
```
sjswerdloff, line 164: suggest using error level.
This makes the current running instance of the application not able to fulfill it's intended purpose.
Warnings are for "this isn't good, and it could cause problems, but you might be able to squeeze by"

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+            logging.warning("    %s", filepath)
+            logging.warning(
+                "Directory may not exist or you may not have write " "permission"
+            )
+            # Failed - Out of Resources - IOError
+            status_ds.Status = 0xA700
+        except Exception as exception:
+            logging.warning("An error occured saving the Dicom file:")
```
sjswerdloff, line 172: suggest using error level as mentioned above

sjswerdloff: If 

(0000,0902) | Error Comment | ErrorComment | LO | 1 | This field contains an application-specific text description of the error detected.
-- | -- | -- | -- | -- | --



is available through pynetdicom in the status_ds, it is highly useful to the SCU (or the people responsible for the SCU).  Which in turn often saves grief for those responsible for the SCP.

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+        file_ds.is_little_endian = context.transfer_syntax.is_little_endian
+        file_ds.is_implicit_VR = context.transfer_syntax.is_implicit_VR
+
+        status_ds = Dataset()
+        status_ds.Status = 0x0000
+
+        try:
+            # We use `write_like_original=False` to ensure that a compliant
+            #   File Meta Information Header is written
+            file_ds.save_as(filepath.as_posix(), write_like_original=False)
```
sjswerdloff, line 161: saving a file can take a long time, and create a bottleneck.
Is this something that could be done async ?

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+        # Initialise the Application Entity
+        self.ae = AE(ae_title=self.ae_title)
+
+        # Add the supported presentation context
+        self.ae.add_supported_context(VerificationSOPClass)
+        for context in StoragePresentationContexts:
+            self.ae.add_supported_context(context.abstract_syntax)
+
+        handlers = [
+            (evt.EVT_C_MOVE, self.on_c_store),
```
sjswerdloff, line 68: C-MOVE is not C-STORE.
If the SCP receives a C-MOVE-RQ it will not be given an object to store.
Suggest deleting this for now.  
Implementing will involve being able to find objects on the filesystem, potentially only identified by SOP Instance UID (no hierarchy information), and the ability to issue a C-STORE-RQ to an entity only identified by AE Title (no IP address or port information), which then relies on the SCP having the means to get the IP address and port information based on the AE Title (e.g. via configuration or using Supplement 67 (Configuration Management)).  NB: not many implementations of Supp 67 (pity).

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -0,0 +1,124 @@
...
+
+        assoc.release()
+
+    # Check we got a valid result
+    assert result == 0
+
+
+@pytest.mark.pydicom
+def test_dicom_listener_send(listener):
+    """Test to ensure that running dicom listener recieves a stores a Dicom file
```
sjswerdloff, line 73: i before e except after c ("or when sounding like A, as in neighbor or weigh"): "receives"
personally I struggle with the spelling of "occurrence", so... no judgment.

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -0,0 +1,124 @@
...
+    # Send the data to the listener
+    ae = AE()
+    ae.add_requested_context(RTPlanStorage)
+    assoc = ae.associate(listener.host, listener.port, ae_title="PYMEDPHYSTEST")
+    assert assoc.is_established
+    status = assoc.send_c_store(test_dataset)
+    assert status.Status == 0
+    assoc.release()
+
+    # Check that it was recieved
```
sjswerdloff, line 113: consistency is a good sign :-) ...
"received"

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -0,0 +1,124 @@
...
+
+    METHOD_MOCK.reset_mock()
+
+    # Create a test Dicom object
+    test_uid = pydicom.uid.generate_uid()
+    test_dataset = dicom_dataset_from_dict(
+        {
+            "SOPClassUID": RTPlanStorage,
+            "SOPInstanceUID": test_uid,
+            "SeriesInstanceUID": test_uid,
```
sjswerdloff, line 84: it's better to not have two UIDs being the same, even if one is for the Series and one for the Instance (unless of course they are intentionally the same Study or Series UID indicating common heritage).

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -0,0 +1,124 @@
...
+@pytest.mark.pydicom
+def test_dicom_listener_send(listener):
+    """Test to ensure that running dicom listener recieves a stores a Dicom file
+    """
+
+    METHOD_MOCK.reset_mock()
+
+    # Create a test Dicom object
+    test_uid = pydicom.uid.generate_uid()
+    test_dataset = dicom_dataset_from_dict(
```
sjswerdloff, line 80: an alternative would be to download one of the test data plans from zenodo and have a reasonably compliant RT Plan for the test.

sjswerdloff:
Hi @pchlap ...
Simon promoted me to maintainer, and I think part of that is to offload some work from him (at least while I'm on Sabbatical).
I like what I see.  
But Simon has encouraged a very high level of discipline, so I am following his lead and being quite picky.

pchlap:
> Hi @pchlap ...
> Simon promoted me to maintainer, and I think part of that is to offload some work from him (at least while I'm on Sabbatical).
> I like what I see.
> But Simon has encouraged a very high level of discipline, so I am following his lead and being quite picky.

Thanks @sjswerdloff 
I appreciate the review and definitely happy for you to take it apart and be picky. Thats the best way to improve and produce high quality code.

I'll work my way through your comments and I also need to figure out why the tests are failing (since they pass when I run locally).

SimonBiggs:
> I also need to figure out why the tests are failing (since they pass when I run locally).

I think you may need to import pydicom from `pymedphys._imports` instead of directly. That's to support environments that wish to use minimal dependencies.

README.rst
```diff
@@ -111,7 +111,7 @@ Active contributors
 
 * `Phillip Chlap`_
     * `University of New South Wales`_, Australia
-    * `South Western Sydney Local Health District`_, Australia
+    * `Ingham Institute`_, Australia
```
SimonBiggs, line 114: Logos are welcome for anyone who's company/organisation wants them. I understand if not though.

SimonBiggs:
> I think you may need to import pydicom from `pymedphys._imports` instead of directly. That's to support environments that wish to use minimal dependencies.

@pchlap if you'd like I can make the changes required to make the clean tests pass this time through. That way you can see what's needed for that one.

SimonBiggs:
> Simon promoted me to maintainer, and I think part of that is to offload some work from him (at least while I'm on Sabbatical).

Thanks @sjswerdloff for jumping in here. Massively appreciated.

SimonBiggs (changes requested):
Hi @pchlap,

This is a pretty good size. It certainly is on the bigger end, but it hasn't crossed over into too big :).

I guess it's telling that I felt like I could review it in one sitting and it didn't take ages. Most of the bits here and there are things that will smooth out in the future (such as pylint, pymedphys._imports and numpy style docstrings).

There are two key changes I have though, have a check of the `tempfile` logic as discussed in one of my comments, and second, need a test that flexes the CLI since, as far as I can tell, that is going to be the main method by which this utility is spun up.

All other comments are "Nit." from me, meaning a "nitpick". Feel free to implement those or not depending on what you think is best.

@sjswerdloff had some key change requests also.

Thanks @pchlap, really appreciate this! Looking forward to more :).

lib/pymedphys/_dicom/connect/listen.py
SimonBiggs: I'm not the biggest fan of the `except Exception` here. I personally would much rather it stop working, scream at me, and fail fast, than keep on trucking for weeks making me think everything is okay, when instead it's been silently dropping or munging data.

pchlap:
> Hi @pchlap,
> 
> This is a pretty good size. It certainly is on the bigger end, but it hasn't crossed over into too big :).
> 
> I guess it's telling that I felt like I could review it in one sitting and it didn't take ages. Most of the bits here and there are things that will smooth out in the future (such as pylint, pymedphys._imports and numpy style docstrings).
> 
> There are two key changes I have though, have a check of the `tempfile` logic as discussed in one of my comments, and second, need a test that flexes the CLI since, as far as I can tell, that is going to be the main method by which this utility is spun up.
> 
> All other comments are "Nit." from me, meaning a "nitpick". Feel free to implement those or not depending on what you think is best.
> 
> @sjswerdloff had some key change requests also.
> 
> Thanks @pchlap, really appreciate this! Looking forward to more :).

Thanks Simon! Great comments, very useful. I will work through them today I hope.

lib/pymedphys/_dicom/connect/listen.py
SimonBiggs: I didn't know about that preamble vulnerability. Given this codes aspiration to one day be a part of a hosted API this is a key pick up. Thanks @sjswerdloff. Really appreciate your eyes on this.

lib/pymedphys/_dicom/connect/listen.py
pchlap: I don't feel strongly either way. I think the only advantage of the / syntax is that if you need to chain a number of subdirectories together that can be easier to read than a number of .joinpath()'s.

But if you both prefer joinpath I'll go for that too to keep things consistent in pymedphys.

lib/pymedphys/_dicom/connect/listen.py
pchlap: Thanks for the good advice. In that case I'll avoid overwriting it and log a warning that it was not written due to the file already existing.

pchlap: My question here is what would be an appropriate status code to return if we discarded the file due to the file already existing? I'm looking at the list here and it isn't immediately obvious to me: https://pynetdicom2.readthedocs.io/en/latest/statuses.html

Perhaps an A7XX code, out of resources?

lib/pymedphys/_dicom/connect/listen.py
SimonBiggs: With the chaining, joinpath accepts multiple args where each arg is a chained path. eg the following works:

```python
a_pathlib_path = pathlib.Path.home()
a_pathlib_path.joinpath("something", "else", "as", "well")
```

lib/pymedphys/_dicom/connect/listen.py
pchlap: That is interesting, good to know!

I found this way of setting the preamble: `b"\0" * 128`  from the pydicom docs. This seems to write 128 zero bytes to the preamble, so would that avoid the vulnerability you mentioned? So would this overwrite any (potentially malicious) preamble in the dataset received? Or am I misunderstanding something here?

lib/pymedphys/_dicom/connect/listen.py
SimonBiggs: A thought is that the file on disk could be hashed, and the file just sent could be hashed. If they are the same as each other then everything can move on, no issue. However, if they are different, then an error code would be appropriate. Something like new DICOM file with same SOP instance UID provided as one already on disk, but files do not match.

lib/pymedphys/_dicom/connect/listen.py
pchlap: As for the as_posix function, I do believe that it works on Windows, but there maybe be a more appropriate way to convert a platlib.Path back to a string. Do either of you have a better suggestion to use in scenarios like this?

lib/pymedphys/_dicom/connect/listen.py
pchlap: You're no doubt correct, but I think I had this working in the past so that is could receive files from Orthanc which were sent via a C-MOVE from there...

I'll definitely remove it for now, and will investigate what is needed to have this working properly in the future.

lib/pymedphys/tests/dicom/test_connect.py
pchlap: Oops, thanks :)

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+from pynetdicom.sop_class import (  # pylint: disable=no-name-in-module
+    VerificationSOPClass,
+)
+
+from pymedphys._dicom.connect.base import DicomConnectBase
+
+
+class DicomListener(DicomConnectBase):
+    def __init__(
+        self, storage_directory=tempfile.mkdtemp(), on_released_callback=None, **kwargs
```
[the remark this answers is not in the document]
pchlap: Thanks Simon, this is why contributing to pymedphys is so valuable. I didn't know that and I used to do this all the time in my code, now I'll be sure to never do it again :D

README.rst
pchlap: Yeah UNSW are a bit funny about using their logo without explicit approval. But I'll keep it in mind and raise it next time I am chatting with someone who might know more around the permissions of using it.

lib/pymedphys/_dicom/connect/listen.py
SimonBiggs: str(path) does the trick :). But actually, it's not needed here. We've pinned pydicom to >= 2.0.0 in pymedphys, and that version of pydicom supports pathlib objects natively :).

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -16,22 +16,23 @@
...
 
 class DicomListener(DicomConnectBase):
-    def __init__(
-        self, storage_directory=tempfile.mkdtemp(), on_released_callback=None, **kwargs
-    ):
+    def __init__(self, storage_directory=None, on_released_callback=None, **kwargs):
         super().__init__(**kwargs)
 
+        # If no storage directory is set, create a temporary directory
+        if not storage_directory:
```
SimonBiggs, line 31: if storage_directory is None:

lib/pymedphys/_dicom/connect/listen.py
pchlap: Even better!

lib/pymedphys/_dicom/connect/listen.py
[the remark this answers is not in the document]
SimonBiggs: That's awesome to hear :). It's nice when goals of PyMedPhys like these actually happen :). It's encouraging that this really may be a great way to make us all write better software :)

README.rst
SimonBiggs: I was actually more thinking Ingham Institute, if they're interested...

lib/pymedphys/_dicom/connect/listen.py
sjswerdloff: See http://dicom.nema.org/medical/dicom/current/output/chtml/part07/chapter_9.html#sect_9.1.1.1.9 
Perhaps: Invalid SOP Instance (0117H) - Indicates that the SOP Instance UID specified implied a violation of the UID construction rules.
In conjunction with Simon's comment.
But in general, receiving a duplicate is not in fact an error.
You just document your behaviour in your conformance statement.

You could also respond with:
Warning (Status value is Service Class specific) - This indicates that the peer DIMSE Service User was able to store the composite SOP Instance, but detected a probable error.  Where a Warning would have a status of: 0001 or Bxxx or 0107 or 0116

However, be aware that most implementations will take a status that isn't 0x0000 to be a failure and respond accordingly.

pchlap: Thanks for the great info. I've had a go at making this more graceful. I check the hash on the dicom converted to json string of the existing file and the incoming file. This would only hash the object attributes and not the meta data which I think is desired.

Hmmmm, as I write this I don't think this will work for pixel data, since that wouldn't be output to the json which is what is hashed. I can hash that separately, but am open to suggestion on how to do this better.

If the hashes match, the file is ignored but a 0x0000 status is returned. If it conflicts the file is also discarded but a warning is returned.

lib/pymedphys/_dicom/connect/listen.py
pchlap: I've set the ErrorComment which does seem to be supported by pynetdicom. I've removed that last Exception now Simon so it will crash much more loudly if anything unexpected happens.

README.rst
pchlap: Ah yep, I'll talk to them about it :)

lib/pymedphys/_dicom/connect/listen.py
SimonBiggs: If it conflicts might it be worth storing the file sent in a "debugging" folder or something of the like?

sjswerdloff: > If it conflicts might it be worth storing the file sent in a "debugging" folder or something of the like?

That's a very good idea.
In general it is very useful to have a place to put stuff that failed for some reason for future analysis (or manual repair)

lib/pymedphys/_dicom/connect/listen.py
sjswerdloff: > That is interesting, good to know!
> 
> I found this way of setting the preamble: `b"\0" * 128` from the pydicom docs. This seems to write 128 zero bytes to the preamble, so would that avoid the vulnerability you mentioned? So would this overwrite any (potentially malicious) preamble in the dataset received? Or am I misunderstanding something here?

When receiving a file over the DICOM wire protocol, there is no pre-amble, so data received that way is sterile.
When you write a file with a preamble with 0's, the file is momentarily safe.
But once it is on the file system, malware that seeks DICOM files can modify the preamble to contain a virus.
Should a user at that point double click the DICOM file (thinking that they are going to launch a default application/viewer), the executable (virus) will start running.

So the take-home is that any software reading in a DICOM file with a pre-amble should sterilise it (either by overwriting with zeros, or by removing the preamble).  And any system that stores DICOM files should consider only storing them without the preamble, and either periodically or on activation, sweep through to make sure that none of the DICOM files have preambles (or alternatively, ensure they are sterile, e.g. have all 0's).

For the DICOM Standard organisations point of view see:
https://www.dicomstandard.org/docs/librariesprovider2/dicomdocuments/wp-cotent/uploads/2019/05/faq-dicom-128-byte-preamble-posted1-1.pdf

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -0,0 +1,188 @@
...
+    """pytest fixture to returna a dummy DICOM dataset for testing
+
+    Returns
+    -------
+    test_dataset : ``pydicom.dataset.Dataset``
+        A dummy DICOM dataset
+    """
+
+    # Create a test Dicom object
+    test_uid = pydicom.uid.generate_uid()
```
sjswerdloff, line 82:
```suggestion
    test_uid = pydicom.uid.generate_uid()
    test_series_uid = pydicom.uid.generate_uid()
```

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -0,0 +1,188 @@
...
+    test_dataset = dicom_dataset_from_dict(
+        {
+            "SOPClassUID": RTPlanStorage,
+            "SOPInstanceUID": test_uid,
+            "SeriesInstanceUID": test_uid,
```
sjswerdloff, line 87:
```suggestion
            "SeriesInstanceUID": test_series_uid,
```

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -0,0 +1,124 @@
...
+    args, _ = METHOD_MOCK.method.call_args_list[0]
+    storage_path = args[0]
+    file_path = storage_path / f"RP.{test_uid}"
+    assert file_path.exists()
+
+    read_dataset = pydicom.read_file(file_path.as_posix())
+    assert read_dataset.SeriesInstanceUID == test_dataset.SeriesInstanceUID
+
+    # Clean up after ourselves
+    shutil.rmtree(storage_path.as_posix())
```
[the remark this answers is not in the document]
pchlap: Just putting this together now, where would be the most appropriate place to put that context manager? Is there some sort of test utilities file/module? (I've had a look but nothing was immediately obvious).

lib/pymedphys/tests/dicom/test_connect.py
pchlap: I think I will ultimately do this, but in future pull requests I will add code for DICOM query/retrieve functionality. No doubt I'll need some more/different DICOM files to test that, so I will look at putting them onto Zenodo when that time comes.

lib/pymedphys/_dicom/connect/listen.py
pchlap: Sounds good, I will implement this. From CTP I know it as quarantine :D

lib/pymedphys/_dicom/connect/listen.py
pchlap: That's a great explanation, thank you.

lib/pymedphys/tests/dicom/test_connect.py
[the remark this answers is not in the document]
SimonBiggs: `lib/pymedphys/_utilities/` is a little bit of a catch all (it also feels a little bit like a cop out). As long as the directory is proceeded with an `_` it's part of our private API and we're free to move it around if later we don't like its location. So in short, I am happy for you to make the judgement call and either make a location especially for it and future things like it, or put it somewhere that is currently in place.

As a rule of thumb, I find the best location for something sometimes is only discernible down the track. So it's okay to put it somewhere sensible, and then feel free to move it later.

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -187,3 +198,41 @@ def test_dicom_listener_send_conflicting_file(listener, test_dataset):
...
+            if elapsed >= 3:  # Break if still not connecting after 3 seconds
+                break
+
+            assoc = ae.associate("127.0.0.1", TEST_PORT, ae_title=scp_ae_title)
+        assert assoc.is_established
+        status = assoc.send_c_store(test_dataset)
+        assert status.Status == 0
+        assoc.release()
+
+    series_dir = HERE.joinpath(test_dataset.SeriesInstanceUID)
```
SimonBiggs, line 232: With these sort of tests, we need to be able to support running multiple in parallel. For example, at times I want to spin up Python 3.6, 3.7, 3.8, and 3.9 and run the tests across all Python versions. As such, tests can't be creating and deleting files within a deterministic shared directory. Need to utilise tempdir creation here.

lib/pymedphys/_dicom/connect/listen.py
pchlap: > Hmmmm, as I write this I don't think this will work for pixel data, since that wouldn't be output to the json which is what is hashed. I can hash that separately, but am open to suggestion on how to do this better.

The to_json() function actually does output the pixel data, so this approach will work if the pixel data is conflicting after all.

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -187,3 +198,41 @@ def test_dicom_listener_send_conflicting_file(listener, test_dataset):
...
     shutil.rmtree(storage_path)
+
+
+@pytest.mark.pydicom
+def test_dicom_listener_cli(test_dataset):
+
+    scp_ae_title = "PYMEDPHYSTEST"
+
+    with process(
+        f"poetry run pymedphys dicom listen {TEST_PORT} -d {HERE} -a {scp_ae_title}",
```
SimonBiggs, line 209: I suspect calling `poetry run` here will break Darcy's pydicom master push:

https://github.com/pydicom/pydicom/blob/e58a922c132246a6a9b236eae1fe0015ff9d8879/.github/workflows/master-push.yml#L148-L149

When his tests are run they are run from the production pymedphys install. As such, no poetry available. Does this process appropriately scope if it's simply called as `pymedphys dicom listen`?

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -28,11 +28,11 @@ class DicomListener(DicomConnectBase):
     """
 
     def __init__(self, storage_directory=None, on_released_callback=None, **kwargs):
-        """Create and instance of a Dicom Listener
+        """Create and instance of a DICOM Listener
 
         Parameters
         ----------
-        storage_directory : [type], optional
+        storage_directory : pathlib.Path/str, optional
```
SimonBiggs, line 35:
```suggestion
        storage_directory : pathlib.Path or str, optional
```

See https://numpydoc.readthedocs.io/en/latest/format.html

SimonBiggs (commented):
Beautiful @pchlap :). This is coming together. Just a few comments here.

lib/pymedphys/tests/dicom/test_connect.py
pchlap: It runs for me locally, however I'm not across poetry enough to understand exactly where we need to be using that... :/

pchlap:
Hi @SimonBiggs @sjswerdloff 

As you can see I have addressed most issues. However there are a couple still outstanding:
- Writing file asynchronously to avoid bottleneck: I've put this in the too hard basket for now, how would I return a status without having actually written the file yet?
- The preamble issue, I think for the purposes of this code we can leave it as is for now? Being aware of this vulnerability is important though and we can factor that in as this codebase develops...?
- LGTM is reporting an unused import since I added the import to DicomListener in: lib/pymedphys/dicom.py to make it available in the library. This is how I plan to use it from my scripts, but perhaps I have misunderstood how we properly manage this?
- I have implemented the storing the conflicting files in an "orphan" directory. Perhaps this could be done better, but my thinking is this can be improved upon as the code develops.

Big thanks again to both of you for the expert review, comments and feedback. I'm learning a lot!

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -234,10 +235,10 @@ def test_dicom_listener_cli(test_dataset):
...
 
-    series_dir = HERE.joinpath(test_dataset.SeriesInstanceUID)
+    series_dir = test_directory.joinpath(test_dataset.SeriesInstanceUID)
     file_path = series_dir.joinpath(f"RP.{test_dataset.SOPInstanceUID}")
     read_dataset = pydicom.read_file(file_path)
     assert read_dataset.SeriesInstanceUID == test_dataset.SeriesInstanceUID
 
     # Clean up after ourselves
-    shutil.rmtree(series_dir)
+    shutil.rmtree(test_directory)
```
SimonBiggs, line 244: I find using temp directory as a context manager makes for really nice clean code and this "rmtree" step is then handled appropriately by the standard library itself:

https://docs.python.org/3/library/tempfile.html#examples

lib/pymedphys/_dicom/connect/listen.py
SimonBiggs:
```suggestion
            filepath, {}, file_meta=meta, preamble=b"\0" * 128
```

SimonBiggs (approved):
Hi @pchlap,

I think everything that wasn't "Nit." from me has been addressed, unless I've missed something? I'm going to mark this as an approval from me. If you could wait until @sjswerdloff also marks his approval, and then after that I am happy if merge when you're ready :)

SimonBiggs:
Hi @pchlap,

I think now, the next thing to do is begin "eating your own dog food" (https://en.wikipedia.org/wiki/Eating_your_own_dog_food).

If you're keen on going down that route, I have just released pymedphys 0.34.0-dev12 based on this branch. So, if you so desire, you can pin `pymedphys==0.34.0-dev12` within PlatiPy's `requirements.txt`. And then undergo the rework to utilise what you've done here within PlatiPy.

I find that sort of work, putting it through its real world paces, highlights a range of issues.

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+            existing_ds = pydicom.read_file(filepath)
+            existing_hash = hash(existing_ds.to_json())
+            incoming_hash = hash(dataset.to_json())
+
+            if not existing_hash == incoming_hash:
+                # If the hashes don't match, save conflicting incoming file in an
+                # "orphan" sub-directory.
+
+                # Just give the file a unique name in the orphan directory
+                filename = str(uuid.uuid4())
```
SimonBiggs, line 141: Could this have the original filename with the random characters appended?

```suggestion
                filename = f"{filename}-{uuid.uuid4()}"
```

If recommendation above to include extension is undergone, would need:

```suggestion
                filename = pathlib.Path(
                    f"{filename.stem}-{uuid.uuid4()}"
                )
```

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+            mode_prefix = DICOM_SOP_CLASS_NAMES_MODE_PREFIXES[dataset.SOPClassUID.name]
+        except KeyError:
+            mode_prefix = "UN"
+
+        series_uid = dataset.SeriesInstanceUID
+        series_dir = self.storage_directory.joinpath(series_uid)
+        series_dir.mkdir(exist_ok=True)
+        self.association_directory = series_dir
+
+        filename = "{0!s}.{1!s}".format(mode_prefix, dataset.SOPInstanceUID)
```
SimonBiggs, line 122: Include extension?

```suggestion
        filename = pathlib.Path(
            "{0!s}.{1!s}.dcm".format(mode_prefix, dataset.SOPInstanceUID)
        )
```

lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,207 @@
...
+                # If the hashes don't match, save conflicting incoming file in an
+                # "orphan" sub-directory.
+
+                # Just give the file a unique name in the orphan directory
+                filename = str(uuid.uuid4())
+                filepath = series_dir.joinpath("orphan", filename)
+                filepath.parent.mkdir(exist_ok=True)
+
+                logging.warning(
+                    "DICOM file exists, storing in orphan directory: %s", filename
```
SimonBiggs, line 146: Include full filepath?

```suggestion
                    "DICOM file exists, storing in orphan directory: %s", filepath
```

pchlap:
> Hi @pchlap,
> 
> I think now, the next thing to do is begin "eating your own dog food" (https://en.wikipedia.org/wiki/Eating_your_own_dog_food).
> 
> If you're keen on going down that route, I have just released pymedphys 0.34.0-dev12 based on this branch. So, if you so desire, you can pin `pymedphys==0.34.0-dev12` within PlatiPy's `requirements.txt`. And then undergo the rework to utilise what you've done here within PlatiPy.
> 
> I find that sort of work, putting it through its real world paces, highlights a range of issues.

Good idea! Will do :)

lib/pymedphys/_dicom/connect/listen.py
SimonBiggs: In multiple locations, it does seem like the accepted solution here is to write a preamble of 128 0x00 bytes:

* ["And when writing the DICOM instance back out, DicomObjects creates an empty 128 bytes for the preamble."](https://www.medicalconnections.co.uk/kb/DICOM-Preamble-Vulnerability#workstation-reading-applications:~:text=And%20when%20writing%20the%20DICOM%20instance,empty%20128%20bytes%20for%20the%20preamble.)
* https://pydicom.github.io/pydicom/dev/reference/generated/pydicom.filewriter.dcmwrite.html#pydicom.filewriter.dcmwrite:~:text=128%200x00%20bytes
* ["DICOM receivers commonly add a blank preamble (128 bytes of 00H) onto images they receive."](https://www.dicomstandard.org/docs/librariesprovider2/dicomdocuments/wp-cotent/uploads/2019/05/faq-dicom-128-byte-preamble-posted1-1.pdf)

Also, at the end of the day, the goal is to not be the vector by which an attack comes. So when loading DICOM files from disk, replace the preamble with 00H if one is transfering the file elsewhere as a file. If sending via DICOM network transmissions the preamble is not included anyway.

If the user already has malware executing on their machine manipulating DICOM preambles the goal is to not let other users be attacked via our software with that preamble. We can't protect the user themselves, they're already compromised.

Therefore, saving the zero bytes preamble is being DICOM compliant as far as I can tell. And it is not introducing a transmission attack vector.

lib/pymedphys/cli/dicom.py
```diff
@@ -229,3 +231,21 @@ def anonymise(dicom_subparsers):
     parser.set_defaults(func=anonymise_cli)
 
     return parser
+
+
+def listen(dicom_subparsers):
+    parser = dicom_subparsers.add_parser(
+        "listen", help="Start a Dicom listener on the specified port"
```
SimonBiggs, line 238:
```suggestion
        "listen", help="Start a DICOM listener on the specified port"
```

SimonBiggs (approved):
Hi @pchlap,

A few more "Nitpick" and clarifying comments from me when doing another read-through this morning.

Nothing to take away my approval. :).

lib/pymedphys/_dicom/connect/listen.py
SimonBiggs: Another note on this topic. Including no header as opposed to 0's doesn't actually offer any protection. An executable on the user's system can happily take a DICOM file with no preamble and add a preamble to it. So the only difference of no preamble vs preamble I believe is one compliant to the DICOM file spec, and the other isn't. ...so a preamble of all 0's, from what I can tell is not only a good way to go, it appears to be "the" way to go.

sjswerdloff:
re preamble attack vector:
I stand corrected.
And now we have a documented history of having discussed, evaluated, and considered mitigations to a known security risk.

I encourage future documentation regarding the use of the SCP to provide warning to the implementer (and perhaps encourage them to not provide direct file system access to the files by general users)

sjswerdloff (approved):
But it would be nice to accept / commit  the suggestion from @SimonBiggs where he capitlalizes DICOM, because it is an acronym of: "Digital Imaging and Communications in Medicine"

SimonBiggs:
> And now we have a documented history of having discussed, evaluated, and considered mitigations to a known security risk.

Yup, good call :)

pchlap:
@SimonBiggs Any idea why the PreCommit step has started failing here? Did I manage to break this while I was committing some of the suggested changes directly from this pull request?

SimonBiggs:
> @SimonBiggs Any idea why the PreCommit step has started failing here? Did I manage to break this while I was committing some of the suggested changes directly from this pull request?

Hi @pchlap,

Yeah, potentially. Given editing online doesn't undergo "pre-commit" there potentially are some commits that didn't have the run occur properly.

To fix this, on your machine, run:

```
poetry run pre-commit run --all-files
```

And that will sweep over everything and fix them. Then you should just need to commit the local changes. According to the CI, the problematic files are:

```
Fixing /home/runner/work/pymedphys/pymedphys/lib/pymedphys/_dicom/connect/listen.py
Fixing /home/runner/work/pymedphys/pymedphys/lib/pymedphys/tests/e2e/test_cypress.py
Fixing /home/runner/work/pymedphys/pymedphys/lib/pymedphys/tests/dicom/test_connect.py
```

pchlap:
> > @SimonBiggs Any idea why the PreCommit step has started failing here? Did I manage to break this while I was committing some of the suggested changes directly from this pull request?
> 
> Hi @pchlap,
> 
> Yeah, potentially. Given editing online doesn't undergo "pre-commit" there potentially are some commits that didn't have the run occur properly.
> 
> To fix this, on your machine, run:
> 
> ```
> poetry run pre-commit run --all-files
> ```
> 
> And that will sweep over everything and fix them. Then you should just need to commit the local changes. According to the CI, the problematic files are:
> 
> ```
> Fixing /home/runner/work/pymedphys/pymedphys/lib/pymedphys/_dicom/connect/listen.py
> Fixing /home/runner/work/pymedphys/pymedphys/lib/pymedphys/tests/e2e/test_cypress.py
> Fixing /home/runner/work/pymedphys/pymedphys/lib/pymedphys/tests/dicom/test_connect.py
> ```

Nice, that did the job thanks. Note that the isort step was skipping for me. I tricked it into running by commenting out the `files: ^pymedphys.*\.py$` line in .pre-commit-config.yaml... Is there something wrong with my configuration?

SimonBiggs:
> Nice, that did the job thanks. Note that the isort step was skipping for me. I tricked it into running by commenting out the `files: ^pymedphys.*\.py$` line in .pre-commit-config.yaml... Is there something wrong with my configuration?

That's weird. In the main branch it says the following:

https://github.com/pymedphys/pymedphys/blob/6fcecc21d215f2af47572e6250bb8cbc4aff2cf1/.pre-commit-config.yaml#L23

SimonBiggs:
Hi @pchlap,

You've got approval from @sjswerdloff and myself :). Merge away! 🚀

pchlap:
> You've got approval from @sjswerdloff and myself :). Merge away! 🚀

So it's still telling me that `Merging is blocked` since `the base branch restricts merging to authorised users`... Could you double check my permissions for me?

SimonBiggs:
> So it's still telling me that `Merging is blocked` since `the base branch restricts merging to authorised users`... Could you double check my permissions for me?

Thanks for the heads up. I think I've fixed that now. Let me know.

pchlap:
> Thanks for the heads up. I think I've fixed that now. Let me know.

All set! Thanks.

SimonBiggs:
image

image :)

PR #1169, opened 2020-11-10
Fix the live docs

SimonBiggs:


sjswerdloff (approved):

Matthew-Jennings (approved):
LGTM (after testing that build doesn't infinitely loop)

PR #1179, opened 2020-11-12
All changes between 0.33.x and now

SimonBiggs:
PR for reviewing changes undergone since the last version and updating CHANGELOG.md


lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -0,0 +1,208 @@
...
+        return status_ds
+
+
+def listen_cli(args):
+    """Start a DICOM listener from the command line interface
+    """
+
+    # Set log level to debug
+    root = logging.getLogger()
+    root.setLevel(logging.DEBUG)
```
SimonBiggs, line 190: Hi @pchlap,

For a follow up PR, we actually already have code built into the pymedphys CLI that allows logging level setting.

For example `pymedphys --debug dicom listen` will cause the log level to be debug, and it puts it in the user control. As such can we remove this loglevel control here?

sjswerdloff (approved):

lib/pymedphys/_utilities/test.py
```diff
@@ -0,0 +1,36 @@
+# Copyright (C) 2020 University of New South Wales & Ingham Institute
```
SimonBiggs, line 1: Hi @pchlap,

I think it's probably making the copyright of this one be in my name if that's okay.

lib/pymedphys/dicom.py
```diff
@@ -4,6 +4,7 @@
 # pylint: disable = unused-import
 
 from ._dicom.anonymise import anonymise_dataset as anonymise
+from ._dicom.connect.listen import DicomListener
```
SimonBiggs, line 7: Hi @pchlap,

I actually missed that this was placed in the public API. That means that every attribute and method, and every function input variable name on that class that isn't proceeded with an `_` would become a part of PyMedPhys' public API. In the case of `DicomListener` that means all of the following functions and attributes can result in a breaking change if they need to be adjusted:

```python
~/git/pymedphys$ poetry run ipython
Python 3.8.6 (default, Nov  1 2020, 14:26:05) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.16.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from pymedphys.dicom import DicomListener                                                                                              

In [2]: dicom_listener = DicomListener()                                                                                                       

In [3]: [item for item in dir(dicom_listener) if not item.startswith('_')]                                                                     
Out[3]: 
['ae',
 'ae_title',
 'association_directory',
 'host',
 'on_association_accepted',
 'on_association_released',
 'on_c_echo',
 'on_c_store',
 'on_released_callback',
 'port',
 'start',
 'stop',
 'storage_directory']
```

Before we release version 0.34.0 would we be able to remove this line here from the exposed API? And then, in a future PR (doesn't have to be before 0.34.0) if we want to expose it via the API might we be able to expose it as a single function which internally boots up the class? That makes our API surface area significantly smaller and controllable.

pchlap:
Hi @SimonBiggs 
Just a blanket answer to all your comments in here: That's all fine.

I'll make the adjustments and put it through a separate pull request? Or whats the practice for small changes like this?

pchlap:
> I'll make the adjustments and put it through a separate pull request? Or whats the practice for small changes like this?

I put the fixes into a pull request: #1183

PR #1180, opened 2020-11-12
Add redirect for app.pymedphys.com

SimonBiggs:
Make it so that the website URL we share far and wide is https://app.pymedphys.com and then, if we ever change how we are hosting the online app, we can just change where this URL redirects to.


netlify.toml
```diff
@@ -38,6 +38,12 @@
   status = 301
   force = true
 
+[[redirects]]
+  from = "https://app.pymedphys.com/*"
+  to = "https://share.streamlit.io/pymedphys/pymedphys/main/app.py/:splat"
+  status = 301
```
SimonBiggs, line 44:
```suggestion
  status = 302
```

PR #1183, opened 2020-11-12
Minor DICOM connect fixes

pchlap:
Here are those minor fixes you flagged in #1179

### DIFF
--- a/lib/pymedphys/dicom.py
+++ b/lib/pymedphys/dicom.py
@@ -4,7 +4,6 @@
 # pylint: disable = unused-import
 
 from ._dicom.anonymise import anonymise_dataset as anonymise
-from ._dicom.connect.listen import DicomListener
 from ._dicom.dose import (
     depth_dose,
     dicom_dose_interpolate,
--- a/lib/pymedphys/_dicom/connect/listen.py
+++ b/lib/pymedphys/_dicom/connect/listen.py
@@ -185,10 +185,6 @@ def listen_cli(args):
     """Start a DICOM listener from the command line interface
     """
 
-    # Set log level to debug
-    root = logging.getLogger()
-    root.setLevel(logging.DEBUG)
-
     # Start the listener
     dicom_listener = DicomListener(
         port=args.port, ae_title=args.aetitle, storage_directory=args.storage_directory
--- a/lib/pymedphys/_utilities/test.py
+++ b/lib/pymedphys/_utilities/test.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2020 University of New South Wales & Ingham Institute
+# Copyright (C) 2020 Simon Biggs
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.


lib/pymedphys/dicom.py
```diff
@@ -4,7 +4,6 @@
 # pylint: disable = unused-import
 
 from ._dicom.anonymise import anonymise_dataset as anonymise
-from ._dicom.connect.listen import DicomListener
```
sjswerdloff, line 7: I think a better approach is to make only the listen_cli public, and then expose that.
In other words, rename the other methods by prefixing them with _, and then instead of deleting the line as done here, 
from ._dicom.connect.list import listen_cli

Later, we can decide what interface to expose (there are a variety of architectural approaches to that).

pchlap: What's the thinking behind exposing the listen_cli? Since that would only be called upon from the command line tool, and that can access it via the private API?

sjswerdloff: If someone wants to start an SCP from within their own application, they need to be able to access that programmatically.  They would probably also want to be able to stop the SCP, but that could also just be assumed to take place when the application shut down, so it isn't necessary for a minimum viable API.
Consider an application that has been architected to operate on DICOM data that is on it's local file system.
It doesn't have an SCP, but it's already using PyMedPhys.  So... while the user is running the application, they start up a local SCP.  And then through other means, they transmit DICOM data to the local SCP.  They don't necessarily want people transmitting stuff to their system otherwise (so they don't run it in the background all the time when they aren't using the previously mentioned application).

But frankly, I was thinking that the CLI front end for PyMedPhys would go through the front door (public access to listen_cli).  It might be necessary to expose __init__, for this approach, or perhaps there are other things I've missed or would need to be done differently as well.

However, if the intent is to have the PyMedPhys CLI access DicomListener through a direct import, then just deleting the line above is fine.  I've taken to being careful about direct imports because @SimonBiggs has a fair amount of magic going on involving dynamic import to allow delivery of PyMedPhys without packages that aren't used at runtime for individual scenarios.  I try to take the approach of keeping as many functions private to a module as possible (because I use a docstring generator plug-in, I have taken to starting with them as public while I'm working locally, until I'm reasonably happy with what I have, and add docstring, then I rename with _ prefix to make most functions private, retaining the docstring so the next time I go through the module I've got breadcrumbs even for the private functions).  Relying on private functions, even internally, breaks encapsulation, and that makes for ripple if you decide to refactor if someone (including yourself) uses alot of the private functions instead of having worked on a clean API.

One can certainly treat minimising API for programmatic access as a separate issue and PR.
No one has started programming against DicomListener yet.

SimonBiggs: > Relying on private functions, even internally, breaks encapsulation, and that makes for ripple if you decide to refactor if someone (including yourself) uses alot of the private functions instead of having worked on a clean API.

I think this is a key point here, and it has significant merit. I'd be keen to discuss this in a separate PR with intent to work out what to expose and how.

SimonBiggs (approved):
Thanks @pchlap :)

PR #1188, opened 2020-11-13
A couple of missing items from the changelog!

SimonBiggs:


sjswerdloff (approved):
are these both being maintained manually or is one copied to the other as part of the release process or build or some kind of hook?

SimonBiggs:
> are these both being maintained manually or is one copied to the other as part of the release process or build or some kind of hook?

```
pymedphys dev propagate
```

Copies one to the other (and adds a header warning not to edit the copy). And CI flags if the propagation hasn't occured.

PR #1192, opened 2020-11-14
Create Binary CI/CD

SimonBiggs:
Alrighty! So we have the binary being built within the CI, and being uploaded as an artifact:

https://github.com/pymedphys/pymedphys/suites/1503639809/artifacts/26311199

I managed to pull the file size down to ~140 MB (used LZMA compression instead of zip).

~~I haven't yet got the cypress tests running against the built exe, that can be another day.~~

I have set it up so that the deployed `.bat` script can run `pymedphys dev tests`. It is now successfully working with `pymedphys dev tests --cypress`. There's an issue with pylinac making the CI not work in this case though :/ so for now I'm not calling `pymedphys dev tests` on the binary within the CI.

~~Now I just need to troubleshoot the standard Windows build...~~

Windows build passing!

~~Now I just need to add docstrings around the place so it is a little easier to review.~~ This is the culmination of a lot of trial and error however... hopefully the proof is in [the pudding](https://github.com/pymedphys/pymedphys/suites/1503639809/artifacts/26311199) though...

Alright, hopefully now it's a bit easier to review :). Let me know if you'd like further clarification anywhere.

Thanks @sjswerdloff! :)


binary/build-binary.py
```diff
@@ -0,0 +1,256 @@
...
+    _create_compressed_python_embed()
+
+    _run_pyinstaller_to_build_the_exe(prepend, append, one_file_mode)
+
+
+def _linux_and_windows_support():
+    """Shimming to allow for building the exe on Linux with wine as well as on Windows.
+
+    If the OS isn't Windows all python calls a prepended with ``wine``.
+    Also, if it is being built for Linux --onefile mode is not used due
```
sjswerdloff, line 81: the CI binary task is excluded for linux and MacOS.
Is the implication here that there is partial progress towards building a linux binary?
Or does "being built for Linux" mean: being built on Linux for MS Windows
Or "being built for Linux assuming that Wine is available on the linux host"
Is this going to work for MacOS?  Apparently Wine is available for MacOS, but MacOS isn't Linux.
It is a *nix, but it isn't Linux.

binary/pymedphys.bat
```diff
@@ -0,0 +1,18 @@
...
+@REM     http://www.apache.org/licenses/LICENSE-2.0
+
+@REM Unless required by applicable law or agreed to in writing, software
+@REM distributed under the License is distributed on an "AS IS" BASIS,
+@REM WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+@REM See the License for the specific language governing permissions and
+@REM limitations under the License.
+
+
+"%~dp0\python-embed\python" -m pymedphys %*
```
sjswerdloff, line 18: %~dp0 is the folder in which the executing bat file resides.
%* is the set of arguments to the .bat file/script.

So it's really important that people not copy this file somewhere else.
Or if they create a link to it, that they be careful about setting the working directory.
One of the first things an MS Windows user will do is create an icon/link on their desktop, so it's important that the requirements on this be documented somewhere, and "in the .bat file itself" is a pretty good place to put a copy of that documentation

binary/test.py
```diff
@@ -0,0 +1,119 @@
...
+
+BUILD = HERE.joinpath("build")
+BUILD_DIST = BUILD.joinpath("dist")
+
+REPO_ROOT = HERE.parent
+CYPRESS = REPO_ROOT.joinpath("lib", "pymedphys", "tests", "e2e")
+
+
+def main():
+    if sys.platform == "win32":
```
sjswerdloff, line 35: sys.platform is specified as a compiler define during the build configuration.
os.name checks whether certain os specific modules are available (e.g. posix, nt, ...)
platform.system() actually runs uname and potentially several other functions to determine the system type at run time.

is using sys.platform the most appropriate approach?
Is win32 a safe bet (will some systems come up with win64 or amd64?)?

binary/pymedphys.bat
SimonBiggs: One way the people can use this anywhere is to add the directory it is in to the users PATH, that way they can call `pymedphys dicom listen` for example in any directory, and because the bat file is in their path, it should just work.

That detail would be quite helpful in the batfile itself :). I'll do that.

Alternatively, can you think of a better way to expose the pymedphys CLI?

binary/build-binary.py
SimonBiggs: > Or does "being built for Linux" mean: being built on Linux for MS Windows

That one :).

This executable building is to build a Windows exe, so it's the only CI that should be doing it. Many things in this script are Windows only. At the end of the day, the "wine" is just so that I could "cross-compile". Don't have Windows at home, so wanted to be able to test that it was going to work in the CI without having to wait for the CI.

It's true that MacOS may work. And yes, the should be "built for testing on Linux"

```suggestion
    Also, if it is being built for testing on Linux --onefile mode is not used due
```

binary/test.py
SimonBiggs: See https://docs.python.org/3/library/sys.html#sys.platform

Windows will always return `win32`. Arguably however `platform.system` is a nicer, easier to understand response. Usage of `sys.platform` is more knee-jerk habit at play.

A google search doesn't disagree with my usage here:

https://stackoverflow.com/a/11674977/3912576

lib/pymedphys/tests/dicom/test_anonymise.py
```diff
@@ -458,6 +459,11 @@ def test_anonymise_cli(tmp_path):
 
 def _test_anonymise_cli_for_file(tmp_path, test_file_path):
     temp_filepath = str(tmp_path / "test.dcm")
+
+    python_executable = pmp_test_utils.get_executable_even_when_embedded()
+
+    dicom_anonymise_cli = [python_executable, "-m", "pymedphys", "dicom", "anonymise"]
```
sjswerdloff, line 465: dicom_anonymise_cli
inconsistent with 
def get_pymedphys_dicom_cli():

approach
DRY, but if you are going to, at least do so consistently.

suggest lifting 
def get_pymedphys_dicom_cli():

to a shared module

lib/pymedphys/tests/dicom/test_header_tweaks.py
```diff
@@ -41,6 +42,11 @@
 )
 
 
+def get_pymedphys_dicom_cli():
```
sjswerdloff, line 45: DRY

lib/pymedphys/tests/dicom/test_anonymise.py
SimonBiggs: The idea here was to try and touch the tests as little as possible while making it so that `pymedphys` could be called with `python -m pymedphys` instead.

...need to do that in order to be able to get `pymedphys.bat dev tests` to run.

This was just a way to propagate that change throughout the tests here. Personally, there's a huge amount I'd do to this file, the biggest of which is actually using python's tempfile logic. For now, I'd rather leave this for what it was (putting out a fire), as opposed to a clean up. For a clean up here is a lot of work.

lib/pymedphys/tests/dicom/test_header_tweaks.py
SimonBiggs: What do you mean here sorry?

lib/pymedphys/tests/dicom/test_header_tweaks.py
```diff
@@ -81,7 +87,8 @@ def test_adjust_machine_name():
     assert adjusted_ds != original_ds
     assert adjusted_ds == expected_ds
 
-    command = "pymedphys dicom adjust-machine-name".split() + [
+    command = get_pymedphys_dicom_cli() + [
```
sjswerdloff, line 90: ok, so on top of get_pymedphys_dicom_cli() showing up in a number of places, you are calling it over and over. 
Which suggests that you should have another function that already knows it needs to get that (suggesting lrucache for get_pymedphys_dicom_cli()), and this other function should build a dicom_cli based command with some arguments (which appear to typically be a list of strings)
so something like:
build_dicom_cli_command( *args) or build_dicom_cli_command( my_args: list)

SimonBiggs: Does this comment address your concern?

https://github.com/pymedphys/pymedphys/pull/1192#discussion_r523739729

sjswerdloff (changes requested):
- [x] Address DRY
- [x] either document the portability issue of the .bat file in the .bat file, document it elsewhere, or substantiate why it isn't necessary to document it.

SimonBiggs:
> Address DRY

Let me know if this discussion addresses your request:

https://github.com/pymedphys/pymedphys/pull/1192#discussion_r523739729

lib/pymedphys/tests/dicom/test_header_tweaks.py
sjswerdloff: this function is being placed in multiple separate files.
So I'm asking that that you not repeat the definition, and instead put it in one place, import from that, and invoke it.

SimonBiggs: One is `get_pymedphys_dicom_cli`, the other `get_pymedphys_dicom_anonymise_cli`, the purpose of both is more of a file specific macro to include `python -m pymedphys` at the beginning of the call.

SimonBiggs: Nevermind :), it is in two places :) I can pull it out.

lib/pymedphys/tests/dicom/test_anonymise.py
sjswerdloff: OK.
How about creating one or more issues and reference the files (tests) where this technical debt is being accumulated.
We can go through the issues later and notify the owners of the tests.
It would be best if the centralized function was created first, and then the owners notified that they can make use of those to retire the technical debt.
But the owners could turn around and say "I didn't add that code, why am I being asked to refactor it".

So the minimum is to just point out in an issue that this is something worth addressing at some point.
And clean up can be done a bit at a time.  Otherwise you end up with large PRs.  A single file being extensively modified isn't too bad.  But a single file having a few PRs where each PR addresses a particular cleanup is not so bad.

lib/pymedphys/tests/dicom/test_header_tweaks.py
sjswerdloff: If an issue is created to identify the need for a command construction function that uses get_pymedphys_dicom_cli() internally, then my concern here is addressed.

SimonBiggs:
Thanks for this review @sjswerdloff :)

It was another big one... I did try to make it as nice as I could though.

Let me know if the following changes since your review address your requests:

[`55a75cd...3d6877e`#diff-f21535d003](https://github.com/pymedphys/pymedphys/compare/55a75cd...3d6877e#diff-f21535d00346f8b22020b1c26084d38e903408d05c10cdde7aa1bde5a5ed7f29)

In particular I have added documentation into the `.bat` file.

sjswerdloff (approved):

lib/pymedphys/tests/dicom/test_anonymise.py
SimonBiggs: I partly missed the tree for the forest here in your comment. I think in the end I was able to address the heart of what you were asking. Not exactly sure what to write in an issue for the rest of this though.

SimonBiggs:
Thanks @sjswerdloff :),

I'm a little bit excited about this one. Now, if the GUI get's prettied up a bit and a config GUI gets made, we're on the home stretch to users being able to pick up said GUI and use it in their own clinic, with little-to-no python experience. (maybe) ... let's see.

OneYoungLion:
Hi Simon! I'm sorry if this is not the proper place, I'm still figuring put how GitHub works :) Would you please help me understand how you make a self-contained exe that runs streamlit without the need for Python or streamlit installation?

SimonBiggs:
> Hi Simon! I'm sorry if this is not the proper place, I'm still figuring put how GitHub works :) Would you please help me understand how you make a self-contained exe that runs streamlit without the need for Python or streamlit installation?

Probably not the best place :) Post over on the Streamlit forum and ping me. Aiming for the highest number of relevant people seeing it in order to best leverage time spent for helping the most people.

PR #1199, opened 2020-11-15
[RFR] Making username and password entry for Mosaiq more user friendly

SimonBiggs:
Unfortunately, don't yet have a mock MSSQL Mosaiq database for testing.

### DIFF
--- a/lib/pymedphys/_streamlit/apps/dashboard.py
+++ b/lib/pymedphys/_streamlit/apps/dashboard.py
@@ -23,6 +23,8 @@ from pymedphys._streamlit.utilities import mosaiq as st_mosaiq
 
 
 def main():
+    st.write("# Mosaiq QCL Dashboard")
+
     centres = ["rccc", "nbcc", "sash"]
     servers = {
         "rccc": "msqsql",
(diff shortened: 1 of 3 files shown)


lib/pymedphys/_streamlit/apps/dashboard.py
```diff
@@ -23,7 +23,10 @@
 
 
 def main():
+    st.write("# Mosaiq QCL Dashboard")
+
     centres = ["rccc", "nbcc", "sash"]
```
sjswerdloff, line 28: this will be much more broadly applicable if the centres and servers are read from configuration (falling back to these defaults is fine to save your own facility from having to do the configuration).

SimonBiggs: It is currently unusable at other centres since it is not leaning on configuration. Just not as high priority as other items.

SimonBiggs: Okay to approve given the scope of this PR?

sjswerdloff (approved):

PR #1202, opened 2020-11-16
[0.1] Make Streamlit watch whole library

SimonBiggs:
The goal here is to boot up Streamlit from the root of the library, that way Streamlit will watch and refresh for the whole library by default.

This code here makes the iteration loop for developing the WLutz code manageable. So I'll need to stack that work into this one.


lib/pymedphys/_streamlit/apps/metersetmap/_trf.py
```diff
@@ -41,9 +41,8 @@ def _get_mosaiq_configuration(headers):
     return machine_centre_map, mosaiq_details, mosaiq_servers
 
 
-@st.cache
 def get_logfile_mosaiq_info(
-    headers, machine_centre_map, mosaiq_details, mosaiq_servers
+    headers: "pd.DataFrame", machine_centre_map, mosaiq_details, mosaiq_servers
```
sjswerdloff, line 45: I wasn't aware you could quote a type hint.

sjswerdloff (approved):

lib/pymedphys/_streamlit/apps/metersetmap/_trf.py
SimonBiggs: Yup, that means you don't need pd to be defined at import time :). Win :)

sjswerdloff: That is huge!
Especially given the somewhat tricky approach to import in PyMedPhys.
Ok, so now I can use type hinting more easily and extensively.

SimonBiggs: The only gotchya, is pylint gets a little confused. For example:

```python
import pathlib  # <-- pylint complains of 'unused import'... just need to override.

def a_function(a_variable: "pathlib.Path"):
    do_something()
```

SimonBiggs: > Ok, so now I can use type hinting more easily and extensively.

As we begin to use it more, we should roll out a mypy check within the CI also.

PR #1206, opened 2020-11-17
Remove special MacOS install instructions

SimonBiggs:
`pymssql` has a new maintainer. They've produced MacOS wheels! :) Yay. So No longer will I need special macos install instructions.

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -200,11 +200,6 @@ jobs:
 
 # System installs
 
-    - name: Install FreeTDS on MacOS
-      if: matrix.os == 'macos'
-      run: |
-        brew install freetds
-
     - name: Install Pandoc on Ubuntu
       if: ( matrix.task == 'docs' || matrix.task == 'updates' ) && matrix.os == 'ubuntu'
       run: |
--- a/lib/pymedphys/docs/howto/get-started.rst
+++ b/lib/pymedphys/docs/howto/get-started.rst
@@ -12,8 +12,8 @@ Install Python
 
 In order to make use of the PyMedPhys library, you'll need Python installed on
 your workstation. On Windows we recommend that you install the
-Anaconda Python distribution. Download the latest Anaconda **Python 3** (not 2)
-version from `here <https://www.anaconda.com/download/>`__.
+Anaconda Python distribution. Download the latest Anaconda
+version from `here <https://www.anaconda.com/products/individual#Downloads>`__.
 
 .. note::
 
@@ -44,8 +44,8 @@ Python. The ``[user]`` option is needed to install pymedphys with its
 dependencies which you may need during your use of pymedphys.
 
 
-A minimal installation
-----------------------
+A minimal installation (Advanced users only)
+--------------------------------------------
 
 If you're not interested in installing PyMedPhys' dependencies you can choose
 to skip the ``[user]`` option as so:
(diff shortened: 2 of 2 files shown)


sjswerdloff (approved):

sjswerdloff:
has the version for pymssql been set so the wheel will be available?

SimonBiggs:
> has the version for pymssql been set so the wheel will be available?

Given our software can work with previous versions it is not a library's responsibility (in fact they certainly shouldn't) lock out versions that it is possible to work with. Package managers will at times refuse to install software if a non-conflicting version pinning cannot be found. If we were using a feature of pymssql that required the latest version, then we'd pin it. If we were utilising private APIs or unstable APIs that provide no stability guarantee, then we should also pin it. Otherwise not.

SimonBiggs:
> > has the version for pymssql been set so the wheel will be available?
> 
> Given our software can work with previous versions it is not a library's responsibility (in fact they certainly shouldn't) lock out versions that it is possible to work with. Package managers will at times refuse to install software if a non-conflicting version pinning cannot be found. If we were using a feature of pymssql that required the latest version, then we'd pin it. If we were utilising private APIs or unstable APIs that provide no stability guarantee, then we should also pin it. Otherwise not.

In the `requirements.txt` file however, and in the `poetry.lock` files we have pinned it:

https://github.com/pymedphys/pymedphys/blob/a42ac76680226f9e432af7c81d76658a376930ce/requirements.txt#L91

Just not the pyproject.toml (or equivalently the setup.py)

sjswerdloff:
> > > has the version for pymssql been set so the wheel will be available?
> > 
> > 
> > Given our software can work with previous versions it is not a library's responsibility (in fact they certainly shouldn't) lock out versions that it is possible to work with. Package managers will at times refuse to install software if a non-conflicting version pinning cannot be found. If we were using a feature of pymssql that required the latest version, then we'd pin it. If we were utilising private APIs or unstable APIs that provide no stability guarantee, then we should also pin it. Otherwise not.
> 
> In the `requirements.txt` file however, and in the `poetry.lock` files we have pinned it:
> 
> https://github.com/pymedphys/pymedphys/blob/a42ac76680226f9e432af7c81d76658a376930ce/requirements.txt#L91
> 
> Just not the pyproject.toml (or equivalently the setup.py)

If we delete instructions for how to install the earlier versions, then we are not really sypporting those ("possible to work with, but you'll have to figure out how to get it installed, and it used to be an N step process, so... best of luck").  But if it's in the poetry.lock and requirements.txt, then that's really what a user who "just wants to install it and get on with things" will get, and that's fine.

SimonBiggs:
If someone runs `pip install pymedphys[user]` and they already have pymssql on their machine, they will keep their old version. But that means they have successfully installed it in the past. If they install it today without pymssql on their machine then it will get the latest version unless another one of our dependencies explicitly pins to not use the latest version.

SimonBiggs:
If it helps, for an authoritative instruction on the topic, see:

> It is not considered best practice to use `install_requires` to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades.

-- https://packaging.python.org/discussions/install-requires-vs-requirements/#install-requires-vs-requirements-files

> Whereas `install_requires` requirements are minimal, requirements files often contain an exhaustive listing of pinned versions for the purpose of achieving repeatable installations of a complete environment.

-- https://packaging.python.org/discussions/install-requires-vs-requirements/#requirements-files

PR #1210, opened 2020-11-18
[STACK 0.1] Use editable install on streamlit

SimonBiggs:
This was also sort of "putting out a fire". I went to `app.pymedphys.com` and the server was broken because the following file changed:

https://github.com/pymedphys/pymedphys/blob/a42ac76680226f9e432af7c81d76658a376930ce/app.py#L1-L31

But the streamlit server wasn't rebuilt, and as a result it was using an out of date pymedphys install. So, the requirements.txt install needs to be an "editable" version. "editable" installs of dependencies are only supported when a setup.py exists, hence bringing that back.


SimonBiggs:
At the end of the day, this is a symptom of not have a CI process that calls the live version of https://app.pymedphys.com.

lib/pymedphys/_dev/propagate.py
```diff
@@ -139,15 +204,21 @@ def propagate_requirements():
         "poetry export --without-hashes -E docs -E user -f requirements.txt --output requirements.txt",
         shell=True,
     )
+
+    # The editable install `-e` is used here so that https://app.pymedphys.com
+    # is always utilising the latest pymedphys build on main without needing
+    # a server rebuild. It also means that should a user edit the git repo
+    # they will be utilising their edits within their environment as opposed
```
sjswerdloff, line 211: I think that's a plus for developers

sjswerdloff (approved):

lib/pymedphys/_dev/propagate.py
SimonBiggs: Yup, I agree.

PR #1217, opened 2020-11-20
[RFR] Enable the library tests for the binary

SimonBiggs:
Well that was unexpected. I was able to get the binary's tests to pass by using a recent commit hash of pylinac. Pylinac recently removed the `py_gui` import from its `__ini__.py` and that's now allowed us to utilise it here.

Hi @pchlap,

To make the pinnacle CLI tests work I needed to prepend the equivalent of `python -m` in front of `pymedphys experimental pinnacle ...etc`.

Would you be able to give this a bit of a look over and make sure you're happy with it? It's a little messy the way I adjusted it, if you have a preference for how to make it work let me know.

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -194,7 +194,7 @@ jobs:
       if: steps.poetry-cache.outputs.cache-hit == 'true'
       shell: bash
       run: |
-        poetry run pip --version >/dev/null 2>&1 || rm -rf .venv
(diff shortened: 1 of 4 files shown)


pchlap (approved):
Hi @SimonBiggs 
This looks fine to me. No doubt we could refactor out a function to help build the CLI command there but I don't think thats necessary for these experimental CLI tests. I can do that when/if I continue some work on this code.

PR #1239, opened 2020-12-01
[RFR] Add Discourse commenting directly within the docs

SimonBiggs:
I'll fix the failing tests before merging. Given this isn't actually submitting anything to the library itself it doesn't necessarily need the same level of scrutiny, but it's helpful to have you abreast of what I'm doing on the docs side of things.


SimonBiggs:
Added a comments section at the bottom of each documentation page:

image

sjswerdloff (approved):
once the tests pass.
There was a botocore issue here as well, so odds are a lot of the problems were timeouts
I kicked off a rerun of jobs.
But it also looks like a lot of jobs are duplicates, so maybe something weird there?

PR #1240, opened 2020-12-01
Comparison to flash gamma

SimonBiggs:
See https://github.com/samuelpeet/flashgamma/issues/7


examples/drafts/gamma/compare_to_flash_gamma.ipynb
```diff
@@ -0,0 +1,343 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "plot_passing_rates(pass_rates)"
+   ]
+  },
+  {
```
sjswerdloff, line 277: might be nice to have additional markdown here to separate the pymedphys calculation from the plotting of the FlashGamma pass rates

setup.py
```diff
@@ -185,6 +185,7 @@
     "dicom": ["pydicom>=2.0.0", "pynetdicom", "pylibjpeg-libjpeg"],
     "docs": ["jupyter-book>=0.8.3", "sphinx-argparse", "sphinxcontrib-napoleon"],
     "doctests": ["tensorflow>=2.2.0", "black>=19.3b0,<20.0", "tomlkit"],
+    "external": ["flashgamma"],
```
sjswerdloff, line 188: external is a very broad term.
in the property  (land/housing) world, this kind of thing is referred to as "comparables" or "comps".

sjswerdloff (approved):
assuming botocore issue can be resolved...
Might be a good time to update that to 1.19.27 (latest), unless there is a conflict

setup.py
SimonBiggs: > external is a very broad term.

Yeah, good call.

> in the property (land/housing) world, this kind of thing is referred to as "comparables" or "comps".

I like it :) Will make the change.

SimonBiggs:
@sjswerdloff do you know what happened here? If you fixed it, thanks :)

sjswerdloff:
Re-ran jobs.
It smelled like a timeout

examples/drafts/gamma/compare_to_flash_gamma.ipynb
SimonBiggs: Thanks @sjswerdloff, I cleaned this up a bit :)

Issue #1246, opened 2020-12-04
pseudonymisation fails on DA or DT with embedded UTC (e.g. "20201201093456+1100")


sjswerdloff:
found by OnkoDICOM

PR #1247, opened 2020-12-04
parse using pydicom.valuerep.DT

sjswerdloff:
combine with tzinfo from passed in value even if None to ensure operating on datetime.datetime objects
DT has problems with addition of timedelta or at least it did in initial testing.

See #1246

### DIFF
--- a/lib/pymedphys/_experimental/pseudonymisation/strategy.py
+++ b/lib/pymedphys/_experimental/pseudonymisation/strategy.py
@@ -159,34 +159,36 @@ def _pseudonymise_CS(value):
     return my_sliced_pseudonym
 
 
+def _add_tzinfo(d: datetime.datetime, tz: datetime.timezone) -> datetime.datetime:
+    return datetime.datetime.combine(d.date(), d.time(), tz)
+
+
 def _pseudonymise_DA(
     value, format_str=DICOM_DATE_FORMAT_STR, earliest_study=DEFAULT_EARLIEST_STUDY_DATE
 ):
     epoch_start = EPOCH_START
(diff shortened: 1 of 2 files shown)


SimonBiggs (approved):
LGTM 🙂 

Were you interested in getting these bugfixes released sooner rather than later?

An option that shouldn't be too hard would be to make a dev release for trialing within onkodicom?

Release process for a dev release is as follows:

```
# Step 1: edit version number in pyproject.toml only

# 2 - 4:

poetry build
poetry deploy
```

sjswerdloff:
I have started looking at hypothesise, but it will take some further analysis on my part to see if it's actually useful for my purposes.  I need to generate DICOM conformant values for each of the VR and iterate against the methods for each VR with the variety of values generated.  It's kind of like fuzzying the input, but in a very constrained fashion.

sjswerdloff:
> LGTM 🙂
> 
> Were you interested in getting these bugfixes released sooner rather than later?
> 
> An option that shouldn't be too hard would be to make a dev release for trialing within onkodicom?
> 
> Release process for a dev release is as follows:
> 
> ```
> # Step 1: edit version number in pyproject.toml only
> 
> # 2 - 4:
> poetry run pymedphys dev propagate
> 
> poetry build
> poetry deploy
> ```

Yes.  I'll get started on it...
I'll ask for help if I bump in to problems...

sjswerdloff:
@SimonBiggs 
when I type in
`poetry deploy`
I get:

`The command "deploy" is not defined.`

should I be using 
`poetry publish`

SimonBiggs:
Yup, publish, that's the one. Sorry got confused with yarn's (?) CLI.

PR #1253, opened 2020-12-07
[STACK 0.1] Move to experimental

SimonBiggs:
Hi @sjswerdloff,

This is the first PR of attempting to merge in some of the work from `wlutz-arc`. This PR is aiming to start simple. It is moving all current "experimental apps" into the "experimental directory".

PR #1254, opened 2020-12-07
[STACK 0.2] Add experimental to isort

SimonBiggs:
I changed the isort settings so that experimental imports are placed below normal pymedphys imports flagging their hierarchy with respect to the rest of the library.


pyproject.toml
```diff
@@ -141,9 +141,10 @@ known_medphys = [
     "pydicom", "pylinac", "pynetdicom", "libjpeg"
 ]
 known_first_party = ["pymedphys"]
+known_experimental = ["pymedphys._experimental"]
 sections = [
     "FUTURE", "STDLIB", "THIRDPARTY", "GENERAL_SCIENTIFIC", "MACHINE_LEARNING",
-    "MEDPHYS", "FIRSTPARTY", "LOCALFOLDER"
+    "MEDPHYS", "FIRSTPARTY", "EXPERIMENTAL", "LOCALFOLDER"
```
SimonBiggs, line 147: This section here is the only piece that was actually updated. Everything else came from running `poetry run pre-commit run --all-files`.

sjswerdloff (approved):

sjswerdloff:
it looks like some modules need to be committed/moved...

SimonBiggs:
> it looks like some modules need to be committed/moved...

Yeah, oops. Fixed that in https://github.com/pymedphys/pymedphys/pull/1253.

SimonBiggs:
> it looks like some modules need to be committed/moved...

A better practice would be to only mark for review once I see that the CI tests pass.

sjswerdloff:
LGTM 👍

PR #1257, opened 2020-12-07
Move tensorflow out of `_imports` given its own import magic

SimonBiggs:
Tensorflow importing would bypass pymedphys' `just-in-time` import logic. This would result in slow initial import times across the board. This moves tensorflow into its own `pymedphys._imports.slow` package.

### DIFF
--- a/lib/pymedphys/_experimental/autosegmentation/pipeline.py
+++ b/lib/pymedphys/_experimental/autosegmentation/pipeline.py
@@ -19,7 +19,7 @@ import pathlib
 
 from pymedphys._imports import numpy as np
 from pymedphys._imports import pydicom
-from pymedphys._imports import tensorflow as tf
+from pymedphys._imports.slow import tensorflow as tf
 
 from pymedphys._data import download
 
--- a/lib/pymedphys/_imports/__init__.py
+++ b/lib/pymedphys/_imports/__init__.py
@@ -7,7 +7,7 @@ from pymedphys._vendor import apipkg
 
 HERE = pathlib.Path(__file__).parent
 
-imports_for_apipkg = _parse.parse_imports(HERE)
+imports_for_apipkg = _parse.parse_imports(HERE.joinpath("imports.py"))
 apipkg.initpkg(__name__, imports_for_apipkg)  # type: ignore
 
 THIS = importlib.import_module(__name__)
--- a/lib/pymedphys/_imports/__init__.py
+++ b/lib/pymedphys/_imports/__init__.py
@@ -1,10 +1,17 @@
+import importlib
 import pathlib
 
-from pymedphys._imports import _magic
+from pymedphys._imports import _parse
+
+from pymedphys._vendor import apipkg
 
 HERE = pathlib.Path(__file__).parent
 
-IMPORTABLES = _magic.import_magic(HERE)
+imports_for_apipkg = _parse.parse_imports(HERE)
+apipkg.initpkg(__name__, imports_for_apipkg)  # type: ignore
+
+THIS = importlib.import_module(__name__)
+IMPORTABLES = dir(THIS)
 
 # This will never actually run, but it helps pylint know what's going on
 if "numpy" not in IMPORTABLES:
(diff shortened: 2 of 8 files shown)


sjswerdloff:
Why does putting the imports in slow cause different behaviour / outcome?
Is *.slow processed differently?

While I can see code changes that look like a namespace change it is not clear why this changes the outcome.

The purpose is clear but the fundamental principle is not

SimonBiggs:
> The purpose is clear but the fundamental principle is not

Yup, I'll explain below inline.

lib/pymedphys/_experimental/autosegmentation/pipeline.py
```diff
@@ -19,7 +19,7 @@
 
 from pymedphys._imports import numpy as np
 from pymedphys._imports import pydicom
-from pymedphys._imports import tensorflow as tf
+from pymedphys._imports.slow import tensorflow as tf
```
SimonBiggs, line 22: Throughout the code base, to import tensorflow, it now needs to be imported from a separate module. It is no longer included within `pymedphys._imports` instead, it is only included within `pymedphys._imports.slow`

lib/pymedphys/_imports/__init__.py
```diff
@@ -1,34 +1,13 @@
...
-    asname = alias.asname
-
-    if asname is None:
-        asname = alias.name
-
-    imports_for_apipkg[asname] = alias.name
-
-
+imports_for_apipkg = _parse.parse_imports(HERE.joinpath("imports.py"))
 apipkg.initpkg(__name__, imports_for_apipkg)  # type: ignore
```
SimonBiggs, line 11: Importantly, the reason this works, is because this file here, `pymedphys/_imports/__ini__.py` is having its module object (`pymedphys._imports`) be mutated inplace by apipkg.

So, previously, apipkg, would insert the `just-in-time` import wrapper object attached to this module here.

Tensorflow has some of its own magic that runs off and checks for GPUs and a range of other things. Tensorflow's magic would activate the moment any other module was imported from `pymedphys._imports` even if it wasn't `tensorflow`.

lib/pymedphys/_imports/slow/imports.py
```diff
@@ -0,0 +1,3 @@
+# pylint: disable = unused-import, reimported, import-error
+
+import tensorflow
```
SimonBiggs, line 3: The tensorflow import is now within this file.

lib/pymedphys/_imports/slow/__init__.py
```diff
@@ -0,0 +1,20 @@
+import importlib
+import pathlib
+
+from pymedphys._imports import _parse
+
+from pymedphys._vendor import apipkg
+
+HERE = pathlib.Path(__file__).parent
+
+imports_for_apipkg = _parse.parse_imports(HERE.joinpath("imports.py"))
```
SimonBiggs, line 10: Which means it gets included within this dictionary here.

lib/pymedphys/_imports/slow/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+apipkg.initpkg(__name__, imports_for_apipkg)  # type: ignore
```
SimonBiggs, line 11: And then apipkg uses this dictionary to mutate the `pymedphys._imports.slow` module object to contain a `just-in-time` `tensortflow` import wrapper. That way, the initialisation slow down caused by tensorflow will only occur when a module is imported from `pymedphys._imports.slow`.

Importantly, now, there is nothing to do with `tensorflow` importable from the `pymedphys._imports` module object. Instead, now, only the `pymedphys._imports.slow` module object.

sjswerdloff: so this isn't really about something being slow.
It's about tensorflow import behaviour.
what about pymedphys._imports.tensorflow
or
pymedphys._imports.ownmagic

SimonBiggs: Well, it is about it being slow per-say. I don't mind the import magic so much, it's just that it delays every pymedphys CLI command on the order of ~1-2 seconds on some systems (depending on GPUs, OS, and how long it takes tensorflow to do its initialisation).

sjswerdloff (approved):
I'd prefer a package/module name that was more closely tied to the reasoning why someone would put another package in here, but this does the job, and it suggests that either this is for putting packages slow to import or because tensorflow was slow to import.

PR #1258, opened 2020-12-07
[STACK config -> base] Some configuration updates to support WLutz

SimonBiggs:
Added an "icom" entry to the site_directories config wrapper. Added various items to the example `config.toml` file for use within WLutz.


lib/pymedphys/_streamlit/utilities/config.py
```diff
@@ -53,6 +53,15 @@ def get_export_directory_from_site_config(site_config, export_directory):
 
 @st.cache
 def get_site_directories():
+    """A config wrapper that retrieves a dictionary that maps site to directory.
+
+    Returns
```
sjswerdloff, line 58: typing would be very useful here, because reading the name of the function, I would think that I would get a dictionary mapping a site name to a directory.  The docstring clarifies that isn't the case.  But this is in a utilities package, even if it is intended to be internal to pymedphys streamlit app development, so it needs to be quite clear what the developer is going to get back and how to use it.  In particular, until the config.toml is documented so that the "what is getting returned" is documented in that, it's important that what comes back and how it gets used be as clearly defined as possible. 

from typing import Dict

def get_site_directories()->Dict[str, Dict[str, Any]]:

or better yet have Any replaced with the function type reference (I'm not familiar with exactly how that's done, so I punted to Any...).

site-specific/cancer-care-associates/config.toml
```diff
@@ -1,20 +1,40 @@
...
+# to:
+#     {"site": {"rccc": { ... }, "nbcc": { ... } }}
+# The question here, 'is order important'? And maybe order preservation is
+# helpful, in which case it should remain scoped within a list as opposed to
+# a dictionary.
 [[site]]
 name = "rccc"
 
-    # TODO: For config v2 change 'export-directories' to just 'directories'
+    # TODO [v2]: Change 'export-directories' to just 'directories'
```
sjswerdloff, line 31: best to leave inputs and outputs separate.
however, is this a set of directories to which pymedphys exports, or is this a directory to which other applications export, and is therefore considered an import directory for pymedphys?
Will pymedphys have places where it wants to export something as inputs to these applications/protocols?

sjswerdloff (changes requested):
please incorporate typing for get_site_directories()

lib/pymedphys/_streamlit/utilities/config.py
SimonBiggs: Yup, good call. Will do :).

site-specific/cancer-care-associates/config.toml
SimonBiggs: In the case of input/output I was thinking of making that clear within the directories themselves if that was appropriate. Thinking of these config files more as "root directories" for a given scenario. Each application can then run away and make well defined sub directories relative to these roots if need be. What do you think regarding that?

The `icom = '\\rccc-physicssvr\iComLogFiles'` entry is being used within my `wlutz-arc` branch to point to where the iCom log files are being stored. However, it is not unreasonable that the very iCom listener that is listening to the iCom stream and saving these log files (also PyMedPhys) could pull from this config option. So in one sense, for the `pymedphys icom listen` CLI this would be an export directory, for the WLutz Arc GUI it would be an input. ... the concepts of export/import input/output tend to be blurred.

Looking for any further input on this you might have. Thoughts?

lib/pymedphys/_streamlit/utilities/config.py
SimonBiggs: > or better yet have Any replaced with the function type reference (I'm not familiar with exactly how that's done, so I punted to Any...).

So, it took me quite a bit to work out how to do it, but I figure it'll be quite helpful in the future, so it was worth the time investment. But I managed to restrict the output type to something that is more descriptive.

Let me know your thoughts :).

SimonBiggs:
@sjswerdloff I think I've made the changes here you were after. Let me know what you think 🙂

sjswerdloff (approved):

PR #1259, opened 2020-12-07
Update figsize within WLutz reporting

SimonBiggs:
Figsize was updated for it to represent the default column width within streamlit.

### DIFF
--- a/lib/pymedphys/_wlutz/reporting.py
+++ b/lib/pymedphys/_wlutz/reporting.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Tuple
 
 from pymedphys._imports import numpy as np
 from pymedphys._imports import plt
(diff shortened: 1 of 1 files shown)


lib/pymedphys/_wlutz/reporting.py
```diff
@@ -20,17 +21,52 @@
...
+    y: "np.array",
+    img: "np.array",
+    bb_centre: Tuple[float, float],
+    field_centre: Tuple[float, float],
+    field_rotation: float,
+    bb_diameter: float,
+    edge_lengths: Tuple[float, float],
+    penumbra: float,
     units="(mm)",
 ):
```
sjswerdloff, line 34: go the whole way!  Typing for the output (figure and axes...)

sjswerdloff (approved):
yeah!  better documentation and typing

lib/pymedphys/_wlutz/reporting.py
SimonBiggs: Good call. I'll leave this for this one, but will make a habit of it in the future :).

PR #1262, opened 2020-12-08
Changelog writing for 0.35.0

SimonBiggs:
@sjswerdloff if you want to give it a shot feel free to begin committing entries to this changelog here using the utility PR over at https://github.com/pymedphys/pymedphys/pull/1261 as a guide.

I probably won't get to this tonight myself. I'm happy to do this in tandem with you, do it all, or let you do it. Up to you. I'll dive in at some point probably either tomorrow day or evening.


sjswerdloff:
@SimonBiggs leave this to me.  I'll ask for help when I bump in to something and need assistance.
It might take me a few days.

I see the propagate issue with changelog.md which appears to be an automated updated based on release-notes.md
I'm guessing that once the release-notes are complete, I should update pyproject.toml with 0.35.0 (assuming there are either new features or a public interface change, otherwise 0.34.x), do a local propagate, commit that, and then the issue will go away.

No time like the present to handle a maintainer's responsibility.

SimonBiggs:
> I'm guessing that once the release-notes are complete, I should update pyproject.toml with 0.35.0 (assuming there are either new features or a public interface change, otherwise 0.34.x), do a local propagate, commit that, and then the issue will go away.

Yup, perfect. That's the process. :slightly_smiling_face: :rocket:

SimonBiggs:
Once I've released to PyPI I create a `0.35.x` branch with that version, and then I also create a `v0.35.0` tag. In the changelog of the tag I copy in the changelog from the `release-notes.md`. Also, for now, let's not upload the `GUI.exe` file as I'm not yet ready for that to go into propagation. Also, let's not yet mention it in the news.

SimonBiggs:
I might still also do a scan through of the changes in https://github.com/pymedphys/pymedphys/pull/1261. I find it helps me get a periodic birds eye view of the changes. So I'll at least also do that I think.

SimonBiggs:
If the changelog PR has too much noise, the following view is also very helpful:

https://github.com/pymedphys/pymedphys/issues?q=is%3Aclosed+merged%3A%3E%3D2020-11-13+sort%3Acreated-asc

...also, if you change your mind regarding the changelog, I'd be more than happy to pick it up. I don't want to use up all of your juice and leave none left for reviewing/committing.

sjswerdloff:
The closed and merged after "previous release date" PR view is a better overview, I was thinking of creating a view like that, so thank you for providing the query.
I'll be fine.  Just had other things happening this morning.

lib/pymedphys/docs/release-notes.md
```diff
@@ -8,6 +8,87 @@ This project adheres to
...
+
+* Installation on MacOS (Intel) has been simplified and is now the same as for other platforms, thanks to work by the new pymssql maintainers.
+
+### "Stable" API changes
+
+(Won't truly be stable until after a 1.0.0 release)
+
+#### Breaking changes
+
+* DicomListener removed from public API (access is only through CLI)
```
SimonBiggs, line 23: Pretty sure this API never got exposed within a release.

lib/pymedphys/docs/release-notes.md
```diff
@@ -8,6 +8,87 @@ This project adheres to
...
+
+#### Deprecations
+
+
+#### New features
+
+* `pymedphys dicom listen` stores incoming DICOM objects in a directory hierarchy: PatientID/Study Instance UID/Series Instance UID
+aligning with the DICOM Q/R hierarchy, and more suitable for use with tools like dicompyler and OnkoDICOM
```
SimonBiggs, line 31: If this was in regular use, this possibly is a breaking change. I'm not sure if we should consider "how we store data" part of our API... What do you think?

lib/pymedphys/docs/release-notes.md
```diff
@@ -8,6 +8,87 @@ This project adheres to
...
+
+#### Bug fixes
+
+* `pymedphys dicom listen` logging was bypassing CLI directives
```
SimonBiggs, line 35: I don't think this original bug ever made it into the 0.34.0 release either.

lib/pymedphys/docs/release-notes.md
sjswerdloff: You're right.  It was removed prior to 0.34.0
I'll remove that.

lib/pymedphys/docs/release-notes.md
```diff
@@ -8,6 +8,87 @@ This project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [0.35.0]
+
+### News around this release
+
+* Installation on MacOS (Intel) has been simplified and is now the same as for other platforms, thanks to work by the new pymssql maintainers.
```
SimonBiggs, line 15: Move to docs section? I think reserve this news section for the sort of thing we'd be keen on posting to Linkedin and the like. Not sure if there's anything like that for this release?

lib/pymedphys/docs/release-notes.md
sjswerdloff: It isn't a change in the API (Application Programmer Interface) itself, it isn't a change in the CLI either.

We can't say for sure it would be a breaking change.
If someone is looking at output/* then yes.  If they are globbing output/** (maybe output/**/*?) then it isn't.
But it is worth warning the user.

I'm not sure how regular the use could be given that 0.34 was released quite recently.

lib/pymedphys/docs/release-notes.md
sjswerdloff: If I were a Mac user, I'd be interested in knowing I don't have to do a bunch of things other than `pip install pymedphys[user]`

What we might want to do is have a section on installation.
Changing the documentation on installation is documentation.
Changing how one performs the installation is not documentation, it's installation.

But it's a question of what you consider "Newsworthy".

There isn't anything in this release that you want to talk about (e.g. the binary) that is going to be of interest to the casual observer, unless that casual observer is a MacOS user who is thinking of maybe installing PyMedPhys.

For someone who wants to know "should I upgrade", that usually involves reading more deeply (does it fix a bug I'm having issues with, does it improve performance for something that matters to me, is it going to break my current code).

Having a binary is a game changer for many, so when that comes out, it should be in the News.

SimonBiggs: Yes certainly. The binary is something I'd be happy to be in the news. Once it's ready :). Any section that is empty can just be removed. A presentation at a conference, a paper, a new contributor (or new maintainer 🎉🙂) would all count for this section in my opinion.

lib/pymedphys/docs/release-notes.md
SimonBiggs: I guess I was sort of thinking of the file output from the CLI being the "output"/"return" API of the CLI. I think if we go down that path we're going to trigger breaking API changes for all sorts of things. I'm happy to not have it be declared as part of our API.

lib/pymedphys/docs/release-notes.md
```diff
@@ -8,6 +8,85 @@ This project adheres to
...
+#### Logistics changes
+
+
+
+#### Pseudonymise
+
+##### Bug fixes
+
+* Pseudonymisation of Decimal String (e.g. Patient Weight) was failing
+* Pseudonymisation of Date, Time or DateTime elements with embedded UTC offsets would fail.
```
SimonBiggs, line 58: This should probably be scoped under bug fixes within "Experimental API" changes.

lib/pymedphys/docs/release-notes.md
SimonBiggs: > What we might want to do is have a section on installation.

More than happy to have that added in 🙂

lib/pymedphys/docs/release-notes.md
sjswerdloff: Ok.
But in the future, can we simplify the release notes to have just
News
Broken public API
New Features/Enhancements
Bug fixes

and then annotation of some sort regarding whether it is
Documentation
Code
Installation

I'm thinking I'd like to start naming/title PRs so that it can be used almost verbatim as the entry in the release notes.

SimonBiggs: Yup, I like that. That heading structure, the annotation, and having PRs named or tagged to make the task easier all makes a lot of sense.

Towncrier is also an interesting approach to this problem which brings us a step closer to having main always be release ready:

https://github.com/twisted/towncrier

SimonBiggs: > But in the future, can we simplify the release notes to have just ...

I wouldn't be opposed to you making those changes here in this release if you'd prefer.

SimonBiggs:
@sjswerdloff I just wanted to let you know that this release doesn't need to wait on https://github.com/pymedphys/pymedphys/pull/1263. More than happy for you to push ahead and get this release undergone. You are free to wait if you'd like, but I have never made it a thing to be awaiting a certain PR for a release.

And on that note, I can't actually review this PR as it is my own. So I can't press the "approve" tick.

sjswerdloff:
@SimonBiggs I don't understand why the build is now failing, although I do see the error message.
I do understand why the toml got reformatted (I've now removed the Better TOML and "Even Better TOML" extensions from my vscode).

sjswerdloff:
fyi, poetry run pymedphys dev tests and poetry run pymedphys dev docs 
both passed on my local environment (Ubuntu 20.04)

sjswerdloff (approved):

sjswerdloff (approved):

PR #1263, opened 2020-12-09
Automated chart checks

rembishj:
Improvements have been made to the streamlit apps for pre-treatment data transfer checks and weekly chart checks, and they've been added to the experimental gui. 

I have not yet set up a config file, so the networks are hardcoded and you will need to change them accordingly to test at your site.


SimonBiggs:
> Improvements have been made to the streamlit apps for pre-treatment data transfer checks and weekly chart checks, and they've been added to the experimental gui.
> 
> I have not yet set up a config file, so the networks are hardcoded and you will need to change them accordingly to test at your site.

🎈 Awesome that they're all working embedded within the PyMedPhys GUI!! 🎈 🎉 😄 

Given this is (will be) all within experimental this merge should be quite simple. Would you be able to, for now, move your mosaiq helper additions into an experimental module? Otherwise you will be required to go through all sorts of scrutiny to get that merged.

Could you also merge the following two folders into one called "chartchecks" or something similar? Just to reduce the number of folders at the top level of that directory.

image

After you've made those changes I think we'll opt for merging this in and then as you make changes going forward you can keep submitting PRs and I can review the new snippets that come through.

SimonBiggs:
Also, can you run `poetry run pymedphys dev tests --pylint` locally, and the fix up the things that it winges about?

rembishj:
> Also, can you run `poetry run pymedphys dev tests --pylint` locally, and the fix up the things that it winges about?

I've run this and encountered an attribute error with the sip.setapi in the matplotlib backends. From some quick googling, it looks like matplotlib 3.3 doesn't play well with sip.setapi but version 3.2 does so I switched to that version and it works now. Made all but just a few of the suggested changes to my code, only since those might take a little bit more thinking to fix

lib/pymedphys/_mosaiq/helpers.py
```diff
@@ -271,8 +271,8 @@ def get_incomplete_qcls(cursor, location):
         """
         SELECT
             Ident.IDA,
-            Patient.Last_Name,
             Patient.First_Name,
+            Patient.Last_Name,
```
SimonBiggs, line 275: There's still a few changes to this helpers script which could possibly break other tools (it's hard to know without Mosaiq automated testing :/). If these are needed could you undergo them in a separate helpers script within `_experimental`?

pyproject.toml
```diff
@@ -43,7 +43,7 @@ PyYAML = {version = "*", optional = true}  # groups = ["user", "dev"]
 requests = {version = "*", optional = true}  # groups = ["user", "dev"]
 
 numpy = {version = ">= 1.12", optional = true}  # groups = ["user", "dev"]
-matplotlib = {version = "*", optional = true}  # groups = ["user", "dev"]
+matplotlib = "3.2"  # groups = ["user", "dev"]
```
SimonBiggs, line 46: Would you be able to provide a more detailed reasoning for this?

SimonBiggs (changes requested):
Looking good :). The main changes needed that are left are moving all of the code changes from `lib/pymedphys/_mosaiq/helpers.py` into something within `lib/pymedphys/_experimental/chartchecks/...` and making sure to leave that `helpers.py` module unchanged for now. Without having Mosaiq CI it's hard to think/work through the implications of changing those files for now.

The other one is that matplotlib pinning. We'll have to adjust how that's been done, and if there is a way around it, I'd rather not need to do it at all. Can you go into more detail as to what happened to need to do that?

pyproject.toml
sjswerdloff: I'll start looking at the matplotlib compatibility issues

> matplotlib 3.3 doesn't play well with sip.setapi but version 3.2 does

https://github.com/ipython/ipython/issues/276 has a bit to say about it, or at least something very strongly related to it.
So there might be some magic in calling sip.setapi with the right version at the right time or doing the import in a particular order

SimonBiggs (approved):
@rembishj this is beautiful :slightly_smiling_face: 

I've made a few final touch ups to take it the last few steps, and I am now ready to merge :slightly_smiling_face: :balloon:

Happy new year! :tada:

lib/pymedphys/_experimental/chartchecks/dvh_helpers.py
```diff
@@ -0,0 +1,44 @@
+from pymedphys._imports import dicompylercore
+from pymedphys._imports import pandas as pd
+from pymedphys._imports import plotly, plt
+from pymedphys._imports import streamlit as st
+
+
+def plot_dvh(rs_file, rd_file):
+    rs_file = "P:/Share/Chris/BAKER_JEANETTE_61163789/RS.1.3.46.670589.13.15476.20191204085114.491830"
+    rd_file = "P:/Share/Chris/BAKER_JEANETTE_61163789/RD.1.3.46.670589.13.15476.20191204085119.163981"
+    rp_file = "P:/Share/Chris/BAKER_JEANETTE_61163789/RP.1.3.46.670589.13.15476.20191204085117.752796"
```
SimonBiggs, line 10: Is this a name that is allowed to be shared?

lib/pymedphys/_experimental/chartchecks/helpers.py
```diff
@@ -0,0 +1,417 @@
+# Copyright (C) 2018 Cancer Care Associates
```
SimonBiggs, line 1: Can you change this copyright to your own, there's not much left from the original file here.

Also, this license header, once you put your copyright on it, can you put this same header at the top of all of your files?

SimonBiggs (approved):
Just a couple of hopefully little things. Make those changes and we hopefully should be ready to merge (as long as all the tests pass).

lib/pymedphys/_experimental/chartchecks/dvh_helpers.py
rembishj: Oops, great catch. This should be removed.

lib/pymedphys/_experimental/chartchecks/dvh_helpers.py
```diff
@@ -0,0 +1,62 @@
...
+        dvh_structures["bincenters"].append(calcdvh.bincenters)
+        dvh_structures["counts"].append(calcdvh.counts)
+        dvh_structures["volume"].append(calcdvh.volume)
+    #     ax.plot(structures[i]['bincenters'], structures[i]['counts']/structures[i]['volume'], label = calcdvh.name)
+    #     ax.legend(bbox_to_anchor=(1.05, 1), loc='upper left', borderaxespad=0.)
+    # ax.set(ylim=(0,1))
+    # st.pyplot(fig)
+    data = pd.DataFrame.from_dict(dvh_structures)
+    data["rel_volume"] = data["counts"] / data["volume"]
+    d1 = data.explode("bincenters").drop(columns=["counts", "rel_volume"])
```
rembishj, line 50: Any ideas how to make my DVH plotting quicker? It was much faster when I used matplotlib, however I like using plotly as it gives me an interactive DVH (I can select which structures I want displayed, hover over to get values, etc). The main problem with plotly is (to my knowledge) I can't use an array within a dataframe cell. I have to use the explode function to give every bin it's own row, and I believe this is the main part slowing it down.

lib/pymedphys/_experimental/chartchecks/dvh_helpers.py
SimonBiggs: Kk, is this the sort of thing that needs to be scrubbed from the GitHub history? (As in, every commit on GitHub has its history stored). It's a bit labour intensive to scrub something from the history... But I can make it happen.

Is that something that is needed here?

lib/pymedphys/_experimental/chartchecks/dvh_helpers.py
SimonBiggs: I'd need to look into this quote a bit deeper to answer this one. If you could produce a mini Streamlit app, one with matplotlib and one with plotly and then post the code on https://discuss.streamlit.io they should be able to give you a nice answer quicker than I can :)

SimonBiggs:
Hi @rembishj,

I have done some git history foo and have made it so that according to GitHub the file `dvh_helpers.py` didn't exist until just now:

https://github.com/pymedphys/pymedphys/blame/weekly_checks/lib/pymedphys/_experimental/chartchecks/dvh_helpers.py

If you could make sure to delete your local version of PyMedPhys and re-clone as described in https://github.com/pymedphys/pymedphys/issues/1297 that'd be awesome :).

SimonBiggs:
@rembishj I'm ready to merge this if you're ready for me to do the merge?

SimonBiggs:
Merging :) 🎉

rembishj:
> Merging :) 🎉

Whoop!! Thanks for helping get this ready to merge! I’ll be back in the US Sunday, so hopefully the time difference won’t be as much of a burden with communication :)

SimonBiggs:
No burden :) all good. I just figured it was at a good place to merge and I
wasn't so keen on having diverge again.

PR #1267, opened 2020-12-13
Upgrade e2e yarn lock

SimonBiggs:
Only failure is in upgrades. This PR is to fix:

image

The above isn't really an issue as we are not using those yarn dependencies to host a webpage, just to run our tests. Nevertheless, doesn't hurt to run an upgrade there.

### DIFF
--- a/lib/pymedphys/tests/e2e/yarn.lock
+++ b/lib/pymedphys/tests/e2e/yarn.lock
@@ -47,9 +47,9 @@
     lodash.once "^4.1.1"
 
 "@samverschueren/stream-to-observable@^0.3.0":
-  version "0.3.0"
-  resolved "https://registry.yarnpkg.com/@samverschueren/stream-to-observable/-/stream-to-observable-0.3.0.tgz#ecdf48d532c58ea477acfcab80348424f8d0662f"
-  integrity sha512-MI4Xx6LHs4Webyvi6EbspgyAb4D2Q2VtnCQ1blOJcoLS6mVa8lNN2rkIy1CVxfTUpoyIbCTkXES1rLXztFD1lg==
+  version "0.3.1"
+  resolved "https://registry.yarnpkg.com/@samverschueren/stream-to-observable/-/stream-to-observable-0.3.1.tgz#a21117b19ee9be70c379ec1877537ef2e1c63301"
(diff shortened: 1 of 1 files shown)


lib/pymedphys/tests/e2e/yarn.lock
```diff
@@ -969,11 +969,16 @@ ms@2.0.0:
   resolved "https://registry.yarnpkg.com/ms/-/ms-2.0.0.tgz#5608aeadfc00be6c2901df5f9861788de0d597c8"
   integrity sha1-VgiurfwAvmwpAd9fmGF4jeDVl8g=
 
-ms@2.1.2, ms@^2.1.1:
+ms@2.1.2:
   version "2.1.2"
   resolved "https://registry.yarnpkg.com/ms/-/ms-2.1.2.tgz#d09d1f357b443f493382a8eb3ccd183872ae6009"
   integrity sha512-sGkPx+VjMtmA6MX27oA4FBFELFCZZ4S4XqeGOXCv68tT+jb3vk/RyaKWP0PTKyWtmLSM0b+adUTEvbs1PEaH2w==
 
+ms@^2.1.1:
```
sjswerdloff, line 977: I'm not familiar with yarn or the yarn lock file, but it looks like some kind of mapping has changed, where a range of versions were going to 2.1.2 before, but now if a specific request for 2.1.2 is made, that's what is provided, but if a request for 2.1.1 or later is requested, 2.1.3 is used.
Unless there is something weird about 2.1.2, I think it would be easier to just have 2.1.3 and "that's what you get".

sjswerdloff (approved):

sjswerdloff:
See https://github.com/executablebooks/jupyter-book/issues/1137 regarding failing update test
I can't find sphinxcontrib-bibtex in pyproject.toml

lib/pymedphys/tests/e2e/yarn.lock
SimonBiggs: This lock file isn't manually produced. Essentially it's just the same as a `poetry.lock` file, except instead of PyPI being the package source, it's NPM.

So in this case, the goal is to accept what yarn determines when an update is run.

PR #1271, opened 2020-12-14
A few changes to the changelog

SimonBiggs:


SimonBiggs:
@termim, wanted to say a massive thank you for all the work you've been doing in `pymssql`.

One example of many, is the following:
https://github.com/pymssql/pymssql/pull/677

Thankyou 🙂

CHANGELOG.md
```diff
@@ -13,39 +13,94 @@ All notable changes to this project will be documented in this file.
...
+    [MyST](https://jupyterbook.org/content/myst.html) Documentation formatting.
+* Increased docstring coverage of public functions
+* Installation on MacOS (Intel) has been simplified and is now the same as for
+  other platforms, thanks to [@termim](https://github.com/termim) who has taken
+  on the mantle of maintaining `pymssql`. See PR
+  [pymssql#677](https://github.com/pymssql/pymssql/pull/677) for more details.
+* Streamlit development now supports reloading during development across the
+  whole PyMedPhys library. See PR
+  [#1202](https://github.com/pymedphys/pymedphys/pull/1202) for more details.
+* A pre-release binary has been built and CI infrastructure around it has begun
```
sjswerdloff, line 63: I think if it isn't in the release, it's best not to mention it.  
Otherwise it's a "future" (which is understandable when one has to market one's product, but I don't think that's the case here).
If it's in a pre-release, that should be in the notes for the pre-release (dev/beta/alpha, whatever), and then show up again in the actual release.

sjswerdloff (approved):
but see comment...

CHANGELOG.md
SimonBiggs: My thoughts here was that since the binary cause a range of flow on changes (eg `python -m pymedphys`) it was worth at least mentioning in a low-profile fashion. It is helpful at least as a record when "looking back" a year from now. Let's say I wanted to find out what tag has the first instance of this in it, I have found that kind of flag as displayed here quite interesting.

sjswerdloff: I'm OK with it.  You have a strong preference in one direction, I have a mild preference in the other.

SimonBiggs: > I'm OK with it. You have a strong preference in one direction, I have a mild preference in the other.

Thanks Stuart 🙂 as our community grows as a group we're only going to have more diverging preferences around the place. It's going to be a learning curve for me on how to appropriately navigate that.

SimonBiggs:
@sjswerdloff As some final steps here, I edited the release to copy in the release notes:

image

And then I created a `0.35.x` branch to track the state of the repo at this release should we ever need to backport a critical bugfix to this version:

```powershell
C:\Users\sbiggs\git\pymedphys [main ≡]> git pull

Already up to date.


C:\Users\sbiggs\git\pymedphys [main ≡]> git checkout -b 0.35.x

Switched to a new branch '0.35.x'


C:\Users\sbiggs\git\pymedphys [0.35.x]> git push

fatal: The current branch 0.35.x has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin 0.35.x



C:\Users\sbiggs\git\pymedphys [0.35.x]> git push --set-upstream origin 0.35.x

Total 0 (delta 0), reused 0 (delta 0)
remote:
remote: Create a pull request for '0.35.x' on GitHub by visiting:
remote:      https://github.com/pymedphys/pymedphys/pull/new/0.35.x
remote:
To github.com:pymedphys/pymedphys.git
 * [new branch]        0.35.x -> 0.35.x
Branch '0.35.x' set up to track remote branch '0.35.x' from 'origin'.
```

I don't intend to be backporting bugfixes at the `0.x.y` stage of PyMedPhys, but, I am doing it more as practice for "the real thing" when we go to `1.x.y`.

sjswerdloff:
whoops, I should have created the 0.35.x branch as soon as I did the release.
And yes, we absolutely need to have those support branches from 1.* 
Following the GitFlow approach, to the best of my knowledge, "that is the way".
And it's a good thing to have in 0.* given the long gestation of PyMedPhys.

SimonBiggs:
> Following the GitFlow approach, to the best of my knowledge, "that is the way".

This is the way 🙂

SimonBiggs:
Ahh, @sjswerdloff, just found something a bit more critical than other things regarding the release. There wasn't any wheel file uploaded:

image

This would likely be because you didn't first run `poetry build` before running `poetry publish`. Would you be able to checkout the `v0.35.0` tag and run a build and then a publish so that PyPI has a wheel as well?

sjswerdloff:
There wouldn't have been a .tar.gz if I hadn't done a poetry build (I did one...).
But there may have not been a wheel built when I did that (I don't have a good enough memory, but I don't recall seeing the wheel built last time).
However, I went ahead and checkout out the 0.35.x branch and did
poetry build
poetry publish

the wheel was built, the wheel made it up, but the .tar.gz was rejected because it was already there.
Given there isn't a functional difference in the source code, I'm OK with leaving it as it is, instead of doing a 0.35.1 bug fix release to ensure/force that the updated changelog is in the source.

I'll try to do 0.36 when it's time for that, and hopefully that will go smoother and faster.

SimonBiggs:
> There wouldn't have been a .tar.gz if I hadn't done a poetry build

A `tar.gz` file is made (and re-made) everytime `pymedphys dev propagate` is run. See:

https://github.com/pymedphys/pymedphys/blob/740b6ab83101afbda8e406fc80a2f8708b227426/lib/pymedphys/_dev/propagate.py#L82-L86

and:

https://github.com/pymedphys/pymedphys/blob/740b6ab83101afbda8e406fc80a2f8708b227426/lib/pymedphys/_dev/propagate.py#L197-L207

sjswerdloff:
ah... so I confused the propagate and the build steps because they have overlapping functionality.

SimonBiggs:
> ah... so I confused the propagate and the build steps because they have overlapping functionality.

Well, propagate isn't aiming to "build". It's just the most consistent way I could think of to create a `setup.py`... One option would be to delete the created `tar.gz` file after propagating to minimise the confusion?

What are your thoughts?

sjswerdloff:
> > ah... so I confused the propagate and the build steps because they have overlapping functionality.
> 
> Well, propagate isn't aiming to "build". It's just the most consistent way I could think of to create a `setup.py`... One option would be to delete the created `tar.gz` file after propagating to minimise the confusion?
> 
> What are your thoughts?

Yes.  It should be deleted.  That way if a poetry publish takes place without a poetry build, it fails completely.
Fail fast is better than not failing at all when it should.  Actually, failing fast is best, period.

SimonBiggs:
> Actually, failing fast is best, period.

🙂 Agreed

PR #1293, opened 2020-12-26
Make BB finding more resilient

SimonBiggs:
I think this has hit the money :).


lib/pymedphys/_experimental/wlutz/findbb.py
```diff
@@ -43,41 +59,76 @@ def find_bb_centre(
...
+    assert len(within_tolerance) == len(BB_SIZE_FACTORS_TO_SEARCH_OVER)
+
+    if np.all(within_tolerance):
+        return median_of_predictions
+
+    if repeats == 0:
+        raise ValueError("Unable to determine BB position within designated repeats")
+
+    out_of_tolerance = np.invert(within_tolerance)
+    if np.sum(out_of_tolerance) > len(BB_SIZE_FACTORS_TO_SEARCH_OVER) / 3:
```
sjswerdloff, line 89: Is this divide by three to address x, y, z or something else?

sjswerdloff (approved):

lib/pymedphys/_experimental/wlutz/findbb.py
SimonBiggs: It's a little bit of a "magic number", essentially the ball bearing is found independently 11 times each time with a slightly different input parameter. If more than a third of them deviate from the median by more than 0.2 mm then raise an error.

Here are the 11 different parameters used:

https://github.com/pymedphys/pymedphys/blob/56dcdf8b29f7d7773551aa343f70fb3e5c45275a/lib/pymedphys/_experimental/wlutz/findbb.py#L26-L38

SimonBiggs: In order to finally exit the recursion though every variation all needs to agree with the median within the tolerance of 0.2 mm:

https://github.com/pymedphys/pymedphys/blob/56dcdf8b29f7d7773551aa343f70fb3e5c45275a/lib/pymedphys/_experimental/wlutz/findbb.py#L82-L83

PR #1294, opened 2020-12-29
Add another failing case, don't allow a large travel distance

SimonBiggs:
Restrict total travel distance of BB searching. Loosen strictness on returning the median if all but one of the BB positions agree.


prototyping/wlutz/should-fail-fast.ipynb
```diff
@@ -144,7 +144,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.6"
+   "version": "3.7.9"
```
sjswerdloff, line 147: Intentional downgrade?

sjswerdloff (approved):

prototyping/wlutz/should-fail-fast.ipynb
SimonBiggs: Not really an issue in these notebooks. It's just a side effect of defaulting to 3.7 and work and 3.8 at home. Notebooks always store which Python version they were last run with... if I had a choice I'd delete that tag altogether to stop the change conflicts.

PR #1296, opened 2020-12-31
[STACK base] Attempt collimator correction

SimonBiggs:


lib/pymedphys/_experimental/streamlit/apps/collimator_corrections.py
```diff
@@ -39,3 +46,277 @@ def main():
...
+
+    # st.write(dataframe)
+
+    treatments = dataframe_by_algorithm["treatment"].unique()
+    # st.write(treatments)
+
+    treatment = treatments[0]
+    # st.write(treatment)
+
+    dataframe_by_treatment = _filter_by(dataframe_by_algorithm, "treatment", treatment)
```
sjswerdloff, line 64: this is the dataframe filtered first by algorithm and then by treatment, but the name of the variable could be construed as dataframe by treatment only (no additional pre-filtering by algorithm).  letters are cheap.  time spent re-reading code is expensive.

lib/pymedphys/_experimental/streamlit/apps/collimator_corrections.py
```diff
@@ -39,3 +46,277 @@ def main():
...
+    dataframe_by_treatment.reset_index(inplace=True)
+    st.write(dataframe_by_treatment)
```
sjswerdloff, line 66: you could put most of the above in a function that returns "dataframe_by_treatment"

lib/pymedphys/_experimental/streamlit/apps/collimator_corrections.py
```diff
@@ -39,3 +46,277 @@ def main():
...
+    coll_corrected = _transform_points_to_field_reference_frame(
+        dataframe_by_treatment, ["diff_x_coll_corrected", "diff_y_coll_corrected"]
+    )
+
+    fig, ax = plt.subplots()
+    ax.plot(gantry, original[:, 0], "o-", alpha=0.3)
+    ax.plot(gantry, logfile_corrected[:, 0], "o-", alpha=0.3)
+    st.pyplot(fig)
+
+    fig, ax = plt.subplots()
```
sjswerdloff, line 231: looks ripe for extract as method and pass in parameters that vary

lib/pymedphys/_experimental/streamlit/utilities/icom.py
```diff
@@ -266,12 +266,8 @@ def _determine_width_length_centre(mlc, jaw):
     width = pd.Series(np.sum(mean_mlc, axis=1), name="width")
     length = pd.Series(jaw[:, 0] + jaw[:, 1], name="length")
 
-    centre_x = pd.Series(
-        (-mean_mlc[:, 0] + mean_mlc[:, 1]) / 2, name="experimental_icom_centre_x"
-    )
-    centre_y = pd.Series(
-        (-jaw[:, 0] + jaw[:, 1]) / 2, name="experimental_icom_centre_y"
-    )
+    centre_x = pd.Series((mean_mlc[:, 0] - mean_mlc[:, 1]) / 2, name="centre_x")
```
sjswerdloff, line 269: deliberate negative of previous commit?

lib/pymedphys/_experimental/wlutz/transformation.py
```diff
@@ -12,18 +12,47 @@
...
+    xx_transformed = transformed[0]
+    yy_transformed = transformed[1]
+
+    xx_transformed.shape = xx.shape
+    yy_transformed.shape = yy.shape
+
+    return xx_transformed, yy_transformed
+
+
+def translate_and_rotate_transform(centre, rotation):
```
sjswerdloff, line 45: the order in which one translates and rotates matters.
the function name indicates translate first, then rotate.
The code applies rotate first, then translate.
That could cause some confusion should someone else use the function.

sjswerdloff (approved):
collimator_corrections could use some refactoring to reduce the size of the function

lib/pymedphys/_experimental/streamlit/utilities/icom.py
SimonBiggs: Yup, previously I had just added in the result with a sort of "note to self experimental flag, I haven't tested if the coords are right here". Once I took some data, I had the x and y sign the wrong way round.

SimonBiggs (commented):
> collimator_corrections could use some refactoring to reduce the size of the function

Yup, certainly, normally I'd agree, but that is a paint swipe of prototyping code. The entire file itself is likely to be deleted and merged into the WLutz GUI, at that stage, the subset of pieces that were actually useful will be refactored into logical segments. This is certainly "prototyping" in the sense, "code for my own education", as opposed to code that will make it into the final tool.

...This was me trying to get to the bottom of a weird effect where deviations with collimator rotations don't seem to be primarily due to a constant collimator rotation effect. Instead, there does seem to be a gantry angle component. I figured this may be a gravity effect on the leaves, so I pulled the iCom data out to correct for changes with gantry angle (and sure enough there did appear to be gravity effects recorded in the iCom logs). But still, not enough to account for what I am seeing.

Essentially the beam position is changing with something that doesn't seem to be able to be accounted for by the reported iCom MLC/Jaw positions, or the focal spot deviations (which would have the same effect with coll rotation) or gantry sag. ...so... I'm on the hunt for what else could cause these systematic beam position differences.

...I'm back at work on Tuesday, and I plan to take a swathe of more measurements/data to try and isolate what is causing the beam position changes. The end goal is to be able to report the total beam position deviation, but then also report that beam position deviation as a sum of its components... I can't do that though if the a significant deviation contribution is unknown... Anyway... more digging to do.

PR #1298, opened 2021-01-06
Pylint skip coll

SimonBiggs:
That collimator correction code is so bad it intermittently sends pylint into recursive loops :sweat_smile:

https://github.com/pymedphys/pymedphys/runs/1654556093#step:35:1505

...so... this skips pylint on that file for now...

### DIFF
--- a/lib/pymedphys/_experimental/streamlit/apps/collimator_corrections.py
+++ b/lib/pymedphys/_experimental/streamlit/apps/collimator_corrections.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# pylint: skip-file
+
 from pymedphys._imports import numpy as np
 from pymedphys._imports import pandas as pd
 from pymedphys._imports import plt


sjswerdloff (approved):
This is a band aid, right?
Things will get cleaned up, split out, etc so that whatever gets skipped is down to a narrow set of code with well understood reasons for skipping it?

SimonBiggs:
> This is a band aid, right?
> Things will get cleaned up, split out, etc so that whatever gets skipped is down to a narrow set of code with well understood reasons for skipping it?

Yes most certainly. Or that entire file will just get deleted. For now, it is a bit of a dead end. It appears that with gravity the MLCs can move between 0.2 and 0.4 mm and in order to break those deviations out into their own components I need a more accurate time alignment between iCom and iView. So for now, I'm going to shelve that idea (temporarily?) and instead make our WLutz field run at collimators -90 and 90 so that the Jaw is under gravity, not the MLCs (the Jaw showed no such deviation magnitudes).

PR #1304, opened 2021-01-11
[STACK config -> base] Allow config to be changed within Apps

SimonBiggs:
Previously the design was "config" was effectively a global variable that was pulled in by calling a function "get_config". This has been refactored out, now every function that uses the config must have the config passed through to the function, and that way, an app can adjust the config to be something custom before booting up the App.

That way, the online web app can display the "file upload only" config by default for the MetersetMap GUI, and as another benefit, the demo config can be selected via a radio button meaning, to run the Cypress tests it doesn't matter what config you have locally, the demo config will always be used, regardless of the config you have locally on your machine.

---

Even though this PR touches quite a few functions, and those functions don't have docstrings etc, for now, if possible I'd like to enact the following concept:

> In general, reviewers should favor approving a [PR] once it is in a state where it definitely improves the overall code health of the system being worked on, even if the [PR] isn't perfect.

https://github.com/google/eng-practices/blob/master/review/reviewer/standard.md

Meaning, if this PR certainly does improve the overall code health, then it meets the requirements for a PR approval.


sjswerdloff (approved):
Typing of the config object that gets passed in as an argument would be nice

SimonBiggs:
> Typing of the config object that gets passed in as an argument would be nice

Good call. I'll spin this off into a new issue. I think as a part of bumping config to version 2, create a type etc would be very helpful.

PR #1309, opened 2021-01-12
[STACK wlutz-> config -> base] Wlutz experimental work

SimonBiggs:


lib/pymedphys/_experimental/streamlit/apps/wlutz/_utilities.py
```diff
@@ -22,6 +23,17 @@
...
 
 
+def filter_by(dataframe, column, value):
+    filtered = dataframe.loc[dataframe[column] == value]
+
+    return filtered
+
+
+# https://stackoverflow.com/a/16090640/3912576
+def natural_sort_key(s, _nsre=re.compile("([0-9]+)")):
```
sjswerdloff, line 33: Consider import natsort instead of coding locally

sjswerdloff (approved):
Strip body of code that is completely commented when moving out of experimental.
More parameter typing would be nice

lib/pymedphys/_experimental/streamlit/apps/wlutz/_utilities.py
SimonBiggs: Yup, good call. I'll do that before merge.

SimonBiggs:
> Strip body of code that is completely commented when moving out of experimental.

Yup, certainly 🙂 

> More parameter typing would be nice

I'll keep that in mind and see if I can make sure that happens as I am refactoring.

PR #1310, opened 2021-01-12
[STACK pyright -> config -> base] Add in Pyright

SimonBiggs:
Have this as a 'back-up' in case pylint doesn't do its job...


sjswerdloff (approved):
Hope this is not necessary.
Bad smell to dump pylint

SimonBiggs:
> Bad smell to dump pylint

Certainly not dumping pylint. Just including pyright with the goal being if pylint doesn't pick it up, hopefully pyright will?

PR #1315, opened 2021-01-13
Build a hyperlinked dependency tree for the docs

SimonBiggs:
There are some auto-generated svgs in this PR. Only 4 files are actually changed here. The bulk of the code is within 2 files and they are within experimental. To view the PR it might be best to use the following filter:

https://github.com/pymedphys/pymedphys/pull/1315/files?file-filters%5B%5D=.dot&file-filters%5B%5D=.hash&file-filters%5B%5D=.lock&file-filters%5B%5D=.ps&file-filters%5B%5D=.py&file-filters%5B%5D=.toml&file-filters%5B%5D=.txt

There is an o-so-slight change to the API. Essentially a couple of items that appeared to be part of the API but shouldn't have been were adjusted.

---

This PR stemmed from https://pymedphys.discourse.group/t/user-guide-or-manuals/75/4?u=simonbiggs


SimonBiggs:
@sjswerdloff you might be interested in the following:

https://github.com/pymedphys/pymedphys/tree/bring-back-dep-tree/lib/pymedphys/docs/trees

sjswerdloff (approved):

SimonBiggs:
Also, I'm not actually following the "functions used" here. Instead just module imports.

Modules themselves could be represented as graphviz "clusters" and then the nodes could be the functions within those modules that are depended upon. The URLs for those "function nodes" could then pull the line numbers from the ast objects, and then use those line numbers to make the GitHub links go straight to the function within the node. If linking to line numbers it is probably worth also pulling the git commit hash so that the URLs point to a commit hash, and therefore the linenumbers always align...

We could then use ast to find where the particular imported function is used within the file and only traverse the dependencies if a function from that module is actually utilised (as opposed to just imported).

@maveLC what are your thoughts?

PR #1326, opened 2021-01-18
[STACK unet] Create UNet based on minified animal CTs

SimonBiggs:
Heya @matthewdeancooper,

I'd be keen to get your opinion on the following:

https://colab.research.google.com/github/pymedphys/pymedphys/blob/54926066771e6a30bbf03b3a741aa22d3044f316/prototyping/unet.ipynb

My goal was to make it as simple as possible but for it to still work well.

---

I still need to add explanations etc, it is something that I could have run for 3 epochs and include within the docs...

---

~~TODO:~~

* ~~Add a model weights file that can be used to show what it looks like after quite a bit of training~~
* ~~Add to the docs, and have the docs make it undergo 3 epochs~~
* ~~Load up the weights, then include plots of a range of cases at the end~~
* ~~Add in more descriptions around the various steps~~

[EDIT] -- CPU training takes about 1 minute per epoch. That's a bit too much for within the docs. Consider that at a later point.


sjswerdloff:
8000 lines of code to review in one PR.
What would Simon Say?

SimonBiggs:
> 8000 lines of code to review in one PR.
> What would Simon Say?

It's all notebooks, all in prototyping, Simon would say good to go :)

SimonBiggs:
(Simon might not even read it... He might just verify that nothing has
crept somewhere it shouldn't have and then declare that it wasn't read, but
because of its location it's okay to merge)

lib/pymedphys/_experimental/autosegmentation/unet.py
```diff
@@ -0,0 +1,274 @@
...
+    grid_size,
+    output_channels,
+    max_filter_num=None,
+    number_of_filters_start=32,
+    min_grid_size=8,
+    num_of_fc=2,
+    batch_normalisation=True,
+    use_dropout=True,
+):
+    inputs = tf.keras.layers.Input((grid_size, grid_size, 1))
```
sjswerdloff, line 15: docstring

lib/pymedphys/_experimental/autosegmentation/unet.py
```diff
@@ -0,0 +1,274 @@
...
+def encode(
+    x,
+    number_of_filters,
+    number_of_convolutions=2,
+    kernel_size=3,
+    dropout_rate=1.0,
+    pool=True,
+    batch_normalisation=True,
+):
+    x = _convolutions(
```
sjswerdloff, line 189: docstring

lib/pymedphys/_experimental/autosegmentation/unet.py
```diff
@@ -0,0 +1,274 @@
...
+def decode(
+    x,
+    skip,
+    number_of_filters,
+    number_of_convolutions=2,
+    kernel_size=3,
+    dropout_rate=1.0,
+    batch_normalisation=False,
+):
+    x = _conv_transpose(x, number_of_filters, kernel_size)
```
sjswerdloff, line 212: docstring

sjswerdloff (changes requested):
docstring for public functions.
I'd settle for the automatically generated docstring and twenty five words or less describing what the public function does.

I didn't bother going through most of the ipynb, they were listed as too large to display the diff automatically.
The problem I see with that much notebook stuff is that it's very awkward to code review, compared to "straight up python".  That suggests that there should be a tool that makes it easier to review (e.g. side by side presentation of a rendered notebook, with separate diff of the actual code within each cell).

lib/pymedphys/_experimental/autosegmentation/unet.py
SimonBiggs: Here I might instead opt for just deleting the code altogether instead of giving it any more love. In the end I didn't use it.

...generally I prefer to leave experimental code hanging around for a bit to refer back to before deleting it though...

SimonBiggs:
@sjswerdloff in the end I opted to delete `unet.py`, in order to decouple the final notebook from PyMedPhys I inlined the parts of that file that were useful into the notebook.

> The problem I see with that much notebook stuff is that it's very awkward to code review, compared to "straight up python". That suggests that there should be a tool that makes it easier to review (e.g. side by side presentation of a rendered notebook, with separate diff of the actual code within each cell).

I agree, and there are tools out there. For a while we were using https://www.reviewnb.com/ but, I found it too noisy within the PRs. We could consider giving it another go...

SimonBiggs:
@sjswerdloff here's an example of a commit diff displayed on nbreview:

https://app.reviewnb.com/pymedphys/pymedphys/commit/54926066771e6a30bbf03b3a741aa22d3044f316/

SimonBiggs:
Hi @sjswerdloff,

Given all code here is within prototyping and nothing is within experimental (`unet.py` was deleted) I'm going to merge. Hopefully that's okay given the "changes requested" was placed on this PR and has not yet been removed.

PR #1328, opened 2021-01-20
adding database optional parameter to mosaiq.connect; updated noteboo…

dg1an3:
…ks to use test db

### DIFF
--- a/lib/pymedphys/_mosaiq/connect.py
+++ b/lib/pymedphys/_mosaiq/connect.py
@@ -24,6 +24,10 @@ from pymedphys._imports import keyring, pymssql
 KEYRING_SCOPE = "PyMedPhys_SQLLogin_Mosaiq"
 
 
+def get_storage_name(hostname, port=1433, database="MOSAIQ"):
+    return f"{KEYRING_SCOPE}_{hostname}:{port}/{database}"
+
+
 class WrongUsernameOrPassword(ValueError):
     pass
 
@@ -195,7 +199,7 @@ def _connect_with_credential_then_prompt_if_fail(
 
     """
     server, port = _separate_server_port_string(sql_server_and_port)
-    storage_name = f"{KEYRING_SCOPE}_{sql_server_and_port}/{database}"
+    storage_name = get_storage_name(server, port=port, database=database)
 
     user, password = _get_username_password(
         storage_name,
--- a/lib/pymedphys/_mosaiq/connect.py
+++ b/lib/pymedphys/_mosaiq/connect.py
@@ -1,3 +1,4 @@
+# Copyright (C) 2021 Derek Lane, Cancer Care Associates
 # Copyright (C) 2018 Cancer Care Associates
 
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/lib/pymedphys/_mosaiq/connect.py
+++ b/lib/pymedphys/_mosaiq/connect.py
@@ -84,34 +84,31 @@ def execute_sql(cursor, sql_string, parameters=None):
     return data
 
 
-def get_username_and_password_without_prompt(storage_name, database="MOSAIQ"):
-    user = keyring.get_password(f"{database}SQL_username", storage_name)
-    password = keyring.get_password(f"{database}SQL_password", storage_name)
+def get_username_and_password_without_prompt(storage_name):
+    user = keyring.get_password("MosaiqSQL_username", storage_name)
+    password = keyring.get_password("MosaiqSQL_password", storage_name)
 
     return user, password
 
 
-def save_username(sql_server_and_port, username, database="MOSAIQ"):
-    keyring.set_password(f"{database}SQL_password", sql_server_and_port, username)
+def save_username(storage_name, username):
+    keyring.set_password("MosaiqSQL_username", storage_name, username)
 
 
-def save_password(sql_server_and_port, password, database="MOSAIQ"):
-    keyring.set_password(f"{database}SQL_password", sql_server_and_port, password)
+def save_password(storage_name, password):
+    keyring.set_password("MosaiqSQL_password", storage_name, password)
 
 
 def _get_username_password(
-    storage_name,
-    user_input=input,
-    password_input=getpass,
-    output=print,
-    database="MOSAIQ",
+    storage_name, user_input=input, password_input=getpass, output=print
 ):
-    user, password = get_username_and_password_without_prompt(storage_name, database)
+    user, password = get_username_and_password_without_prompt(storage_name)
 
     if user is None or user == "":
         output(
-            "Provide a user that only has `db_datareader` access to the "
-            "'{}' database at '{}'".format(database, storage_name)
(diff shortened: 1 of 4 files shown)


lib/pymedphys/_mosaiq/connect.py
```diff
@@ -151,7 +151,11 @@ def delete_credentials(sql_server_and_port):
 
 
 def _connect_with_credential_then_prompt_if_fail(
-    sql_server_and_port, user_input=input, password_input=getpass, output=print
+    sql_server_and_port,
+    user_input=input,
+    password_input=getpass,
+    output=print,
+    database="MOSAIQ",
```
SimonBiggs, line 158: The keyring password "key" will need to be adjusted to have the database name added to it so that login credentials for one database on the same hostname:port combo are not attempted to be used for a different database name that happens to be served on that same combo.

SimonBiggs (changes requested):
Awesome 🙂 👍 

One change requested, all code that uses `keyring.set_password`, `keyring.get_password` or `keyring.delete_password` needs to be changed to include the database name within it so that there isn't a clash.

lib/pymedphys/_mosaiq/connect.py
dg1an3: you mean instead of 

`keyring.get_password("MosaiqSQL_username", storage_name)`

change to

`keyring.get_password(f"{database_name}SQL_username", storage_name)`

SimonBiggs: Either there, or within the storage name. Figured I'd leave the design decision on that front to your discretion 🙂.

SimonBiggs (approved):
Yup, beautiful, that fixes the issue 🙂

SimonBiggs:
@dg1an3 I've just given you merge access to the main branch. When you're happy with this PR can you give it a merge?

image

lib/pymedphys/_mosaiq/connect.py
dg1an3: sorry I altered the approach to put it in the storage_name, though I'm not sure I understand how the storage name shows up in credential manager

SimonBiggs: > sorry I altered the approach to put it in the storage_name, though I'm not sure I understand how the storage name shows up in credential manager

It's all good. I am happy to go with this approach. If we find out down the line we'd like to adjust it further we are certainly free to do that.

SimonBiggs:
Probably the main improvement I can see now is to refactor out `PyMedPhys_MosaiqSQL`. Placing a constant at the top of the file called `KEYRING_SCOPE = "PyMedPhys_MosaiqSQL"` should work well. Once you're happy with where this is at, let me know and I'll run through some manual tests after the new changes.

lib/pymedphys/_mosaiq/connect.py
dg1an3: I made a change to the order of parameters for keyring.get_password/set_password  -- please have a look and tell me if it makes sense to you.  I think the entries in Windows Credential manager make more sense if "storage_name" (first parameter) is the one with the server/database, and "user_name" (second parameter) is _PyMedPhys_MosaiqSQL_xxx_ -- what do you think?

lib/pymedphys/_mosaiq/connect.py
```diff
@@ -184,24 +195,28 @@ def _connect_with_credential_then_prompt_if_fail(
 
     """
     server, port = _separate_server_port_string(sql_server_and_port)
+    storage_name = f"{KEYRING_SCOPE}_{sql_server_and_port}/{database}"
```
dg1an3, line 198: form the keyring storage name from: KEYRING_SCOPE, server + port, and database name

lib/pymedphys/_mosaiq/connect.py
SimonBiggs: Yup, that certainly makes sense to me. I'll checkout your PR now and make sure things still work happily :)

SimonBiggs (changes requested):
See https://github.com/pymedphys/pymedphys/pull/1328#issuecomment-764025379

lib/pymedphys/_streamlit/utilities/mosaiq.py
```diff
@@ -14,7 +14,9 @@
 
 from pymedphys._imports import streamlit as st
 
+from pymedphys._mosaiq import _separate_server_port_string
```
dg1an3, line 17: I don't know if importing _separate_server_port_string is a great idea, but it seemed the least intrusive approach

SimonBiggs: Yeah, I'd recommend renaming that function from `_separate_server_port_string` to `separate_server_port_string`, add a short docstring, and then utilise it within `_streamlit/utilities/mosaiq` as you have. One thing though, instead of re-importing, just utilise the already imported `msq_connect` scope:

```python
  hostname, port = msq_connect.separate_server_port_string(server)  # probably best not reusing server variable here
  storage_name = msq_connect.get_storage_name(hostname, port)
  username, password = msq_connect.get_username_and_password_without_prompt(
      storage_name
  )
```

Also, after renaming something like that, it can be really helpful to run:

```
poetry run pymedphys dev lint
```

It'll tell you all the places in the code base where that name change was problematic.

SimonBiggs: Or, another option (up to you), might be to create a function within `_mosaiq.connect` that looks like the following:

```python
def get_username_and_password_from_hostname_port_string(hostname_port_string)
	  hostname, port = separate_server_port_string(hostname_port_string)
	  storage_name = get_storage_name(hostname, port)
	  username, password = get_username_and_password_without_prompt(
	      storage_name
	  )
```

SimonBiggs (changes requested):
Yup, great, that fixed that issue. Upon testing there was one more fix needed though (sorry about all this -- it's really highlighting the need for a Mosaiq CI ...).

The following two lines:

https://github.com/pymedphys/pymedphys/blob/c8bb4db0f1562d283bbf937fbeea371816764e96/lib/pymedphys/_streamlit/utilities/mosaiq.py#L61

https://github.com/pymedphys/pymedphys/blob/c8bb4db0f1562d283bbf937fbeea371816764e96/lib/pymedphys/_streamlit/utilities/mosaiq.py#L74

Need to be changed to use `storage_name` instead of `server`.

SimonBiggs (approved):
Awesome :) thanks for all the back and forth @dg1an3 . Feel free to merge when you're happy with this one.

SimonBiggs (commented):
Something else to squeeze into this PR, can you change the copyright header of the connect.py file to:

```
# Copyright (C) 2021 Derek Lane, Cancer Care Associates
# Copyright (C) 2018 Cancer Care Associates
```

SimonBiggs:
Beautiful. Thanks @dg1an3!

PR #1329, opened 2021-01-20
Chart checks work

rembishj:
going to start doing these pull requests more often now so you can keep up with everythin @SimonBiggs 

Pretty sure I've done this correctly.... please let me know :)


poetry.lock
```diff
@@ -1407,8 +1407,13 @@ name = "parso"
 version = "0.8.1"
 description = "A Python Parser"
 category = "main"
+<<<<<<< HEAD
```
rembishj, line 1410: you might want to double check what happened here, still figuring things out a bit with poetry @SimonBiggs

poetry.lock
```diff
@@ -1407,8 +1407,13 @@ name = "parso"
...
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+=======
+optional = true
+python-versions = ">=3.6"
+>>>>>>> c8bb4db0f1562d283bbf937fbeea371816764e96
```
rembishj: you might want to double check what happened here, still figuring things out a bit with poetry @SimonBiggs

SimonBiggs:
I suspect during your merge you may have broken the `poetry.lock` file. As a fix for that you can delete the lock file entirely then run:

```
poetry update
```

This has the downside of forcing all packages to upgrade, but I'm okay for that to happen as currently the "upgrade tests" are passing.

SimonBiggs:
Actually, scrap that advice. Better advice, that should work every time something like this happens regardless is to go to the version within the `main` branch on GitHub:

https://github.com/pymedphys/pymedphys/blob/main/poetry.lock

And replace your local version with that version found on GitHub, then run `poetry run pymedphys dev propagate`

SimonBiggs:
Actually... even better again... just run:

```
git pull
git checkout chart_checks
git checkout origin/main poetry.lock
git add -A
git commit -m "replacing poetry lock with upstream"
git push
```

That will replace your current file with whatever is seen on the main branch.

rembishj:
> Actually... even better again... run the following:
> 
> ```
> git pull
> git checkout chart_checks
> git checkout origin/main poetry.lock
> poetry run pymedphys dev propagate
> git add -A
> git commit -m "replacing poetry lock with upstream"
> git push
> ```
> 
> That will replace your current file with whatever is seen on the main branch.

I suspected I may have done that... I've given this a shot but encountered an issue. Not sure how to implement the run with --user suggestion or if I should do something else.

image

SimonBiggs:
Actually you don't want to follow any of its suggestions. Environment error is part of the joys of development on Windows. You need to make sure your Poetry environment isn't in use when running that command. So, you need to shut down streamlit, jupyter, and an text editor that might be live running code for introspection...

SimonBiggs:
Yup, next you should just need to run the following locally:

```
poetry run pymedphys dev lint
```

And then fix up any complaints it has :)

rembishj:
by the way, what should I be calling these pull requests as I start submitting them more often?

SimonBiggs:
> by the way, what should I be calling these pull requests as I start submitting them more often?

Within many a conversation about the code changes will likely ensue. I would recommend naming the PR based on something you changed that day, that way, it makes it easier to search later and look back at a conversation if there is anything there that you would like to look back on for reference in the future...

lib/pymedphys/_experimental/chartchecks/compare.py
```diff
@@ -185,22 +188,27 @@ def compare_to_mosaiq(dicom_table, mos_table):
...
+                mosaiq_list = pd.concat([mosaiq_list, add_mosaiq], axis=1)
+
             # continue if the value is not in Mosaiq
             else:
                 continue
 
         values_table = values_table.append(dicom_list, ignore_index=True)
         values_table = values_table.append(mosaiq_list, ignore_index=True)
 
+        dicom_list = pd.DataFrame()
```
SimonBiggs, line 201: The word "list" has a very special meaning in Python, given this isn't a Python list I would change this variable's name.

lib/pymedphys/_experimental/chartchecks/compare.py
```diff
@@ -229,3 +237,20 @@ def weekly_check_color_results(val):
         return ["background-color: #C1FFC1"] * len(val)
     else:
         return ["background-color: #EE6363"] * len(val)
+
+
+def specific_patient_weekly_check_color_results(val):
```
SimonBiggs, line 242: Not a must... but the rest of the code base uses british english... might be some value to stick to the standard of the code-base so that people who want to call PyMedPhys related things can stick to one standard. -- referring to "color" vs "colour".

lib/pymedphys/_experimental/chartchecks/helpers.py
```diff
@@ -168,6 +169,16 @@ def get_all_dicom_treatment_info(dicomFile):
             except (TypeError, ValueError, AttributeError):
                 dicomBeam["tolerance"] = 0
 
+            if dicomBeam["manufacturer"] == "Varian":
+
+                dicomBeam["gantry_angle"] = (180 - dicomBeam["gantry_angle"]) % 360
+                dicomBeam["collimator_angle"] = (
+                    180 - dicomBeam["collimator_angle"]
+                ) % 360
+                dicomBeam["couch_ang"] = (180 - dicomBeam["couch_ang"]) % 360
```
SimonBiggs, line 178: These lines are all duplicating the same work. When this happens, pull the logic out into a separate function and then call that. For example:

```python
def _invert_angle(angle):
    return (180 - angle) % 360


for key in ["gantry_angle", "collimator_angle", "couch_ang"]:
    dicomBeam[key] = _invert_angle(dicomBeam[key])
```

Also, if you change "couch_ang" -> "couch_angle" to be consistent with your other naming conventions you can change the above to:

```python

angle_keys = [key for key in dicomBeam.keys() if "angle" in key]

for key in angle_keys:
    dicomBeam[key] = _invert_angle(dicomBeam[key])
```

That second version has the benefit of also working should you add another "angle" key that also needs transformation.

Also :), change the variable name `dicomBeam` to `dicom_beam` (snake case).

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -51,17 +67,21 @@ def main():
...
+        # Create a dataframe of plan information from DICOM RP file
         dicom_table = get_all_dicom_treatment_info(files["rp"])
         dicom_table["tolerance"] = [
             TOLERANCE_TYPES[item] for item in dicom_table["tolerance"]
         ]
         dicom_table = dicom_table.sort_values(["field_label"])
 
+        # Using MRN from RP file, find patient in MOSAIQ and perform query
         mrn = dicom_table.iloc[0]["mrn"]
         with connect.connect(server) as cursor:
```
SimonBiggs, line 82: Be aware, that connecting to mosaiq like this within a streamlit app will cause it to keep connecting and disconnecting to the server. Streamlit has ways of caching a database connection so that it can stay open. Also, if the user needs to provide a username and password to the database it'll appear in the terminal.

I have made a streamlit mosaiq login function over at:

https://github.com/pymedphys/pymedphys/blob/553732eb8c2f722f7123a66c4e57755db1510881/lib/pymedphys/_streamlit/utilities/mosaiq.py#L20-L38

I'd recommend using the following cached wrapper function:

https://github.com/pymedphys/pymedphys/blob/553732eb8c2f722f7123a66c4e57755db1510881/lib/pymedphys/_streamlit/utilities/mosaiq.py#L88-L90

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -97,23 +121,25 @@ def main():
...
         dicom_DOB = dicom_table.iloc[0]["dob"]
         if DOB == dicom_DOB[0:4] + "-" + dicom_DOB[4:6] + "-" + dicom_DOB[6:8]:
             st.success("DOB: " + DOB)
         else:
             st.error("DOB: " + DOB)
 
-        ########################################################################################################################
-        # check the approval status of various sections in Mosaiq
+        ####################################################################################################################
+        # Section for checking approval statuses
```
SimonBiggs, line 139: Opt for putting sections of code into either their own functions, or even their own files over comments like these.

site-specific/mays/transfer_check_mays.py
```diff
@@ -29,7 +32,7 @@
 
 from pymedphys._experimental.chartchecks.dvh_helpers import plot_dvh
 
-
+streamlit_analytics.start_tracking()
```
SimonBiggs, line 35: Keen to hear how it goes :)

SimonBiggs (approved):
Thanks Jake :) This is great.

I have added a few comments throughout, but I'm going to merge this without those changes. That'll let you get to them when you see fit :).

Also, any comment I make, please feel free to say "I'd rather not for such-n-such reason" or, "would this approach be better". Discussion really helps get the best outcome and help everyone understand. You understand the decisions behind why you do things while you're coding, sometimes my recommendations may just be not a good idea. Make sure to say if you think so, it's okay :)

SimonBiggs:
Also, does the visual changes in streamlit 0.75.0 for wide mode make it look better for your needs?

https://github.com/streamlit/streamlit/releases/tag/0.75.0

site-specific/mays/transfer_check_mays.py
rembishj: I'm having some issues when trying to implement the analytics into the app run in the pymedphys gui. I'm sure there's a simple reason that I'm missing. Also I'm not sure if this will help me exactly with what I need. I was hoping to have a way to time results, and this just seems to show traffic without any time information (again, unless I'm missing something).

rembishj:
> Also, does the visual changes in streamlit 0.75.0 for wide mode make it look better for your needs?
> 
> https://github.com/streamlit/streamlit/releases/tag/0.75.0

It definitely looks better now! Everything was stretched to the max before, but this seems to have fixed things :)

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
rembishj: I believe I've done this correctly now... my code still runs which is a good sign. Might need you to double check this one to be sure though!

SimonBiggs: > Might need you to double check this one to be sure though!

Yup, in the next PR I'll have a look :).

site-specific/mays/transfer_check_mays.py
SimonBiggs: Given Streamlit's written in Python, essentially "anything" is possible. So there certainly is a way to manipulate things into getting the timing information you want. The only question is how complicated it needs to be to make it happen.

Would you be able to detail what it is exactly you're after and I'll see if I can think of a neat solution.

SimonBiggs:
> It definitely looks better now! Everything was stretched to the max before, but this seems to have fixed things :)

Beautiful, let's bump to Streamlit v0.75.0 and for those pages where you need more width you can utilise wide mode :).

site-specific/mays/transfer_check_mays.py
rembishj: I guess what would be ideal is just a time starts as soon as the file is uploaded for the transfer check, and then the user can click a button at the bottom once they've finished to write the amount of time it took them to complete the transfer check. I believe my problem when I tried this a while ago was that the start time would change any time the user did anything to update the page.

PR #1336, opened 2021-01-22
Continued progress on weekly chart check

rembishj:


lib/pymedphys/_experimental/chartchecks/helpers.py
```diff
@@ -35,7 +39,7 @@ def get_all_dicom_treatment_info(dicomFile):
     except (TypeError, ValueError, AttributeError):
         prescriptionDescription = ""
 
-    for fraction in dicom.FractionGroupSequence:
+    for fraction in dicom[0x300A, 0x0070]:
```
SimonBiggs, line 42: Is there a reason here to change from the Dicom Tag name attribute call? I find calling dicom files by their byte tags not as easy to know what is going on.

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -16,8 +16,12 @@
 
 from pymedphys._mosaiq import connect
 from pymedphys._streamlit import categories
+from pymedphys._streamlit.utilities.mosaiq import uncached_get_mosaiq_cursor
```
SimonBiggs, line 18: Just import `get_mosaiq_cursor`, no need to copy the caching wrapper logic in :)

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -33,8 +37,14 @@
...
 
 
+@st.cache(allow_output_mutation=True, suppress_st_warning=True)
+def get_mosaiq_cursor(server):
+    return uncached_get_mosaiq_cursor(server)
+
+
 def main():
     server = "PRDMOSAIQIWVV01.utmsa.local"
+    cursor = get_mosaiq_cursor(server)
```
SimonBiggs, line 46: It's worth opening the Windows Credential Manager (https://support.microsoft.com/en-us/windows/accessing-credential-manager-1b5c916a-6a16-889f-8581-fc16e8165ac0), deleting the old credentials, and seeing if the password input is now done through Streamlit.

pyproject.toml
```diff
@@ -76,7 +76,7 @@ shapely = { version = ">=1.7.0", optional = true }  # groups = ["user", "dev"]
 
 # Streamlit's pinning is due to its current beta nature and the fact that some
 # private APIs are utilised.
-streamlit = { version = "0.74.1", optional = true }  # groups = ["user", "dev"]
+streamlit = "^0.75.0"  # groups = ["user", "dev"]
```
SimonBiggs, line 79: Replace this with:

```suggestion
streamlit = { version = "0.75.0", optional = true }  # groups = ["user", "dev"]
```

and then run `poetry run pymedphys dev propagate`

Need to have the above syntax for Streamlit for two reasons, we need to pin Streamlit to an exact version given its API is in flux, and we need to have "optional = true" so that people can install PyMedPhys without installing Streamlit if they wish :).

SimonBiggs (approved):
> succumb to british english for uniformity sake

Hehe :smile: 

Jake, that was a pleasure to review. It's nice seeing you take the refactoring ideas from some review comments and applying them elsewhere :).

lib/pymedphys/_experimental/chartchecks/helpers.py
rembishj: It's because a while back I noticed that while the DICOM format has all the same tags, they aren't always nested the same way (BeamSequence might be within FractionGroupSequence or it might be outside). I was trying something here to see if it would work and just forgot to switch back for the time being. Essentially this only works with our Pinnacle exports as far as I'm aware and will crash with another RP file because of the nesting issues so I was trying to brainstorm a way to make it more universal!

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
rembishj: Just confirmed this indeed works 👍

lib/pymedphys/_experimental/chartchecks/helpers.py
SimonBiggs: > BeamSequence might be within FractionGroupSequence or it might be outside

That's weird, I've never seen BeamSequence (https://dicom.innolitics.com/ciods/rt-plan/rt-beams/300a00b0) nested within FractionGroupSequence (https://dicom.innolitics.com/ciods/rt-plan/rt-fraction-scheme/300a0070) :/

rembishj: I may misremembering exactly what the issue was... I’ll look into it again on Monday! I just remember thinking that maybe the problem could be solved by calling tags instead is all :)

sjswerdloff: Referenced Beam Sequence.
Which has some elements in common with Beam Sequence.
Delayed binding / loosely typed language s makes it easy to accidentally treat them almost the same.

SimonBiggs: > Delayed binding / loosely typed language s makes it easy to accidentally treat them almost the same.

We certainly do have the option to create a typed DICOM wrapper around pydicom if we believe this is a sufficient issue...

sjswerdloff: I think there are benefits to having duck typing for this, and attempting to access something that isn't there should show up quickly in an automated test, unless the automated test inappropriately mocks/populates elements that are not part of the sequence item.  So... no I think we can leave it as is.

rembishj: It actually seems the issue was much simpler than I remembered... There just seems to be a Tolerance sequence missing from our institution's Raystation RP files that wasn't in the Pinnacle RP files.

PR #1342, opened 2021-01-25
setting up SQL server instance in container for testing

dg1an3:
testing to see if a SQL server instance is created when the test action runs


.github/workflows/library.yml
```diff
@@ -105,6 +105,15 @@ jobs:
 
     runs-on: ${{ matrix.os }}-latest
 
+    services:
```
dg1an3, line 108: @SimonBiggs so I guess the first thing I discovered is that the SQL server instance should only be set up to run for the test runners that run in containers (is this only the ubuntu test job)?

SimonBiggs:
A tip here, when I'm experimenting with CI I temporarily delete all other CI files and tasks. That way I don't use up all of my GitHub action machines while iterating.

Also if you edit the file within the online GitHub editor interface there is some GitHub actions specific linting that is undergone.

.github/workflows/library.yml
SimonBiggs: This particular job here is Windows, Linux and MacOS. Is there a way to make the SQLServer service start only if the `matrix.task` = 'mosaiq'?

lib/pymedphys/tests/mosaiq/test_can_we_db.py
```diff
@@ -0,0 +1,15 @@
+import pymssql
+
+
+@pytest.mark.msq_db
```
SimonBiggs, line 4: For the mark `msq_db` to work, need to add something like:

```python
    "msq_db": {
        "options": ["--msq_db"],
        "help": "run test that depend on a Mosaiq Database",
        "description": "mark test as needing a Mosaiq Database",
        "skip_otherwise": True,
    },
```

To the following options:

https://github.com/pymedphys/pymedphys/blob/7011896e99d51926e04befdde1673f871c0d693d/lib/pymedphys/conftest.py#L8-L33

pyproject.toml
```diff
@@ -138,7 +138,7 @@ readme-renderer = { version = "*", optional = true }  # groups = ["dev"]
 
 
 [tool.poetry.extras]
-user = ["tqdm", "attrs", "watchdog", "keyring", "packaging", "PyYAML", "requests", "matplotlib", "scipy", "Pillow", "imageio", "scikit-image", "pymssql", "natsort", "tomlkit", "pynetdicom", "pylibjpeg-libjpeg", "dbfread", "xmltodict", "timeago", "xlsxwriter", "plotly", "dicompyler-core", "numpy", "pandas", "pydicom", "shapely", "streamlit", "streamlit-ace", "streamlit-analytics", "pylinac", "scikit-learn"]
+user = ["tqdm", "attrs", "watchdog", "keyring", "packaging", "PyYAML", "requests", "matplotlib", "scipy", "Pillow", "imageio", "scikit-image", "pymssql", "natsort", "tomlkit", "pynetdicom", "pylibjpeg-libjpeg", "dbfread", "xmltodict", "timeago", "xlsxwriter", "plotly", "dicompyler-core", "numpy", "pandas", "pydicom", "shapely", "streamlit", "streamlit-ace", "streamlit-analytics", "pylinac", "scikit-learn", "sqlalchemy"]
```
SimonBiggs, line 143: This part of the file is autogenerated. Instead you'll want to add:

```toml
sqlalchemy = { version = "*", optional = true }  # groups = ["user", "dev"]
```

above, and then run:

```
```

SimonBiggs: You'll then need to add sqlalchemy to the following file:

https://github.com/pymedphys/pymedphys/blob/main/lib/pymedphys/_imports/imports.py

Then you should be able to import it with:

```python
from pymedphys._imports import sqlalchemy
```

This is to support optional imports within PyMedPhys. All external imports of PyMedPhys are "lazy loaded", meaning, if the user doesn't have that package installed they can still use the parts of PyMedPhys that don't need it.

SimonBiggs:
For this error:

https://github.com/pymedphys/pymedphys/pull/1342/checks?check_run_id=1772979853#step:21:219

I have just add the following two parameters to the top level exposed `pymedphys.mosaiq.connect` API:

https://github.com/pymedphys/pymedphys/pull/1352

That means `keyring` doesn't need to be used at all for these sorts of tests. We can sort out getting a CI supported keyring setup at a later date for more end-to-end testing use cases.

lib/pymedphys/tests/mosaiq/test_mock_patient_table.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+    connection = msq_connect.connect_with_credential(
+        sa_user, sa_password, msq_server, port=1433, database=test_db_name
+    )
+
+    cursor = connection.cursor()
+
+    result_all = mosaiq.execute(
+        cursor,
```
SimonBiggs, line 40: I've gone and broken this with my API change. Let me fix that.

SimonBiggs (approved):
Looks great @dg1an3. This is really beautiful to see. It's really nice to see the beginnings of the Mosaiq parts of the library being tested.

SimonBiggs:
Happy for you to merge, I made a couple of changes to update for the new API, so feel free to merge as long as you're okay with that :slightly_smiling_face:.

PR #1343, opened 2021-01-25
update transfer and weekly check

rembishj:
allow comparison to work even when field names don't match exactly, use plotly instead of matplotlib to graph couch positions in weekly check

### DIFF
--- a/lib/pymedphys/_experimental/chartchecks/helpers.py
+++ b/lib/pymedphys/_experimental/chartchecks/helpers.py
@@ -89,7 +89,6 @@ def get_all_dicom_treatment_info(dicomFile):
                 "field_label": dicom.BeamSequence[bn - 1].BeamName,
                 "field_name": dicom.BeamSequence[bn - 1].BeamDescription,
                 "machine": dicom.BeamSequence[bn - 1].TreatmentMachineName,
-                "manufacturer": dicom.BeamSequence[bn - 1].Manufacturer,
                 "rx": prescriptionDescription[fn - 1],
                 "modality": dicom.BeamSequence[bn - 1].RadiationType,
                 "position": dicom.PatientSetupSequence[0].PatientPosition,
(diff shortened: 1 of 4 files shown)


lib/pymedphys/_experimental/chartchecks/helpers.py
```diff
@@ -173,7 +172,7 @@ def get_all_dicom_treatment_info(dicomFile):
             except (TypeError, ValueError, AttributeError):
                 dicom_beam["tolerance"] = 0
 
-            if dicom_beam["manufacturer"] == "Varian":
+            if dicom_beam["machine"] == "Vault 1-IMRT":
```
SimonBiggs, line 175: It'll be worth sorting out using the config file for the sorts of things. It might be worth running through how to set up these items in a config file with you on a video call at some point.

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -15,7 +15,7 @@
 from pymedphys._imports import streamlit as st
 
 from pymedphys._streamlit import categories
-from pymedphys._streamlit.utilities.mosaiq import uncached_get_mosaiq_cursor
+from pymedphys._streamlit.utilities.mosaiq import get_mosaiq_cursor
```
SimonBiggs, line 18: Yup 😊 Nice.

SimonBiggs (approved):
LGTM :)

lib/pymedphys/_experimental/chartchecks/helpers.py
rembishj: Yes, I agree. I have been trying to keep this in mind, but definitely taken the easy route at some points.

SimonBiggs: Easy route is perfectly fine to get something up and running 🙂

PR #1344, opened 2021-01-25
UNet lung tumour case

SimonBiggs:
Hi @matthewdeancooper,

In case @sjswerdloff is occupied, would you like to take your hand at reviewing my code here in preparation for a merge?

---

This was an attempt at a "simple" lung tumour UNet. It was not successful. I've archived it away for later reference though.

---

A notebook that contours lung tumours. The goal here is to create a beginners tutorial that makes a range of assumptions to make the task quicker and easier. Key assumptions/adjustments are the following:

* Instead of finding a tumour on a full study set, it is tasked with contouring a tumour assuming there is one on that slice. Meaning, all training and validation samples are filtered down to those that do have tumours on them.
* The original 512x512 image size is scaled down to 128x128.
* Masks are not binary between 1 and 0, but instead a floating point value between 1 and 0 depending on how much of the volume encompasses that pixel.
* No data augmentation is undergone for this simple example.
* No pipeline is created, the full dataset can fit in RAM.

https://colab.research.google.com/github/pymedphys/pymedphys/blob/8656be923a8c81b9804054d128aac9d6166f93f5/prototyping/unet-lung.ipynb

---

Trying a 3D UNet:

https://colab.research.google.com/github/pymedphys/pymedphys/blob/57759d0354c9faf560e3a19802d6e48fa5641b02/prototyping/auto-segmentation/sb/05-mini-data-take-2/unet-lung.ipynb

---

None of these worked well, but I would still like to merge them into prototyping.


SimonBiggs:
Hi @sjswerdloff,

Here's this PR on nbreview:

https://app.reviewnb.com/pymedphys/pymedphys/pull/1344/files/

matthewdeancooper:
I think the way you are currently splitting is causing training data to leak into the validation data.
image

Instead.

image

SimonBiggs:
Yes it is! Arghh, rookie error. Thanks @matthewdeancooper.

matthewdeancooper:
> Yes it is! Arghh, rookie error. Thanks @matthewdeancooper.

Looks good now!

matthewdeancooper:
@SimonBiggs 

Will need to fix this up now too.

image

SimonBiggs:
Cheers @matthewdeancooper, it should have triggered the "smell test". It seemed to good to be true...

SimonBiggs:
It no longer seems to converge. I was advised that a 3D UNet architecture would likely be needed for this dataset. I wonder if some data augmentation could get it over the line without needing a 3D UNet? Any thoughts @matthewdeancooper?

SimonBiggs:
For a "hello world" example I'm going to stick to the dog CT dataset for now:

https://github.com/pymedphys/pymedphys/pull/1326#issue-556687797

Unless @matthewdeancooper, you can think of a way to make the lung dataset work without introducing significant complexity.

matthewdeancooper:
> It no longer seems to converge. I was advised that a 3D UNet architecture would likely be needed for this dataset. I wonder if some data augmentation could get it over the line without needing a 3D UNet? Any thoughts @matthewdeancooper? Given my "cheats" of only training and testing of slices that have tumour I thought that maybe I might have sufficiently created an "easy mode" for the data set... but it seems not...

The issue is likely that the tumours are very small relative to the image and that the loss places equal weight on negatives and positives. The model can achieve high accuracy by assuming tumours never exist and since we are descending by gradients, this is the dominating path to decrease loss. I haven't re-run the training but I faced similar issues with my model getting trapped in a local minimum for rectum segmentation. If we were to bias the loss to focus on false negatives then we might see some improvement.

SimonBiggs:
Hmm, if you are able to swap the loss function out and have success I'd be very interested :slightly_smiling_face:.

sjswerdloff:
> Hi @sjswerdloff,
> 
> Here's this PR on nbreview:
> 
> https://app.reviewnb.com/pymedphys/pymedphys/pull/1344/files/

I consider nbreview a big improvement for notebook review...

matthewdeancooper:
> Hmm, if you are able to swap the loss function out and have success I'd be very interested .

Happy to have a play around.

SimonBiggs:
> Happy to have a play around.

So I created a variant built on a weighted bce loss:

https://colab.research.google.com/github/pymedphys/pymedphys/blob/2f072af55d3555dd88c934e96f9f781bc70ee86f/prototyping/auto-segmentation/sb/05-mini-data-take-2/unet-lung-2d-wbce.ipynb

It doesn't appear to be converging.

I've also made a 3D UNet version:

https://colab.research.google.com/github/pymedphys/pymedphys/blob/2f072af55d3555dd88c934e96f9f781bc70ee86f/prototyping/auto-segmentation/sb/05-mini-data-take-2/unet-lung-3d-bn-and-wbce.ipynb

...I'll see if it's had any success in the morning.

sjswerdloff (changes requested):
all public functions should have docstring

matthewdeancooper:
@SimonBiggs 

Sure will have a read through now :)

prototyping/auto-segmentation/sb/05-mini-data-take-2/unet-lung-2d-bn-and-wbce.ipynb
```diff
@@ -0,0 +1,518 @@
...
+    "def decode(\n",
+    "    x,\n",
+    "    skip,\n",
+    "    number_of_filters,\n",
+    "    number_of_convolutions=2,\n",
+    "):\n",
+    "    x = _conv_transpose(x, number_of_filters)\n",
+    "    x = _activation(x)\n",
+    "\n",
+    "    x = tf.keras.layers.concatenate([skip, x], axis=3)\n",
```
matthewdeancooper, line 294: axis=-1 would mean we dont need to count to double check it is correct

SimonBiggs:
> all public functions should have docstring

There isn't any?

prototyping/auto-segmentation/sb/05-mini-data-take-2/unet-lung-2d-bn-and-wbce.ipynb
```diff
@@ -0,0 +1,518 @@
...
+    "    \n",
+    "x = tf.keras.layers.Conv2D(\n",
+    "    output_channels,\n",
+    "    1,\n",
+    "    activation=\"sigmoid\",\n",
+    "    padding=\"same\",\n",
+    "    kernel_initializer=\"he_normal\",\n",
+    ")(x)\n",
+    "\n",
+    "model = tf.keras.Model(inputs=inputs, outputs=x)"
```
matthewdeancooper, line 346: Might be nice to have a small test here

```python
batch_size = 5 
test_input = np.random.rand(batch_size, grid_size, grid_size, output_channels)
test_output = model([x])
assert test_output.shape == test_input.shape
```

prototyping/auto-segmentation/sb/05-mini-data-take-2/unet-lung-2d-bn-and-wbce.ipynb
```diff
@@ -0,0 +1,518 @@
...
+    "colab": {
+     "base_uri": "https://localhost:8080/",
+     "height": 1000
+    },
+    "id": "0TNPQy5AJLRg",
+    "outputId": "b5235583-0a18-432c-e493-48a569349519"
+   },
+   "outputs": [],
+   "source": [
+    "history = model.fit(\n",
```
matthewdeancooper, line 434: If this is for demonstration purposes, we can also test if the model has enough degrees of freedom to overfit a subset of our data.

```python
sub_index = 10
overfit_test = model.fit(
    training_images[:sub_index], 
    training_masks[:sub_index],
    epochs=500,
    validation_data=(validation_images, validation_masks)
```

matthewdeancooper (approved):
Good to go. Just minor suggestions.

matthewdeancooper:
BTW, very cool that Colab lets you use a GPU with ~ 15 G of ram.

SimonBiggs:
> Good to go. Just minor suggestions.

Yup, some good calls there. I'll make those changes, just to the 2D unet version, as that's the one I'll fly with initially. Given Stuart has specifically placed the "change requested" flag I won't merge until he removes that.

SimonBiggs:
> BTW, very cool that Colab lets you use a GPU with ~ 15 G of ram.
> 
> 

...yeah... it's pretty awesome ... ...tiding me over until I purchase a 2x3090 version of one of these:

https://lambdalabs.com/gpu-workstations/vector

matthewdeancooper:
> > Happy to have a play around.
> 
> So I created a variant built on a weighted bce loss:
> 
> https://colab.research.google.com/github/pymedphys/pymedphys/blob/2f072af55d3555dd88c934e96f9f781bc70ee86f/prototyping/auto-segmentation/sb/05-mini-data-take-2/unet-lung-2d-wbce.ipynb
> 
> It doesn't appear to be converging.
> 
> I've also made a 3D UNet version:
> 
> https://colab.research.google.com/github/pymedphys/pymedphys/blob/2f072af55d3555dd88c934e96f9f781bc70ee86f/prototyping/auto-segmentation/sb/05-mini-data-take-2/unet-lung-3d-bn-and-wbce.ipynb
> 
> ...I'll see if it's had any success in the morning.
> 
> Hmm, some success. But nothing much really. Two pathways I see forward:
> 
> * Pull in data augmentation
>   
>   * Likely worth pulling this in first, as this is the "least complicated" path forward.
>   * If this works with the 2D UNet we are still within the scope of a "simplish" notebook.
> * Utilise full resolution lung scans but undergo 3D "patch based" training.
>   
>   * Potentially the lack of convergence is because I have made it harder by making the images low-res
>   * Patch based 3D UNets have had success in the literature by randomly defining points within datasets, then taking a 3D grid around that point, let's say with dimensions 64x64x64, and then making sure that the weighting of patches utilised for training have 50% with tumour and 50% without (so, a fraction of non-tumour patches are discarded to aid training). This hits a few issues:
>     
>     * The patch size can be adjusted to account for the computational power of the training machine, and it can be tuned in the z direction independently of the x/y directions so as to make it more or less "3D",
>     * The detail of the image can be full quality while still retaining the ability to run on 'lesser' hardware
>     * Partially addresses the tumour class imbalance weighting issues

I think the way forward fundamentally depends on where the issue lies and the goals of this project (in terms of complexity). Data augmentation is a type of regularisation. It increases the spread of your distribution to increase the effective size; however, we are really trying to teach the model to be invariant to the types of transformations it will see in real use. If you see training metrics improve while validation gets worse then its a hint that what the model has learned is training data specific - regularisation could help here.

Dropout(p=0.2) would be my first attempt at regularising as it is easy to implement, although, it is usually used in conjunction with data augmentation.

Additionally, if using class weights already then there may be little benefit to discarding data to achieve a 50/50 distribution.

prototyping/auto-segmentation/sb/05-mini-data-take-2/unet-lung-2d-bn-and-wbce.ipynb
```diff
@@ -0,0 +1,518 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "id": "FGU8fYdkJLRc"
+   },
+   "outputs": [],
+   "source": [
+    "def load_dataset_type(directories, shuffle=True):\n",
+    "    image_suffix = '_image.png'\n",
```
sjswerdloff, line 113: so, this to me is a public function.
And it has no docstring.
But maybe it's not possible to docstring public functions within Jupyter Notebooks.

sjswerdloff (approved):

prototyping/auto-segmentation/sb/05-mini-data-take-2/unet-lung-2d-bn-and-wbce.ipynb
SimonBiggs: Yeah, it's not very reasonable to be able to import them from another python script. So their not "exported" or "public" in that sense.

There are ways that people do it, by hacking at Python's import logic... but that is not something I am supporting here...

SimonBiggs:
> Good to go. Just minor suggestions.

Hi @matthewdeancooper, I'm going to implement these within the animal dataset notebook. Given that notebook isn't in this PR I'll merge this PR and spin out these comments into its own issue:

https://github.com/pymedphys/pymedphys/issues/1351

PR #1346, opened 2021-01-26
[STACK wide-mode] Add wide mode to App, utilise columns, make logo be SVG based

SimonBiggs:
Pre-commit issue fixed in https://github.com/pymedphys/pymedphys/pull/1348

---

@rembishj did you want to try your hand at reviewing this one? This is me swapping the App over to wide-mode to support your usage of it :) (and also make our clinical dashboard easier to use).

Then given I was touching the "looks" of the app, I also fixed up the logo. ...and made a slight mess of things by upgrading black :/... but I fixed that in another PR.


lib/pymedphys/_streamlit/index.py
```diff
@@ -99,7 +118,7 @@ def _get_apps_from_module(module):
 
 
 def main():
-    st.set_page_config(page_title="PyMedPhys", page_icon=FAVICON)
+    st.set_page_config(page_title="PyMedPhys", page_icon=FAVICON, layout="wide")
```
rembishj, line 121: If only there was a way to set this for each individual app... It seems you can only do it for the overall project. But overall, I think wide-mode should be better.

rembishj (approved):
Looks good! :)

SimonBiggs:
> Looks good! :)

Awesome, thanks @rembishj.

If you had it in you, https://github.com/pymedphys/pymedphys/pull/1347 and https://github.com/pymedphys/pymedphys/pull/1348 would also be awesome to have a review.

That first one is making the MetersetMap GUI use columns, the second one isn't actually changing anything, it's just running the updated version of `black`.

PR #1347, opened 2021-01-27
[STACK columns -> wide-mode] Meterset Map Columns

SimonBiggs:
Pre-commit issue fixed in https://github.com/pymedphys/pymedphys/pull/1348


rembishj (approved):
This one looks good too! Is there any way to scale down the size of the plots at the end though? I feel it's a bit too in your face in the wide mode

SimonBiggs:
> I feel it's a bit too in your face in the wide mode

I agree. If only there was a "centred column" option within the `columns` API...

PR #1349, opened 2021-01-27
Refactor Mosaiq Login

SimonBiggs:
When expanding Mosaiq connect to include the database option the API of Mosaiq connect showed its brittleness. Since server and port are passed via a combined string no benefits of typing/linting are seen. Even after the previous refactor, the port was dropped off the connection logic.

This refactor aims to address the root cause of that miss-step, the fact that the inputs to connect mangled both hostname and port into the one parameter.

---

I'd be keen for your thoughts as I go along @dg1an3, as you have most recently been involved in wading through this code of mine.


lib/pymedphys/_mosaiq/connect.py
```diff
@@ -17,50 +17,28 @@
...
+from . import credentials as _credentials
 
 
 def connect_with_credential(
-    sql_server_and_port, username, password, database="MOSAIQ"
+    username: str,
+    password: str,
+    hostname: str,
+    port: int = 1433,
+    database: str = "MOSAIQ",
```
dg1an3, line 30: is the order of the new parameters ok?  it seems like it would be less confusing if the order of the new parameters were: hostname, port, username, password, database, because then they would be the same relative order as before.  or am I over-thinking it?

dg1an3 (approved):

lib/pymedphys/_mosaiq/connect.py
SimonBiggs: > hostname, port, username, password, database, because then they would be the same relative order as before. or am I over-thinking it?

I agree that shooting for consistency is ideal. However, during PyMedPhys 0.*.* Beta stage if there are ways to unify the API across the library I have been prioritising that over invoking breaking changes.

I tried to build out a model of grouping "server parameters" which were consistent between many functions all at the end.

SimonBiggs:
Heya @rembishj,

I've made a few breaking changes to how Mosaiq login works. I've updated your files that used the API, so hopefully it should "just work". Can you checkout this branch locally and make sure I haven't broken anything of yours before I merge it?

lib/pymedphys/_mosaiq/api.py
```diff
@@ -0,0 +1,138 @@
...
+    ----------
+    hostname : str
+        The IP address or hostname of the SQL server.
+    port : int, optional
+        The port at which the SQL server is hosted, by default 1433
+    database : str, optional
+        The MSSQL database name, by default "MOSAIQ"
+    alias : Optional[str], optional
+        A human readable representation of the server, this is the name
+        of the server presented to the user should their not be
```
SimonBiggs, line 48:
```suggestion
        of the server presented to the user should there not be
```

lib/pymedphys/_mosaiq/api.py
```diff
@@ -0,0 +1,138 @@
...
+    hostname: str,
+    port: int = 1433,
+    database: str = "MOSAIQ",
+    alias: Optional[str] = None,
+) -> "pymssql.Cursor":
+    """Connect to a Mosaiq SQL server.
+
+    The first time running this function on a system will result in a
+    prompt to login to the Mosaiq SQL server. The provided credentials
+    will be stored within the operating systems password storage
```
SimonBiggs, line 34:
```suggestion
    will be stored within the operating system's password storage
```

lib/pymedphys/_mosaiq/api.py
```diff
@@ -0,0 +1,138 @@
...
+    """Execute SQL queries on a Mosaiq database.
+
+    Parameters
+    ----------
+    cursor : pymssql.Cursor
+        A database cursor. This can be retrieved by calling
+        ``pymedphys.mosaiq.connect``
+    query : str
+        The SQL query to execute. Do not parse Python variables directly
+        into this string. Instead utilise create this string as if you
```
SimonBiggs, line 80:
```suggestion
        into this string. Instead create this string as if you
```

rembishj:
> Heya @rembishj,
> 
> I've made a few breaking changes to how Mosaiq login works. I've updated your files that used the API, so hopefully it should "just work". Can you checkout this branch locally on your machine and make sure I haven't broken anything of yours before I merge it?

After making the one little change, it seems my app is still working properly. Is there a reason you didn't leave the argument as sql_string and switched to query instead? Not sure if that will break anything elsewhere

lib/pymedphys/_experimental/chartchecks/weekly_check_helpers.py
```diff
@@ -28,62 +28,64 @@
...
-        todays_date = date.today() + timedelta(days=3)
-        todays_date = todays_date.strftime("%b %d, %Y")
-        # todays_date = "Dec 4, 2020"
-        incomplete = incomplete[
-            (incomplete["task"] == "Weekly Chart Check")
-            & (incomplete["due"] == todays_date)
-        ]
-        incomplete = incomplete.drop(columns=["instructions", "task", "due", "comment"])
-        incomplete = incomplete.reset_index(drop=True)
+    cursor = _pp_mosaiq.connect("PRDMOSAIQIWVV01.utmsa.local")
```
dg1an3, line 31: @rembishj @SimonBiggs are you intending to have a hardcoded server name here?

SimonBiggs: It's certainly now how it needs to be in the long term. But we have a bit of a tiered quality bar within PyMedPhys. Given this is scoped under the 'experimental" package essentially "anything goes that doesn't break the build".

I am hoping to at some point go through a video call with @rembishj and show him the ropes on how to utilise site configuration for these sorts of things. But, for the time being he has other PhD demands that are taking precedence. For now I'm happy if he's getting something to work that works within his centre.

Also, it's not just @rembishj, I too have some experimental items where I have (for now) left the configuration hard coded:

https://github.com/pymedphys/pymedphys/blob/5c10f54d167f9e6416a60acbb300275d3deb60cd/lib/pymedphys/_experimental/streamlit/apps/dashboard.py#L30-L41

rembishj: Yep, what Simon said. I would like to make it more universal, but just trying to get it working first and then focus on that.

SimonBiggs:
> After making the one little change, it seems my app is still working properly. Is there a reason you didn't leave the argument as sql_string and switched to query instead? Not sure if that will break anything elsewhere

I had already broken a few things with the API (no longer supporting "with" syntax, changing the port input method, etc) I figured I would refine other parameters to. I didn't like that "string" was in the name, that's a form of "Hungarian typing" and the typing information is already in the Python type hint, no need to double it up in the name. At the end of the day it is a "query" so, opted that way. Happy for another name if you have a preference :).

> also, this was my first time checking out someone elses branch and pushing my own commits. Hope I did it correctly!

Yup 😊 perfect! 🎉

PR #1355, opened 2021-01-28
Remove electron factors from API

SimonBiggs:
There has been negligible interest in the module, the API was riddled with spelling mistakes, and in the future the approach can be exposed via the application. If there is interest to have the code exposed once again this can be undergone after refining the API.


lib/pymedphys/docs/lib/ref/index.rst
```diff
@@ -12,5 +12,4 @@ itself. Below is the documentation for a range of these libraries.
     dicom
     gamma
     metersetmap
-    electronfactors
```
sjswerdloff, line 15: shouldn't the documentation be moved to experimental rather than deleted?
Alternatively, if you are truly deleting electronfactors, shouldn't it be removed from streamlit apps?
Maybe there is a different PR that did some of this?

lib/pymedphys/_experimental/streamlit/apps/electrons.py
```diff
@@ -20,7 +20,7 @@
 from pymedphys._imports import plt
 from pymedphys._imports import streamlit as st
 
-import pymedphys.electronfactors as electronfactors
+import pymedphys._electronfactors as electronfactors
```
sjswerdloff, line 23: this suggests you are moving electronfactors to experimental?

sjswerdloff (changes requested):
If this is just the first step in removal, let me know and I'll approve.
But it looks like this is half "move to experimental" and half "delete".
I would hate to "lose" documentation for something that isn't deleted.
I could see electron factors becoming of interest with electron FLASH treatment in the future

SimonBiggs:
> But it looks like this is half "move to experimental" and half "delete".

It's not a move to experimental, just a removal from the public API. The electrons code shouldn't go to experimental, the code itself is well utilised and "mature". It's just that it doesn't appear like it's being used anywhere except at RCCC. The only thing that is really an issue is the current exposed API.

> I could see electron factors becoming of interest with electron FLASH treatment in the future

From what I understand with flash... I doubt they'll be shaping the beam with LMPA cutouts...

> I would hate to "lose" documentation for something that isn't deleted.

The documentation isn't being lost :) It's still all there as docstrings. And one day, if someone does shows interest, it can come back :).

sjswerdloff (approved):

PR #1350, opened 2021-01-27
Docs work (categorise into App Users, Lib Users & Contributors)

Matthew-Jennings:
Closes #1334

PR #1356, opened 2021-01-28
Fix readme links

SimonBiggs:
Also :slightly_smiling_face:, there's another part of the README that may need adjustment ...

https://github.com/pymedphys/pymedphys/blob/d978400c215b8282dcc61a9b3d26dba3ff99fb1a/README.rst#L145-L151

:wink: :wine_glass: :rocket: [... thinks about emoticon choices :thinking: ... maybe it's best not to drink and take a rocket at the same time ...]


Matthew-Jennings:
Is there a way for me to review the new docs on my phone? Is there an online render?

SimonBiggs:
> Is there a way for me to review the new docs on my phone? Is there an online render? (I'm away from any PC...)

I can making a "docs-staging.pymedphys.com" which is pushed to for every PR? We would need some way to signify that a specific PR/commit should update the staging docs otherwise two PRs being worked on at once will cause the staging docs to oscillate. Maybe of the branch name has "docs" in it, then any push to that branch will deploy to "docs-staging.pymedphys.com"? What do you think?

SimonBiggs:
I'm going to merge this one as I don't want the readme links to remain broken.

Matthew-Jennings:
> I can making a "docs-staging.pymedphys.com" which is pushed to for every PR? We would need some way to signify that a specific PR/commit should update the staging docs otherwise two PRs being worked on at once will cause the staging docs to oscillate. Maybe if the branch name has "docs" in it, then any push to that branch will deploy to "docs-staging.pymedphys.com"? What do you think?

Sounds like a reasonable idea! It's probably not critical. I've just been spending a fair amount of time away from home lately, and I wanted to still provide a useful review. It's just hard to do so for docs stuff without being able to see the rendered version. Not sure if it'll happen often enough to bother?

lib/pymedphys/docs/README.rst
```diff
@@ -197,8 +197,8 @@ Past contributors
 
 .. _`Release Notes`: ./CHANGELOG.md
 
-.. _`App Users Guide`: https://docs.pymedphys.com/app/index.rst
-.. _`Library Users Guide`: https://docs.pymedphys.com/lib/index.rst
-.. _`CLI Users Guide`: https://docs.pymedphys.com/cli/index.rst
-.. _`Contributors Guide`: https://docs.pymedphys.com/contrib/index.rst
-.. _`General`: https://docs.pymedphys.com/general/index.rst
+.. _`App Users Guide`: https://docs.pymedphys.com/app/index.html
```
Matthew-Jennings, line 200: Sorry about this! I kinda blindly copied from [this line](https://github.com/pymedphys/pymedphys/pull/1350/files#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168fL283).

SimonBiggs: No need to be sorry 🙂 I reviewed it too 🙂.

Bound to be a few hiccups when this level of re-structuring is undergone.

PR #1358, opened 2021-01-28
clean up transfer check helpers

rembishj:


lib/pymedphys/_experimental/chartchecks/helpers.py
```diff
@@ -30,6 +30,43 @@ def _invert_angle(angle):
     return (180 - angle) % 360
 
 
+def get_dicom_wedge_info(beam_reference, field):
+
+    for cp in field.ControlPointSequence:
+        if cp.WedgePositionSequence[0].WedgePosition == "OUT":
```
SimonBiggs, line 36: Above this code I would recommend having the following:

```python
if len(cp.WedgePositionSequence) != 1:
    raise ValueError("Expected only one wedge sequence per control point")
```

As a rule of thumb, before indexing to `[0]` always put a test like that above it as you are inherently making an assumption about the format of the data. If that assumption is broken, you want the application to fail, kicking and screaming.

lib/pymedphys/_experimental/chartchecks/helpers.py
```diff
@@ -30,6 +30,43 @@ def _invert_angle(angle):
...
+        "coll_x2": [0, 1],
+        "coll_y1": [1, 0],
+        "coll_y2": [1, 1],
+    }
+
+    colls = {}
+    for key, value in keys.items():
+        colls[key] = (
+            field.ControlPointSequence[0]
+            .BeamLimitingDevicePositionSequence[value[0]]
```
SimonBiggs, line 62: The order here isn't always consistent. Instead, you need to look to a seperate field to see which index corresponds to which collimation device.

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -75,12 +75,12 @@ def main():
...
-        dicom_table = get_all_dicom_treatment_info(files["rp"])
-        dicom_table = dicom_table.sort_values(["field_label"])
-        # except AttributeError:
-        #     st.write("Please select a new RP file.")
-        #     st.stop()
+        try:
+            dicom_table = get_all_dicom_treatment_info(files["rp"])
+            dicom_table = dicom_table.sort_values(["field_label"])
+        except AttributeError:
+            st.write("Please select a new RP file.")
```
SimonBiggs, line 82: To give a bit more detail in streamlit here (if that is desired) you can do:

```python
except AttributeError as e:
    st.write(e)
    st.write("Please select a new RP file.")
    st.stop()
```

SimonBiggs (approved):
These are really nice chunks to review @rembishj 🙂. Also, it is quite nice seeing your coding style improving as we go along 🙂.

rembishj:
> These are really nice chunks to review @rembishj 🙂. Also, it is quite nice seeing your coding style improving as we go along 🙂.

Thank you! It might've taken awhile, but I feel I'm finally learning to make my code a little neater and less janky 😄

lib/pymedphys/_experimental/chartchecks/helpers.py
rembishj: Ahhh, yes this is a good call. I will make a change.

lib/pymedphys/_experimental/chartchecks/helpers.py
rembishj: Any ideas where I might be able to find that information off the top of your head? If not, I will do some searching!

SimonBiggs: Not off the top of my head sorry 🙁 ... I do have code somewhere in PyMedPhys that deals with it though... Not available to find that at the moment though.

PR #1363, opened 2021-01-29
Make internet installer `PyMedPhys-Setup.exe`

SimonBiggs:
Made the [`PyMedPhys-Setup.exe`](https://github.com/pymedphys/pymedphys/releases/download/v0.36.0-dev2/PyMedPhys-Setup.exe) into an internet based installer that calls the Python file found within https://bootstrap.pymedphys.com/get-pymedphys.py using the Python environment defined within https://bootstrap.pymedphys.com/python-urls/win-amd64.

This should allow us maximum flexibility and allow us to keep a static `PyMedPhys-Setup.exe` installer. Which means we can make a once off, one-year, purchase of an EV code-signing certificate to sign that exe file and make it so that the executable is trusted as coming from the "PyMedPhys" publisher. As long as the file is signed with a timestamp, even when the certificate expires it will still be able to be trusted.

The cheapest EV code signing certificate I could find was $350 USD for one year:

https://www.ssl.com/certificates/ev-code-signing/buy/


SimonBiggs:
I'll keep this code in the tree for now. I might want to re-use some of it... but not sure how much I'll be doing that after more thought.

SimonBiggs:
Just throwing a few thoughts "on paper".

---

What about not having an .exe file at all? What about distributing a .zip file that contains a Python distro within it, and a .bat file that runs the install. Something like:

```
PyMedPhys-v0.36.0.zip/
|    INSTALL-PyMedPhys-v0.36.0.bat
|    INSTALL-PyMedPhys-latest.bat
|
|--  PyMedPhys-v0.36.0/
|    |    start-pymedphys-app.bat
|    |    run-app-tests.bat
|    |    run-library-tests.bat
|    |    LICENSE
|    |
|    |--  bin/
|    |    |    pymedphys.bat
|    |    |
|    |--  embedded-python/
|    |    |    python.exe
|    |    |    ...
|    |    |
|    |--  custom/
|    |    |--  apps/
|    |    |    |    an-example-user-customisable-streamlit-app.py
|    |    |    |    another.py
|    |    |    |    ...
|    |    |    | 
|    |    |--  tests/  # will run when run-app-tests.bat is called
|    |    |    |    an-example-user-customisable-streamlit-app.js
|    |    |    |    another.js
|    |    |    |    ...
```

What INSTALL-PyMedPhys-v0.36.0.bat does is the following:
Copies PyMedPhys-v0.36.0 to %USERPROFILE%/.pymedphys/versions/PyMedPhys-v0.36.0
Adds %USERPROFILE%/.pymedphys/installations/PyMedPhys-v0.36.0/bin to the user path
And deletes other %USERPROFILE%/.pymedphys/installations/*/bin path entries if they exist
Adds a shortcut with a pretty icon to the desktop to the GUI booter

---

Or, what if I go to the original model where there isn't a `Setup.exe` but instead a `PyMedPhys-App.exe`. And instead of "installing" the "installation" is thought as more of a "cache".

When the `PyMedPhys-App.exe` runs it checks the user's home drive to see if an environment has been already been cached. If something is cached then it boots that up. After booting up and the app is running, periodically the app will auto-query the server to see if there is an update. If there is an update, the new PyMedPhys version will download to the cache in the background. Then a prompt will appear within the Streamlit app saying an update is available. That prompt can be clicked and the Streamlit server will reboot using the new version. Make it so that this "auto-update" functionality can be switched off. Also, allow users to use the Streamlit interface to select a PyMedPhys version and change and be frozen on that version.

At the end of the day, the zip directory tree mapped out above would be the very same versioned zip that "PyMedPhys-App.exe" would run off and download. So someone who wants just a plain old offline only install can use that zip file. But the default supported installation would be via "PyMedPhys-App.exe".

sjswerdloff (approved):
The perfect is the enemy of the good.
And this looks good.

PR #1362, opened 2021-01-29
Clarify lib purpose in docs, make propagate and pre-commit commit the fixes to the issues

SimonBiggs:

PR #1364, opened 2021-01-29
adding sqlalchemy to pyproject.toml and imports

dg1an3:
@SimonBiggs I was a little confused about how to add sqlalchemy to the imports, so I separated it.

These are the two initial changes to add sqlalchemy.  After this, when I run
`poetry run pymedphys dev propagate`
on my local repo there are quite a few changes to requirements.txt, requirement-deploy.txt, etc.

am I supposed to then commit those changes as well?


SimonBiggs:
> am I supposed to then commit those changes as well?

Normal rule of thumb with version control is to not commit build artefacts. So normally, no. But here a trade off has been taken where, for the benefit of someone being able to use files that they're familiar with (requirements.txt) the down side of having to commit extra cruft is accepted.

So, yup :slightly_smiling_face: commit away :slightly_smiling_face:.

In fact, if you wanted to review the following PR:

https://github.com/pymedphys/pymedphys/pull/1362

I made it so that propagate and pre-commit is automatically handled by the CI. Meaning, if either isn't done, then the CI commits that for you. So that might be nice to merge and then merge into this one and see if it works nicely :slightly_smiling_face:.

dg1an3:
It seems to have done the propagation correctly! :)

SimonBiggs (approved):
LGTM :)

PR #1367, opened 2021-01-31
Fix propagate conflicts with CI

SimonBiggs:
The goal here is that should there be any merge conflicts that are just due to `pymedphys dev propagate` then those conflicts are handled automatically and merged back in.


.github/workflows/conflicts.yml
```diff
@@ -0,0 +1,67 @@
...
+        uses: snok/install-poetry@v1.1.1
+        with:
+          version: 1.1.4
+
+      - name: Install PyMedPhys with propagate environment
+        if: steps.conflicts.outputs.status == 'true'
+        run: |
+          poetry install -E propagate
+
+      - name: Fix conflicts
```
sjswerdloff, line 47: Can you provide a description of what this is doing to resolve the conflicts other than overwriting with the latest?  Does this error out if there is a 'real' conflict. What 'not really a conflict' is this addressing?

SimonBiggs: Can do, so at this point here:

https://github.com/pymedphys/pymedphys/blob/718638a4211d72bea5ffe7a4a30f484080f6fd80/.github/workflows/conflicts.yml#L53-L54

The main branch has been merged and any conflicting files are set to untracked.

https://github.com/pymedphys/pymedphys/blob/718638a4211d72bea5ffe7a4a30f484080f6fd80/.github/workflows/conflicts.yml#L56

Here, the propagation artefacts are set to match with the current branch. This is required because poetry commands can't run when there are merge conflicts within the lock files.

Then, the "deterministic" propagate is run:

https://github.com/pymedphys/pymedphys/blob/718638a4211d72bea5ffe7a4a30f484080f6fd80/.github/workflows/conflicts.yml#L57

And here is where the issue lies. This command was run on something within main branch, and it has been run on something else within the branch in question. Therefore it actually needs to be run on the combination of main and the current branch. So that's what is happening here. This will overwrite those files that were set above.

At this point only the propagation artefacts are added to the staging area (any other conflicting files that are not propagation artefacts would not be added here).

https://github.com/pymedphys/pymedphys/blob/718638a4211d72bea5ffe7a4a30f484080f6fd80/.github/workflows/conflicts.yml#L59-L60

That commit in that last line there will fail if there still remains any conflicting files.

So...

> Does this error out if there is a 'real' conflict. 

Yup

> What 'not really a conflict' is this addressing?

Deterministic propagation artefacts.

.github/workflows/conflicts.yml
```diff
@@ -0,0 +1,67 @@
...
+        if: steps.conflicts.outputs.status == 'true'
+        run: |
+          git config --local user.email "merge-conflicts-bot@pymedphys.com"
+          git config --local user.name "PyMedPhys Merge Conflicts Bot"
+
+          git fetch origin main
+          git merge origin/main || true
+
+          git checkout HEAD poetry.lock setup.py requirements* pyproject.hash
```
SimonBiggs, line 56:
```suggestion
          git checkout HEAD poetry.lock
```

.github/workflows/conflicts.yml
SimonBiggs: I opted to make an adjustment. Instead of reverting the main's version of the lock file I have instead opted to use the following:

https://github.com/python-poetry/poetry/issues/496#issuecomment-588521241

As, I suspect you may have been hinting at, I previously was just throwing out main's copy of this.

PR #1370, opened 2021-01-31
Don't have propagate bot merge into main

SimonBiggs:
Key here, is it used to do the following:

image

This would pull in the current main into the PR. ...we don't want that. So this change now makes that not happen anymore:

image


.github/workflows/library.yml
```diff
@@ -353,27 +353,29 @@ jobs:
           git config --local user.email "propagate-bot@pymedphys.com"
           git config --local user.name "PyMedPhys Propagate Bot"
 
+          git fetch origin $BRANCH
+          git checkout $BRANCH
```
sjswerdloff, line 357: and this is how it avoids the merge in to main, the commit is now going in to $BRANCH
If that's the case, great.
If not, some explanatory text would be useful.

.github/workflows/library.yml
```diff
@@ -353,27 +353,29 @@ jobs:
...
 
           git commit -m "Fix propagate failures" -a
-          exit 1
-        env:
-          BRANCH: ${{ github.event.pull_request.head.ref }}
 
       - name: Push changes
-        if: matrix.task == 'propagate' && failure() && steps.propagate-test.outputs.status == 'committed'
+        if: steps.propagate-test.outputs.status == 'failed'
         uses: ad-m/github-push-action@057a6ba835d986bfe495dd476a6c4db1d5f9503c
```
sjswerdloff, line 379: is this just git push ?
seems strange to be making use of a "3rd party action".
In general, I think a "uses" warrants some descriptive text, although I suppose I could copy, paste in to a search engine, and find out what it is.
OK.  Did my search, I now know what it is.
But it's not the most recent release version, as far as I can tell.
So... which version is it, and why not use the latest release version?

sjswerdloff (approved):

.github/workflows/library.yml
SimonBiggs: > and this is how it avoids the merge in to main, the commit is now going in to $BRANCH

Yup 🙂, that's the case.

.github/workflows/library.yml
SimonBiggs: > But it's not the most recent release version, as far as I can tell.

I didn't want to pin to master so I just picked the latest when I looked at it. But I agree it is hard to look at a commit string. I wonder if calling a tag instead works:

```suggestion
        uses: ad-m/github-push-action@v0.6.0
```

SimonBiggs: > seems strange to be making use of a "3rd party action".

The key part it is handling is the GitHub token usage. I spose I would need to set up ssh within the action with that API token as the key? Or something like that. Instead, here I opted to go with a pre-packaged solution.

PR #1372, opened 2021-01-31
cleaning up mosaiq-db-tests.yml; adding mosaiq environment flag

dg1an3:
resolving some issues from previous PR to make mosaiq db test workflow more efficient


.github/workflows/mosaiq-db-tests.yml
```diff
@@ -5,17 +5,12 @@ on:
     branches:
       - main
   pull_request:
```
dg1an3, line 7: this seems weird (to have the pull_request with a colon, but nothing below) -- is this the right syntax?

SimonBiggs: Yeah, it is a bit weird. But yup, that's the syntax:

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration

SimonBiggs (approved):
LGTM :) Feel free to merge once you're happy with it.

SimonBiggs (changes requested):
Actually, since the CI fix hasn't merged yet, we actually do have a few failures introduced by this:

https://github.com/pymedphys/pymedphys/pull/1372/checks?check_run_id=1802528572#step:17:28

Let me fix that up.

SimonBiggs:
I've stacked this PR on top of the as yet un-merged branch that fixes the tests-not-failing issue.

SimonBiggs (approved):
Alright, so this PR blew up a little. Essentially, what I had to do was step through all the code that was being imported when the tests were starting up and make sure that every import pulled from `pymedphys._imports`... There were a lot of places where that wasn't happening within some of the internal parts of the library.

...so that's why the changelog here blew up.

When you're happy with the extra changes I've put in here, feel free to merge :).

dg1an3:
@SimonBiggs out of curiosity, is the check that imports are being done correctly something that pylint can scan for?  I'm not really familiar with what it is capable of doing...

SimonBiggs:
> @SimonBiggs out of curiosity, is the check that imports are being done correctly something that pylint can scan for? I'm not really familiar with what it is capable of doing...

Yeah, I hadn't thought if that. It certainly should be able to. PyLint has a plugin facility:

https://docs.pylint.org/en/1.6.0/plugins.html

And pylint is built on astroid:

https://astroid.readthedocs.io/en/latest/

Which can staticly parse a Python file and represent it in "Abstract Syntax Trees" (AST).

Here's an example where I use AST to build out a dependency tree:

https://github.com/pymedphys/pymedphys/blob/5d9a3620b660b09896ecc3c1afa0d140b710a045/lib/pymedphys/_experimental/tree.py#L262-L310

Here I use AST to set up that optional import logic:

https://github.com/pymedphys/pymedphys/blob/5d9a3620b660b09896ecc3c1afa0d140b710a045/lib/pymedphys/_imports/_parse.py#L1-L39

I expect it would certainly be doable to parse for import logic, if the import in question is found within https://github.com/pymedphys/pymedphys/blob/main/lib/pymedphys/_imports/imports.py then it's not allowed to be imported directly, but instead via pymedphys._imports.

The harder bit is that those import functions then are not allowed to utilised "at import time". There are a range of things that happen at import time, function and class definitions and their signatures, top level constants, and function decorators. Any of those tasks can't utilise the optional import modules otherwise they will be force loaded at import time.

PR #1373, opened 2021-01-31
Make online app be on schedule only

SimonBiggs:
Sorry for all the pings that are about to happen ... don't feel like that means they have to be merged soon...

### DIFF
--- a/.github/workflows/online-app.yml
+++ b/.github/workflows/online-app.yml
@@ -1,10 +1,6 @@
 name: OnlineApp
 
 on:
-  push:
-    branches:
-      - main
-  pull_request:
   schedule:
     - cron: '*/5 * * * *'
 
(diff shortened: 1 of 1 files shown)


sjswerdloff (approved):
Why?
What is the purpose of the action in the first place, and why should it be on a schedule, and why should it not be performed when there is a push or pull?
You could probably answer all of that *in* the YAML file.

.github/workflows/online-app.yml
```diff
@@ -1,10 +1,6 @@
 name: OnlineApp
```
SimonBiggs, line 1:
```suggestion
# A test which runs the e2e cypress tests on the online hosted version of the PyMedPhys app
# which is at https://app.pymedphys.com. This test is on a schedule so that there is a
# track record of the application running with time.

# It does not make sense to run it on PR or push as any particular PR or push won't yet actually
# be live on the web-app. This test is independent of any given branch.

name: OnlineApp
```

SimonBiggs:
> Why?
> What is the purpose of the action in the first place, and why should it be on a schedule, and why should it not be performed when there is a push or pull?
> You could probably answer all of that _in_ the YAML file.

Good call. Answered in https://github.com/pymedphys/pymedphys/pull/1373/files#r567575098

PR #1374, opened 2021-01-31
[STACK base] Fix lack of error_code being returned by `pymedphys dev tests`, then clean up the fallout...

SimonBiggs:
The failure ~~isn't~~ wasn't being picked up by the CI:

https://github.com/pymedphys/pymedphys/pull/1374/checks?check_run_id=1802593908#step:39:182

---

Here is the commit that made it so that the tests always passed:

https://github.com/pymedphys/pymedphys/commit/347beb8582967b927d6314255def6093d83f529d

---

All changes since that commit:

https://github.com/pymedphys/pymedphys/compare/347beb8582967b927d6314255def6093d83f529d...main

---

Now to find the source of this:

https://github.com/pymedphys/pymedphys/runs/1802643290?check_suite_focus=true#step:40:2080

...

---

Alright... ...fixed :/... @sjswerdloff and @Matthew-Jennings this one probably needs a post mortem...


SimonBiggs:
@rembishj, heads up, I had to change some of your code in experimental, because in the form it was it was actually breaking the tests ... it's just for 2 weeks the tests CI actually hasn't been erroring if there was an issue :/

SimonBiggs:
Also @dg1an3, the issue being fixed here would have made making your new Mosaiq CI quite a weird experience... as it was always producing a tick, even when the tests failed...

dg1an3:
> 
> 
> Also @dg1an3, the issue being fixed here would have made making your new Mosaiq CI quite a weird experience... as it was always producing a tick, even when the new tests failed...

I thought I was imagining that, but it did seem like the tests were reported as green even when there was a failure

lib/pymedphys/_experimental/chartchecks/test2.py
```diff
@@ -2,8 +2,14 @@
 
 from pymedphys._experimental.chartchecks.helpers import get_all_dicom_treatment_info
 
-file = "C:/Users/rembishj/Documents/raystation_plan/RP1.2.752.243.1.1.20210123122119948.1140.13064.dcm"
 
-plan = get_all_dicom_treatment_info(file)
+def main():
+    file = "C:/Users/rembishj/Documents/raystation_plan/RP1.2.752.243.1.1.20210123122119948.1140.13064.dcm"
```
sjswerdloff, line 7: this test can't pass anywhere but in the specific user's environment.
suggest ensuring that it is anonymised or pseudonymised or has no PHI, and load up to zenodo, and then access via zenodo.

sjswerdloff (changes requested):
See comment regarding inaccessible data.
Also, why are the tests in the code space itself rather than in the tests/ tree?
I'll accept "not intended as a unit test, this is cruft from experimenting", but if it's important enough to keep in the repo, and important enough for an individual to test, then it's important enough for us to test in our CI.

lib/pymedphys/_experimental/chartchecks/test2.py
SimonBiggs: This is one of @rembishj's files within experimental. For now, I'd like to let him learn to crawl before running :). It can certainly be on his to-do list, but I'm quite keen for it to not be a blocker for this PR.

The only change this PR has done to that file is move the top-level code to be scoped under `main()`.

SimonBiggs:
Hi @sjswerdloff,

This one's quite a hard one to juggle. Given the only thing blocking this PR is something that existed already within main, so following the following "golden rule" I'm going to merge this one so that I can untangle this mess...

> In general, reviewers should favor approving a [PR] once it is in a state where it definitely improves the overall code health of the system being worked on, even if the [PR] isn’t perfect.

lib/pymedphys/_experimental/chartchecks/test2.py
sjswerdloff: Please add an issue and ask @rembishj to assign the issue to himself.

The perfect is the enemy of the good, but accumulated cruft is the death of a software package.  So let's make sure that retiring technical debt is a high priority.

SimonBiggs: Yup good call. Will do. See https://github.com/pymedphys/pymedphys/issues/1381

SimonBiggs: And on that note I've also added another issue:

https://github.com/pymedphys/pymedphys/issues/1382

PR #1377, opened 2021-02-01
[STACK cypress -> base] Make `pymedphys dev cypress` command

SimonBiggs:
It starts up the PyMedPhys GUI server in the background and then boots up the interactive e2e testing suite, cypress.


sjswerdloff:
So does the failing cypress test mean success?

sjswerdloff (approved):

SimonBiggs:
> So does the failing cypress test mean success?

Nope, that's the CI test that was removed in https://github.com/pymedphys/pymedphys/pull/1373

PR #1379, opened 2021-02-01
[STACK angles] Filter calcs by gantry and collimator angles

SimonBiggs:
This was undergone for a range of reasons:

* Following on from this work I will be creating DICOM image files from these specific angles so that those DICOM files can be downloaded and input into independent software.
* When using the new iView version there are 10x the number of images. The calculation time is large, it can be beneficial to have a "quick answer" and then let the server calculate the rest afterwards.
* When setting up automated end to end testing for the application I only want to test a few images otherwise the test time will blow out. This allows me to select a couple for which the e2e tests can be undergone and then have their results compared to baseline.


lib/pymedphys/_experimental/streamlit/apps/wlutz/main.py
```diff
@@ -216,18 +220,129 @@ def main():
...
+
+    tolerance = st.number_input(f"{name} angle tolerance", 0, None, default_tolerance)
+
+    return angles, tolerance
+
+
+def _angle_filtering(database_table):
+    gantry_column, collimator_column = st.beta_columns(2)
+
+    default_gantry_angles = [-180, -135, -90, -45, 0, 45, 90, 135, 180]
```
sjswerdloff, line 246: Useful to identify what coordinate system is in use.
I take it this is some native Elekta coordinate system because DICOM is usually non-negative modulo 360, although a certain vendor introduced a change to DICOM to allow values below zero and above 360 for some reason.

sjswerdloff (approved):
type hints would be nice

SimonBiggs:
> type hints would be nice

Yup 🙂, done. I also did a bit of searching and found https://github.com/predictive-analytics-lab/data-science-types which looks awesome, so I added that too.

Here are the changes since your review:

https://github.com/pymedphys/pymedphys/compare/97a2880020ea232bee9cfc995febb1d7f57d7388...make-option-to-only-calc-cardinals 

> Useful to identify what coordinate system is in use.
I take it this is some native Elekta coordinate system because DICOM is usually non-negative modulo 360, although a certain vendor introduced a change to DICOM to allow values below zero and above 360 for some reason.

Yup, good call. I added a note that angles are defined between -180 and 180 degrees. Strictly speaking the Gantry can over-travel, so it is actually a little outside that range. Also, given the beam position is affected by the 2T and 2R steering look-up tables, and the look-up tables are not the necessarily the same for -180 and +180, it is actually quite important for WLutz to not be utilising a modulo based angle definition.

PR #1380, opened 2021-02-01
App - Sum Coincident DICOM doses

Matthew-Jennings:
Hi guys,

This isn't ready yet, but I thought I'd request some initial feedback before I push too far ahead. Any issues/suggestions? I have a couple of specific questions below.

### DIFF
--- a/lib/pymedphys/.pylintrc
+++ b/lib/pymedphys/.pylintrc
@@ -27,7 +27,7 @@ disable =
     import-outside-toplevel,
     no-else-return,
     unbalanced-tuple-unpacking,
-    raise-missing-from,
+    raise-missing-from
     import-error
 
 
--- a/lib/pymedphys/.pylintrc
+++ b/lib/pymedphys/.pylintrc
@@ -28,7 +28,6 @@ disable =
     no-else-return,
     unbalanced-tuple-unpacking,
     raise-missing-from
-    import-error
 
 
 [FORMAT]
--- a/lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
+++ b/lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
@@ -15,7 +15,7 @@ limitations under the License.
 
 import copy
 import pathlib
-from typing import List, Sequence, Optional
+from typing import List, Optional, Sequence
 
 from pymedphys._imports import numpy as np
 from pymedphys._imports import pydicom


lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,120 @@
...
+from pydicom.errors import InvalidDicomError
+
+from pymedphys._dicom.coords import xyz_axes_from_dataset
+from pymedphys._dicom.dose import dose_from_dataset
+from pymedphys._streamlit import categories
+
+STREAMLIT_STATIC_PATH = pathlib.Path(st.__path__[0]) / "static"
+DOWNLOADS_PATH = STREAMLIT_STATIC_PATH / "downloads"
+if not DOWNLOADS_PATH.is_dir():
+    DOWNLOADS_PATH.mkdir()
```
Matthew-Jennings, line 33: Is this an issue? The standard way to provide a file to the user for download ([link](https://discuss.streamlit.io/t/how-to-download-file-in-streamlit/1806/9)) has a 50 MB write limit.

@sjswerdloff does seem to have a working solution to this in pseudonymise ([here](https://github.com/pymedphys/pymedphys/blob/ecd44331ec2dc06c04680b5ea03d4257b38e9d4c/lib/pymedphys/_streamlit/apps/pseudonymise.py#L84)). I had a bit of trouble adapting this solution for this app, and instead found this convenient fix ([associated link](https://github.com/streamlit/streamlit/issues/400#issuecomment-648580840)).

Happy to learn/adapt to a better fix. This was just nice and easy 😄

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,120 @@
...
+        if not ds.Modality == "RTDOSE":
+            st.error(f"File {fh.name} is not a valid DICOM RT Dose file")
+            return None
+
+        datasets.append(ds)
+
+    return datasets
+
+
+def sum_doses_in_datasets(datasets):
```
Matthew-Jennings, line 75: Expect to move to `dicom` module later

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,120 @@
...
+    ds_summed = copy.deepcopy(datasets[0])
+
+    ds_summed.DoseSummationType = "MULTI_PLAN"
+
+    doses = np.array([dose_from_dataset(ds) for ds in datasets])
+    doses_summed = np.sum(doses, axis=0, dtype=np.float32)
+
+    # ds_summed.DoseGridScaling = np.max(doses_summed) / (2 ^ int(ds_summed.HighBit))
+
+    ds_summed.DoseGridScaling = np.max(doses_summed) / (2 ** 32 - 1)
```
Matthew-Jennings, line 86: Will make more robust later. E.g. setting `ds_summed.BitsAllocated =32` explicitly

SimonBiggs:
> Hi guys,
> 
> This isn't ready yet, but I thought I'd request some feedback before I push too far ahead. Specific questions below:

Yup, I couple of ideas. Only things to implement if you think they'd be worthwhile.

I reckon it'd be quite neat to provide a couple of files within the interface that can be downloaded for demo purposes when someone opens it up within the online app. That also doubles up as a set of files that you can use within the Cypress testing of the App.

It'd probably be helpful to print some info about the DICOM files provided, for example Patient ID and name, users could have it as a quick sanity check that they've uploaded the files they expected.

It might also be neat to have the doses from each DICOM file plotted with a "z-slider". This wouldn't be too crazy to implement. Could just have an `st.slider` above a plot that has subplots for each DICOM file and the summation at the end, the slider can represent the z_index. The matplotlib plot would then just plot that slider z index for each the both (all) DICOM files as well as the summation of that slice.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
SimonBiggs: Yeah, let's not use the `st.__path__[0]` file. Instead something like:

```
pathlib.Path.home() / ".pymedphys" / "downloads"
```

Might work well. At the end of the day, I'd expect this variable to be set within a config file and have some form of sensible default if it isn't in the file.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
SimonBiggs: Yup, I think this is a good way to go about it. Play with it in here, move it into the library once you're happy with it.

Matthew-Jennings:
Thanks @SimonBiggs!

> I reckon it'd be quite neat to provide a couple of files within the interface that can be downloaded for demo purposes when someone opens it up within the online app. That also doubles up as a set of files that you can use within the Cypress testing of the App.

Yep, that does sound good!

> It'd probably be helpful to print some info about the DICOM files provided, for example Patient ID and name, users could have it as a quick sanity check that they've uploaded the files they expected.

Good idea, I agree.

> It might also be neat to have the doses from each DICOM file plotted with a "z-slider". This wouldn't be too crazy to implement. Could just have an `st.slider` above a plot that has subplots for each DICOM file and their summation at the end, the slider can represent the z_index. The matplotlib plot would then just plot that slider z index for each the both (all) DICOM files as well as the summation of that slice.

Yep, that would be pretty cool!

On that last point, is there somewhere useful to store generalisable, streamlit-only functions? E.g. I can imagine many apps benefiting from a nice 3D dose plotting function...

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,120 @@
...
+def _axes_in_datasets_match(datasets):
+
+    # Quick shape (sanity) check
+    if not all(
+        ds.pixel_array.shape == datasets[0].pixel_array.shape for ds in datasets
+    ):
+        return False
+
+    # Full coord check:
+    all_concat_axes = [np.concatenate(xyz_axes_from_dataset(ds)) for ds in datasets]
```
SimonBiggs, line 51: 🙂

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,120 @@
...
+        datasets.append(ds)
+
+    return datasets
+
+
+def sum_doses_in_datasets(datasets):
+
+    ds_summed = copy.deepcopy(datasets[0])
+
+    ds_summed.DoseSummationType = "MULTI_PLAN"
```
SimonBiggs, line 79: I haven't seen this sort of thing before. I imagine this area of things @sjswerdloff might have some nice suggestions.

SimonBiggs:
> On that last point, is there somewhere useful to store generalisable, streamlit-only functions?

At the moment I have been storing these sorts of things within either of the two places:

* https://github.com/pymedphys/pymedphys/tree/main/lib/pymedphys/_experimental/streamlit/utilities
* https://github.com/pymedphys/pymedphys/tree/main/lib/pymedphys/_streamlit/utilities

> E.g. I can imagine many apps benefiting from a nice 3D dose plotting function...

Yup 🙂 me too 🚀 🙂

sjswerdloff (commented):
I will take a closer look in the next 24 to 48 hours.
You will want dose summation to reflect the types of the inputs, and mind whether the doses are PHYSICAL or EFFECTIVE and if any one is EFFECTIVE the output type is EFFECTIVE.

The file i/o will fail if it is directory based.
I can try to generalise the approach I took to make it useful for this.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
SimonBiggs: Actually there is a better fourth option. Make a PR to streamlit that makes it so that there can be a configurable downloads path. (Or something equivalent that meets the need here).

I think I might looking to make that happen, that way, we can configure the downloads path to be within the user's drive, and it can be fully supported upstream (and useful to others with similar issues).

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
sjswerdloff, line 75: type hints plus docstring.
If you use VS Code, there are automated docstring generation extensions

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
sjswerdloff: the choice of "MULTI_PLAN" for DoseSummationType is only valid if each of the input doses have DoseSummationType of PLAN.

If you have other DoseSummationTypes on input, you may need to scale them based on number of fractions specified in the relevant fraction group in the plan (if they are of type XXX_SESSION (BEAM, FRACTION, BRACHY).

For a first pass, I would probably limit input to type PLAN (which implies dose for the whole course of treatment, not just a single fraction).
It depends on your intended use.
Its one thing to want to be able to sum something basic like a primary treatment and a boost, it's a little more complicated when you are summing from different modalities (brachy plus photon, photon and proton boost, proton and carbon boost), but it's a whole different story if you are trying to do weighted optimisation of control points or beams (e.g. re-weighting a VMAT based on ~360 "beams" and varying your optimisation criteria).

sjswerdloff: you at least need to confirm that the doses are not from different patients.
PatientID is probably good enough (all have same PatientID).

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,120 @@
...
+
+    doses = np.array([dose_from_dataset(ds) for ds in datasets])
+    doses_summed = np.sum(doses, axis=0, dtype=np.float32)
```
sjswerdloff, line 82: this is strictly opinion, but I would consider starting off with a weighted sum where all of the weights are 1.0
There will be situations where older doses should be de-weighted a bit...
But it depends on your intended use.

sjswerdloff (changes requested):
Ultimately I would aim for fulfiling the role of an IHE-RO Registered Dose Compositor
http://ihe-ro.org/doku.php?id=profiles:dcomp
So the above might provide some useful reference material.
Great start.

I have to hit the Request Changes based on current lack of confirmation of same patient id.
I strongly urge enforcing inputs all being DoseSummationType == PLAN
the other stuff is my typical griping about lack of type hints and docstring.
In experimental, I can't enforce that.
In the non-experimental area I intend to be iron-fisted about public function docstring and vociferous about type hinting.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
SimonBiggs: > Actually there is a better fourth option. Make a PR to streamlit ...

I have found the best way forward with these sorts of things, is implement it first in PyMedPhys, show the working example, and then merge the important pieces of that upstream.

Matthew-Jennings:
Thank you both, @SimonBiggs & @sjswerdloff for your comments! All very useful!

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
Matthew-Jennings: Both are now done :)

In the case of PLAN dose, I've included a check in `sum_doses_in_datasets()` (which will hopefully move to the `dicom` library module) and `_check_files_valid()` (which is specific to this app). It's nice to include the latter because it'll notify the user of the first file(name) that fails the requirement.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
Matthew-Jennings: I've had a go at this. I'm pretty new to type hints. :smile:

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
Matthew-Jennings: At the moment, the intention is to use this function to sum together two "phases"/"trials"/"beamsets" of the same plan ("plan" as defined in RayStation, not DICOM). Primary treatment + boost being one example.

Note the matching coords requirement: it's very unlikely that doses calculated on a different CT scan (as would almost certainly be the case for adding doses from, say, different modalities) will be added using this function as it currently stands. It's not my intention to provide that functionality, at least not yet.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -121,7 +123,9 @@ def patient_ids_in_datasets_are_equal(datasets: list[pydicom.dataset.Dataset]) -
     return all(ds.PatientId == datasets[0].PatientId for ds in datasets)
 
 
-def sum_doses_in_datasets(datasets: list[pydicom.dataset.Dataset]):
+def sum_doses_in_datasets(
+    datasets: Sequence[pydicom.dataset.Dataset],
+) -> pydicom.dataset.Dataset:
```
SimonBiggs, line 128: Anytime a top level function has a type hint that utilises an optional dependency, can we surround it in quotes? That way the `pydicom` module isn't accessed at import time, and the dependency is still able to be optional 🙂 


```suggestion
) -> "pydicom.dataset.Dataset":
```

Matthew-Jennings: I'm happy to do this :smile:. Does this also apply to parameter type hints?

SimonBiggs: > Does this also apply to parameter type hints?

Yup 👍

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
Matthew-Jennings: Would you mind if I waited until a use case comes up that requires weighted summing before implementing? I don't think we'd use it at my institution..

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
SimonBiggs: As an initial work around so that this first pass can be merged, would you be able to nest this directory creation within a function? At the moment, if this was merged, the entire https://app.pymedphys.com may not work... as the "not allowed to make a directory" error will trigger on import (and therefore as the app boots). However, if you bring it in as a function it'll mean only your app won't work 🙂.

Matthew-Jennings: > As an initial work around so that this first pass can be merged, would you be able to nest this directory creation within a function?

Done!

Matthew-Jennings:
> It might also be neat to have the doses from each DICOM file plotted with a "z-slider". This wouldn't be too crazy to implement. Could just have an st.slider above a plot that has subplots for each DICOM file and their summation at the end, the slider can represent the z_index. The matplotlib plot would then just plot that slider z index for each the both (all) DICOM files as well as the summation of that slice.

I actually have something like this nearly ready-to-go in `plotly`, which [Streamlit seems to support nicely](https://docs.streamlit.io/en/stable/api.html#streamlit.plotly_chart). What are your thoughts on using `plotly` vs. alternatives?

SimonBiggs:
> I actually have something like this nearly ready-to-go, but using `plotly`, which [Streamlit seems to support nicely](https://docs.streamlit.io/en/stable/api.html#streamlit.plotly_chart). What are your thoughts on using `plotly` vs. alternatives?

I've really liked Altair. But if you've already got a plotly version I am more than happy to have plotly be used :).

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,241 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import copy
+import os
+import pathlib
+from typing import Sequence, Union
```
SimonBiggs, line 19:
```suggestion
from typing import Sequence, Union, List
```

SimonBiggs: Grab yourself this following extension in VSCode:

https://marketplace.visualstudio.com/items?itemName=ms-pyright.pyright

...we should add it to the recommended extensions list in the repo...

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,241 @@
...
+
+from pymedphys._imports import streamlit as st
+
+import numpy as np
```
SimonBiggs, line 23:
```suggestion
from pymedphys._imports import numpy as np
```

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,241 @@
...
+
+import pydicom
```
SimonBiggs, line 25:
```suggestion
from pymedphys._imports import pydicom
```

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,241 @@
...
+from pydicom.errors import InvalidDicomError
```
SimonBiggs, line 26:
```suggestion
```

Then, add `import pydicom.errors` over here:

https://github.com/pymedphys/pymedphys/blob/c5fa262fd5eb927dcbbe1fced3469f13ad7f1a96/lib/pymedphys/_imports/imports.py#L75

And when using `InvalidDicomError` instead call it as `pydicom.errors.InvalidDicomError`

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
Matthew-Jennings: Thanks! Will do :)

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
Matthew-Jennings: Thanks!

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
Matthew-Jennings: @SimonBiggs, there seems to be a small discrepancy between `pyright` warnings in VS Code and the CI suite. On my local VS Code, I get and error showing `List` is not defined (even though it's in a string version of a type hint!):

image

But the CI suite complains that List is unused:

image

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,319 @@
...
+        If False, the pretty patient name will look like
+        '[Hon]. FirstName [MiddleName] LastName'.
+        Defaults to False.
+
+    capitalise_surname : bool
+        If True, the pretty patient name will have the surname
+        capitalised. If False, surname will be in title case. Defaults
+        to True.
+
+    surname_first : bool
```
SimonBiggs, line 147:
```suggestion
    include_honorific: bool
```

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,319 @@
...
+    ):
+        return False
+
+    # Full coord check:
+    all_concat_axes = [np.concatenate(xyz_axes_from_dataset(ds)) for ds in datasets]
+
+    return all(np.allclose(a, all_concat_axes[0]) for a in all_concat_axes)
+
+
+def get_pretty_patient_name_from_dicom_dataset(
```
Matthew-Jennings, line 121: @SimonBiggs, @sjswerdloff - a handy addition to the library? Does one already exist?

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,319 @@
...
+            accept_multiple_files=True,
+        )
+
+    if st.button("Sum Doses"):
+
+        datasets = _check_files_valid(files)
+
+        if datasets:
+
+            with right_column:
```
Matthew-Jennings, line 298: @SimonBiggs, stolen from `dicom explorer` 😄

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
SimonBiggs: Yup, that's the first thing I thought when I saw that. "Straight to the pool room" for that one 🙂

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
Matthew-Jennings: > I have found the best way forward with these sorts of things, is implement it first in PyMedPhys, show the working example, and then merge the important pieces of that upstream.

I'm not sure I have the know-how to do this alone, but very keen to help/learn?

SimonBiggs (approved):
I'm going to leave an approval here to mark that all things where I have expertise I am happy with 🙂. (Once all of the CI passes)

In my opinion it goes above and beyond the "experimental" domain it's in. However, where feasible, all "experimental" code should aim for that anyway 🙂. 

For it to enter into "library proper" and be able to be bumped up to "BETA" category I think the following steps would need to occur:

* any of @sjswerdloff concerns would need to be addressed,
* the download approach would need to be adjusted (I've taken on that and begun it in https://github.com/pymedphys/pymedphys/pull/1388),
* and lastly a set of end to end tests that flex the GUI with some baseline data and runs a comparison.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
SimonBiggs: Ahh, see, within a string, it still needs to be imported 🙂. ...but pylint doesn't know about its usage when it is within a string. In that case above, you can either ignore the pylint warning, or only put the quotes around Dataset:

```python
Union[List["pydicom.dataset.Dataset"]]
```

Personally I'd opt for not disabling pylint if it is avoidable, but in this case, happy for you to go either way.

SimonBiggs: Also, if you want to run pylint locally:

```
poetry run pymedphys dev lint
```

lib/pymedphys/.pylintrc
```diff
@@ -27,7 +27,8 @@ disable =
     import-outside-toplevel,
     no-else-return,
     unbalanced-tuple-unpacking,
-    raise-missing-from
+    raise-missing-from,
+    import-error
```
SimonBiggs, line 31: Let's not ignore that one here.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
Matthew-Jennings: Thanks! I'll go with the first suggestion.

lib/pymedphys/.pylintrc
Matthew-Jennings: Yeah, I'll revert

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,318 @@
...
+
+CATEGORY = categories.PRE_ALPHA
+TITLE = "Sum Coincident DICOM Doses"
+
+HERE = pathlib.Path(__file__).parent.resolve()
+
+
+def _check_files_valid(
+    files: Sequence[pathlib.Path],
+) -> Union["List[pydicom.dataset.Dataset]", None]:
```
SimonBiggs, line 36:
```suggestion
) -> Optional[List["pydicom.dataset.Dataset"]]:
```

See https://docs.python.org/3/library/typing.html#typing.Optional

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,318 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import copy
+import pathlib
+from typing import List, Sequence, Union
```
SimonBiggs, line 18:
```suggestion
from typing import List, Sequence, Union, Optional
```

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -0,0 +1,318 @@
...
+            "used as a template for the summed DICOM RT Dose file.",
+            ["dcm"],
+            accept_multiple_files=True,
+        )
+
+    if st.button("Sum Doses"):
+
+        datasets = _check_files_valid(files)
+
+        if datasets:
```
SimonBiggs, line 295:
```suggestion
        if not datasets:
	        st.stop()
```

And then all of the below need not be nested. The app will stop at this line if no datasets are uploaded.

Maybe instead have:

```python
if not files:
    st.stop()
```

Right below the `st.file_uploader` line

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -15,7 +15,7 @@
 
 import copy
 import pathlib
-from typing import List, Sequence, Optional
+from typing import List, Optional, Sequence
```
SimonBiggs, line 18: Good little @pymedphys-bot 🙂

Matthew-Jennings: 😆

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
Matthew-Jennings: Yeah, I wondered about the best way to do all of this!

If I go for `if not files`, I'll still either need to keep `if not datasets:` or have `_check_files_valid()` raise errors rather than returning `None`. I initially wanted to avoid `_check_files_valid()` raising errors to avoid the user having to re-upload all files, since I thought that Streamlit would crash when such an error is thrown. Now I've learned that Streamlit has lovely inbuilt error handling, so I can just throw the errors after all...

Matthew-Jennings: So, pyright is hyper picky about this. Unnesting gives this:

image

So, I can either re-nest, or "optionalise" the type hint for `sum_doses_in_datasets()`.

Nesting seems cleaner...

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
SimonBiggs: > are you talking something as simple as a path in a config file?

Actually, after some more thought my idea is to make an `st_download(filename, path)` function to which a file path can be passed. Then the function will run off and make a specific URL point to that file directly, no matter where it is on the file system. The specific URL would have the session ID within it so as to have user's downloads not clash with each other, and a filename.

Began the work in the following PR:

https://github.com/pymedphys/pymedphys/pull/1388

Next part is to set up a thread-safe Queue (https://docs.python.org/3/library/queue.html), and then have a URL `http://localhost:8051/downloads/{session_id}/{filename}` that when called it checks to see if that session_id has run the `st_download` function with the defined filename. If it has, it points the URL to the filepath provided.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
SimonBiggs: Would you be okay if I did a bit of refactoring?

Matthew-Jennings: More than happy!

SimonBiggs: :slightly_smiling_face:

sjswerdloff (approved):

sjswerdloff:
LGTM

lib/pymedphys/.pylintrc
SimonBiggs, line 31:
```suggestion
```

lib/pymedphys/.pylintrc
```diff
@@ -27,7 +27,8 @@ disable =
     import-outside-toplevel,
     no-else-return,
     unbalanced-tuple-unpacking,
-    raise-missing-from
+    raise-missing-from,
```
SimonBiggs, line 30:
```suggestion
    raise-missing-from
```

SimonBiggs:
@Matthew-Jennings I reckon, merge this one in, before it grows any further. Then let https://github.com/pymedphys/pymedphys/pull/1395 stand on it's own as its own PR.

Matthew-Jennings:
> @Matthew-Jennings I reckon, merge this one in, before it grows any further. Then let https://github.com/pymedphys/pymedphys/pull/1395 stand on it's own as its own PR. Especially given it's had its review rounds.

Thanks very much,  @SimonBiggs and @sjswerdloff ! I'll review #1395 today

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
SimonBiggs: Just a heads up. This was solved within PyMedPhys in https://github.com/pymedphys/pymedphys/pull/1402. Looking to merge upstream. Will now track within the following Streamlit issue:

https://github.com/streamlit/streamlit/issues/2722

Matthew-Jennings: Yay! Should I wait and see if Streamlit merge in first?

SimonBiggs: Nah, use our version. We can change it later. Currently, in your current implementation, if anyone installs PyMedPhys with a fresh install of Ubuntu 20.04 with the system version of Python, PyMedPhys and Streamlit will be installed with limited user file access rights. And if they click on your App, and have that function run, the Streamlit App will crash.

... Given PyMedPhys is a library that supports installation on a range of systems, leaving the download option as is, how it has been set up here isn't feasible.

And the version I have set up here should work just fine, even if it isn't merged upstream.

PR #1383, opened 2021-02-01
Sort out testing

rembishj:
Looks like I might've caused some confusion with the names of those files. I guess scratch-files would've been a more accurate description than "test-files". Regardless, they weren't necessary anymore and I should've deleted them earlier.


SimonBiggs:
> Looks like I might've caused some confusion with the names of those files. I guess scratch-files would've been a more accurate description than "test-files". Regardless, they weren't necessary anymore and I should've deleted them earlier.

All good :slightly_smiling_face:, these are in experimental, so you were in the clear. It is helpful to push the code further and further out of its experimental state, which has been happening anyway :slightly_smiling_face:.

SimonBiggs (approved):

PR #1386, opened 2021-02-02
Add dicom header explorer

SimonBiggs:


sjswerdloff (approved):
Basically a no depth dump?
Or does __repr__() recurse ?

SimonBiggs:
> Or does **repr**() recurse ?

repr does recurse 🙂

sjswerdloff:
I'm curious what the performance will be on a large RT Structure Set.
I've crashed/hung other python based tools that try to display everything all at once.
That shouldn't prevent someone from making use of this for other things, but I've taken to just doing an indented recursive text dump and then opening it up in a text editor when I'm not able to use either the DVTK DCMEditor (not sure if that's the proper name), or my (commercial) work environment DICOM Editor.  Maybe dicompyler has a good review tool, but I don't think it's an editor.

SimonBiggs:
> I'm curious what the performance will be on a large RT Structure Set.

You made me curious too so I uploaded the following to the tool:

[RS.1.2.840.10008.5.1.4.1.1.481.3.1592368893_Anonymised.zip](https://github.com/pymedphys/pymedphys/files/5909328/RS.1.2.840.10008.5.1.4.1.1.481.3.1592368893_Anonymised.zip)

It took a good 10 seconds to display, but... the key to not crashing the browser I believe was the fact that it wasn't actually recursing into the Contour Data tags:

```
      ---------
      (3006, 0016)  Contour Image Sequence  1 item(s) ---- 
         (0008, 1150) Referenced SOP Class UID            UI: CT Image Storage
         (0008, 1155) Referenced SOP Instance UID         UI: 1.2.840.113704.1.111.3880.1554938450.278989
         ---------
      (3006, 0042) Contour Geometric Type              CS: 'CLOSED_PLANAR'
      (3006, 0046) Number of Contour Points            IS: "51"
      (3006, 0050) Contour Data                        DS: Array of 153 elements
```

Instead, it just displayed that the contour data was an array and gave its size. So... for structure data, it is limited currently. It would be quite nice to expand on this tool though in the future. I think I will do it on an as needs basis, whenever there's something I want to explore... I'll just add it to the explorer then. (Others welcome to dive in too 🙂)

lib/pymedphys/_experimental/streamlit/apps/dicom_explorer.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+
+    st.markdown(
+        textwrap.dedent(
+            """\
+            ```
+            {}
+            ```
+            """
+        ).format(dicom_header.__repr__())
```
Matthew-Jennings, line 61: @SimonBiggs. Just for my understanding, is `__repr()__` needed here? Did you just include it for explicitness?

Matthew-Jennings:
@SimonBiggs, I just wanted to let you know that I did have a look at this but couldn't test properly until today (and @sjswerdloff is fast!). Looks nice :)

lib/pymedphys/_experimental/streamlit/apps/dicom_explorer.py
SimonBiggs: Without calling `__repr__` internally `__str__` would have been called by default.

I probably should have called it as `repr(dicom_header)` instead of using the dunder call.

...In the past on pydicom I thought `str()` and `repr()` gave a different response. But it seems now that they're essentially the same.

Some objects do have different `repr` and `str`, and strictly `repr` as opposed to string is what is wanted here. When they are different it's because `repr` is designed to be a human friendly representation of the object, and `str` is supposed to be "what the value of the object would be if it was converted to string".

Given here, I want a "representation of the object that makes sense for humans" `repr` is the way to go here.

Here is an example of the difference with an object we use a lot:

```ipython
C:\Users\sbiggs\git\pymedphys [dicom-viewer ≡ +0 ~1 -0 !]> ipython
Python 3.7.9 (tags/v3.7.9:13c94747c7, Aug 17 2020, 18:58:18) [MSC v.1900 64 bit (AMD64)]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.19.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np

In [2]: x = np.array([1, 2, 3])

In [3]: str(x)
Out[3]: '[1 2 3]'

In [4]: repr(x)
Out[4]: 'array([1, 2, 3])'
```

Matthew-Jennings: Is it possible that you have `str` and `repr` the wrong way around? My reading of the difference between `str` and `repr` is along the lines of:

- `repr` is meant to produce an"official" string representation of an object, and if possible, should look like a valid Python expression that could be used to recreate an object with the same value. It's goal is to be unambiguous and major use case is for debugging
- `str` is meant to produce a readable representation of the object. Its goal is readability and major use case is printing.

These SO links seems to corroborate this:
- https://stackoverflow.com/a/2626364/4665458
- https://stackoverflow.com/a/3691117/4665458
- https://stackoverflow.com/a/18393727/4665458

Seeing as the purpose of this app seems to be to produce a human readable version of the DICOM dataset, it seems to me that `str` might actually be more appropriate than `repr`, at least idiomatically.

Importantly, I don't really mind at all. Just learning! :smile:

SimonBiggs: Ahh, you're right! I do have it the wrong way round!

PR #1387, opened 2021-02-02
[STACK base] First pass on iView to DICOM (before big refactor)

SimonBiggs:
It's nice when goals align 🙂. For this PR I need to be able to download a large zip file full of DICOM files... ... so 🙂, I'm going to pause development on this branch for a little to solve https://github.com/pymedphys/pymedphys/issues/1399 first.

---

Make the DICOM conversion tool be its own Streamlit app.


SimonBiggs:
> Create DICOM conformant image from iView data
> To achieve conformance, you probably want to create a nominal RT Plan from the data used for treatment or the original DICOM RT Plan that came from the TPS, identify which Beam the image corresponds to, and then populate from that information.
> e.g. _create_DICOM_image_from_plan_and_iview_data(plan, referenced_beam_number, iview_data)

I love the idea, unfortunately that won't fit here. Many of the times the images here would be service mode beams that have been delivered in "free-form". So open the field, deliver a beam, and record it on the iView. No RT plan in sight... :/

SimonBiggs:
Hi @sjswerdloff,

Given you'd already had a bit of a read of this one I didn't want you to have to double up your review effort. ...and... the next bunch of changes is me moving a whole lot of things. So I didn't want these changes here to get left by the wayside.

The DICOM work here is most certainly partial. My plan is to first refactor all of the iView/iCOM syncing and filtering code out of the WLutz Arc app, into its own stand-alone utility. Then, with that utility I plan to tackle this DICOM creation code. If DICOM focused changes could happen after that refactor as opposed to in this initial pass that would be really helpful.

sjswerdloff:
> > Create DICOM conformant image from iView data
> > To achieve conformance, you probably want to create a nominal RT Plan from the data used for treatment or the original DICOM RT Plan that came from the TPS, identify which Beam the image corresponds to, and then populate from that information.
> > e.g. _create_DICOM_image_from_plan_and_iview_data(plan, referenced_beam_number, iview_data)
> 
> I love the idea, unfortunately that won't fit here. Many of the times the images here would be service mode beams that have been delivered in "free-form". So open the field, deliver a beam, and record it on the iView. No RT plan in sight... :/

Then prepopulate with hardcoded or unique generated, service related values for type 1 elements.  Patient Name, Patient ID,  study, series, instance UID, etc.  At least make it so that the data can be organised and will not get mixed up or misinterpreted.

SimonBiggs:
> Then prepopulate with hardcoded or unique generated, service related values for type 1 elements. Patient Name, Patient ID, study, series, instance UID, etc. At least make it so that the data can be organised and will not get mixed up or misinterpreted.

Just wanting to confirm that this comment has been addressed by https://github.com/pymedphys/pymedphys/pull/1413#discussion_r571819624

sjswerdloff (approved):

PR #1388, opened 2021-02-02
Make Hello World server handler

SimonBiggs:
This makes a new page at `http://localhost:8501/pymedphys` that simply says "Hello World!". I see two uses for this at the moment:

* Create a `http://localhost:8501/docs` that hosts the PyMedPhys docs for the installed version of PyMedPhys
* Create a `http://localhost:8501/downloads/{session_id}/{random_uid}/{filename}` download link that can point to arbitrary files on the filesystem.

Given the `http://localhost:8501/pymedphys` URL isn't harming anyone I figure this is an okay state for a merge.

### DIFF
--- a/lib/pymedphys/_gui.py
+++ b/lib/pymedphys/_gui.py
@@ -53,14 +53,14 @@ def _monkey_patch_streamlit_server():
         app = official_create_app(self)
         base = st.config.get_option("server.baseUrlPath")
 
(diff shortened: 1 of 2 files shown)


SimonBiggs:
Consider scrapping the random UID for now.

sjswerdloff (approved):

PR #1394, opened 2021-02-03
Refactor of `sum_doses.py`

SimonBiggs:
Importantly... when I click to download the created DICOM file... it doesn't work... But, that's okay. Experimental apps don't have to work yet :slightly_smiling_face:. Probably best to get this merged as is as long as it doesn't break the build... Then do further work on it in another PR.

---

Also, I should say, I key goal here was to make it so that no function returns had to use `Optional`, which intern allowed for reduced complexity of the `main()` function.

Also, I chose to make the user feedback be more progressive. One file immediately gives the name and ID and raises any errors without needing to press a button, for example.


lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -272,47 +337,3 @@ def sum_doses_in_datasets(
...
-        st.write("## Upload DICOM RT Dose files")
-        files = st.file_uploader(
-            "Upload at least two DICOM RT Dose files whose doses you'd "
-            "like to add together. The first file uploaded will be "
-            "used as a template for the summed DICOM RT Dose file.",
-            ["dcm"],
-            accept_multiple_files=True,
-        )
-
-    if st.button("Sum Doses"):
```
Matthew-Jennings, line 291: I'd was aiming to provide the ability to sum _more_ than two doses. By having a button, summing won't triggered (and unused files won't be generated) until the user is ready/has uploaded all files they want to sum.

I don't think there's a way to achieve this without a button..

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -28,56 +28,121 @@
...
+
+def _validate_dicom_dataset(ds):
+    if ds.Modality != "RTDOSE":
+        raise ValueError("File is not a valid DICOM RT Dose file")
+
+    if ds.DoseSummationType != "PLAN":
+        raise ValueError("File is not a 'plan' dose")
+
+
+def _validate_comparison_to_initial(ds, ds0):
```
Matthew-Jennings, line 132: I've just worked out we need to require that `ds.DoseUnits = 'GY'`. We can't really add relative doses easily - they'll have different norm points. Or at least the current implementation would fail to do so correctly.

With that in mind, this function isn't necessary and a simple `ds.PatientId == ds0.PatientId` would suffice.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -28,56 +28,121 @@
...
+
+
+def _validate_dicom_dataset(ds):
+    if ds.Modality != "RTDOSE":
+        raise ValueError("File is not a valid DICOM RT Dose file")
+
+    if ds.DoseSummationType != "PLAN":
+        raise ValueError("File is not a 'plan' dose")
+
+
```
Matthew-Jennings, line 131:
```suggestion
if ds.DoseUnits != "GY":
    raise ValueError("File must contain absolute dose")
    
    
```

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -28,56 +28,121 @@
 CATEGORY = categories.PRE_ALPHA
 TITLE = "Sum Coincident DICOM Doses"
 
-HERE = pathlib.Path(__file__).parent.resolve()
 
+def main():
```
Matthew-Jennings, line 32: Yeah, `main()` is nicer at the top

Matthew-Jennings:
> Importantly... when I click to download the created DICOM file... it doesn't work... But, that's okay.

😢 works for me!

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
SimonBiggs: All good, a button's fine 🙂

SimonBiggs:
Would you be happy just to make your adjustments to your preferences in this branch? I'm happy for you to go town.

Matthew-Jennings (approved):

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -51,34 +51,36 @@ def main():
         st.write(e)
         st.stop()
 
-    if not datasets:
-        st.stop()
+    if st.button("Click to Sum Doses"):
```
Matthew-Jennings, line 54: Add button back in

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -51,34 +51,36 @@ def main():
...
-        )
+                * Patient ID: `{datasets[0].PatientID}`
+                * Patient Name: `{get_pretty_patient_name_from_dicom_dataset(datasets[0])}`
+                """
+            )
 
-    if len(datasets) < 2:
-        st.stop()
+        if len(datasets) < 2:
+            # Note that if the user tries to remove a file that has
```
Matthew-Jennings, line 68: This note is something to be aware of. It's actually quite annoying...

Haven't tried working around it yet

Matthew-Jennings:
I'll let you merge if/when you're happy, @SimonBiggs 🙂

SimonBiggs (commented):
I can't approve my own PR, but, LGTM 🙂

PR #1396, opened 2021-02-03
Reduce merge conflicts in `pyproject.toml`

SimonBiggs:
Make the auto-generated extras section split into new lines and be sorted. That way, should a dependency be changed on two separate branches this should now merge with minimal to no conflict.


Matthew-Jennings:
So it occurs to me as I try to review this sort of thing: what would you be specifically looking for from a review?

E.g., a quick glance over the code gives me a little confidence. Reading an explanation of the changes and how you've verified the safely achieve their intention gives me more. Testing myself gives the most. But these require increasing time and understanding.

SimonBiggs:
> So it occurs to me as I try to review this sort of thing: what would you be specifically looking for from a review?
> 
> E.g., a quick glance over the code gives me a little confidence. Reading an explanation of the changes and how you've verified the safely achieve their intention gives me more. Testing myself gives the most. But these require increasing time and understanding.

Will get back to this later, but essentially, my answer will be heavily based (or copied from) on https://github.com/google/eng-practices/blob/master/review/reviewer/index.md

SimonBiggs:
> what would you be specifically looking for from a review?

I think a key point is the following:

https://github.com/google/eng-practices/blob/master/review/reviewer/looking-for.md#every-line-every_line

> Look at every line of code that you have been assigned to review. Some things like data files, generated code, or large data structures you can scan over sometimes, but don't scan over a human-written class, function, or block of code and assume that what's inside of it is okay. Obviously some code deserves more careful scrutiny than other code—that's a judgment call that you have to make—but you should at least be sure that you understand what all the code is doing.
>
> **If it's too hard for you to read the code and this is slowing down the review, then you should let the developer know that and wait for them to clarify it before you try to review it. ... If you can't understand the code, it's very likely that other developers won't either. So you're also helping future developers understand this code, when you ask the developer to clarify it.**

As in, if understanding is taking time opt to request for clarification/explanation. Opt to do that over spending a decent amount of time trying to understand. This has multiple benefits, our discussion with each other here can be read by others and many can learn, explaining the code can help you undergo a more quality code review, it can highlight areas of improvement etc.

So, if you look at every line of code, understand it, and it meets the [requirements for a code review](https://github.com/google/eng-practices/blob/master/review/reviewer/looking-for.md#summary), then it's good to go. If it will take up too much of your time trying to understand it, ask for clarification. The goal is to treat reviewers time as a precious resource. Far better to ask the code writer to explain themselves than struggle about trying to see what's going on.

lib/pymedphys/_dev/propagate.py
```diff
@@ -303,6 +303,17 @@ def propagate_extras():
...
+        extras[group] = sorted(deps)
+
+    extras = tomlkit.item(
+        extras, _parent=pyproject_contents["tool"]["poetry"], _sort_keys=True
+    )
+
+    for _, deps in extras.items():
+        if len(deps.as_string()) > 88:
+            deps.multiline(True)
+
```
SimonBiggs, line 316: As a note, the only part of this PR is human written is the above lines. Everything else is auto-generated by `pymedphys dev propagate`

SimonBiggs:
I've been having a bit of re-read of those engineering practice documents. Each time I read them I get something more out of them. They're very valuable.

They recommend that instead of just simply asking the code writer to explain themselves on GitHub, to instead ask if they can either make a certain part of the code easier to understand, or explain the choices within the code itself. See the following:

https://github.com/google/eng-practices/blob/master/review/reviewer/comments.md#accepting-explanations-explanations

Matthew-Jennings:
I see. I've got a bit of catching up to do! To be honest, I don't even really know what `propogate` _does_ yet. Not enough to feel confident to review changes to it. Maybe we can have a TC and go throw some of this kind of thing? I find conversations a very efficient way to achieve understanding.

SimonBiggs:
> Maybe we can have a TC and go throw some of this kind of thing?

Certainly, more than happy to :slightly_smiling_face:. We should take notes and upload them somewhere to build out the documentation though, try to hit two birds with one stone.

sjswerdloff (approved):

SimonBiggs:
> So it occurs to me as I try to review this sort of thing: what would you be specifically looking for from a review?

I should clarify, that should the review be regarding experimental or prototype only code you are free to skim and approve.

PR #1397, opened 2021-02-03
Make lock conflicts revert to main initially

SimonBiggs:
The lock file will then be updated for any new dependencies with propagate. The only downside of this approach that I can see is that when trying to run "poetry update". If that causes conflicts with the current main branch it'll revert the update.

However, those who are doing upgrades should instead first merge main, make sure there are no conflicts, and then run the update. Given only maintainers really should be running upgrades this seems like a fine compromise.

### DIFF
--- a/.github/workflows/conflicts.yml
+++ b/.github/workflows/conflicts.yml
@@ -42,9 +42,7 @@ jobs:
       - name: Install PyMedPhys with propagate environment
         if: steps.conflicts.outputs.status == 'true'
         run: |
-          poetry install -E propagate
-          pip install pipx
-          pipx install poetry-merge-lock
+          poetry install -E propagate || poetry lock --no-updates && poetry install -E propagate
 
       - name: Fix conflicts
         if: steps.conflicts.outputs.status == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/conflicts.yml
```diff
@@ -42,9 +42,7 @@ jobs:
       - name: Install PyMedPhys with propagate environment
         if: steps.conflicts.outputs.status == 'true'
         run: |
-          poetry install -E propagate
-          pip install pipx
-          pipx install poetry-merge-lock
+          poetry install -E propagate || poetry lock --no-updates && poetry install -E propagate
```
sjswerdloff, line 45: What is the order/grouping of operations.?
I tend to put parentheses to make it obvious

sjswerdloff (approved):

.github/workflows/conflicts.yml
SimonBiggs: That would make the commands run in a subshell. (https://unix.stackexchange.com/a/26069). Does that still work here?

I didn't think bash had order of operations... It was my impression that "&&" isn't really "and" and "||" isn't really "or", but instead `||` means, run the following if the previous failed, and `&&` means, run the following if the previous passed, and `;` means run the following regardless...

SimonBiggs:
Normally I'd wait before getting answers back before merging, but I was quite interested in putting it through its paces given that https://github.com/pymedphys/pymedphys/pull/1395 has a merge conflict :slightly_smiling_face:

PR #1403, opened 2021-02-04
Move sum doses to the library

Matthew-Jennings:
So I figured it *might* be worth that extra push to move functions to the library? :smile:

App can come later.


lib/pymedphys/_dicom/header.py
```diff
@@ -147,3 +148,96 @@ def adjust_RED_by_structure_name_cli(args):
...
+
+    Returns
+    -------
+    str
+        A pretty version of PatientName.
+    """
+
+    names_as_str = str(ds.PatientName)
+
+    if include_honorific and "^^" not in names_as_str:
```
Matthew-Jennings, line 191: Small addition

lib/pymedphys/_dicom/dose.py
```diff
@@ -314,3 +318,72 @@ def require_patient_orientation(ds, patient_orientation):
...
+            "'EFFECTIVE' are supported"
+        )
+    if any(ds.DoseType == "EFFECTIVE" for ds in datasets):
+        ds_summed.DoseType = "EFFECTIVE"
+    else:
+        ds_summed.DoseType = "PHYSICAL"
+    doses = np.array([dose_from_dataset(ds) for ds in datasets])
+    doses_summed = np.sum(doses, axis=0, dtype=np.float32)
+
+    ds_summed.DoseGridScaling = np.max(doses_summed) / (2 ** int(ds_summed.HighBit))
```
Matthew-Jennings, line 384: Small change. Was getting saturation issues. Need to divide by max pixel value of 2 ** 31 (in this case).

lib/pymedphys/_dicom/create.py
```diff
@@ -40,13 +40,17 @@ def set_default_transfer_syntax(dataset):
         dataset.is_implicit_VR = True
 
 
-def dicom_dataset_from_dict(input_dict: dict, template_ds=None):
+def dicom_dataset_from_dict(input_dict: dict, ensure_file_meta=False, template_ds=None):
```
Matthew-Jennings, line 43: I was getting complaints of lacking File Meta when trying to write tests with skeleton DICOM datasets generated from `dict`. This fixed it.

lib/pymedphys/_dicom/header.py
```diff
@@ -147,3 +148,96 @@ def adjust_RED_by_structure_name_cli(args):
     new_dicom_dataset = adjust_RED_by_structure_name(dicom_dataset)
 
     pydicom.write_file(args.output_file, new_dicom_dataset)
+
+
+def pretty_patient_name(
```
Matthew-Jennings, line 153: May belong elsewhere? This seemed most logical of the existing files at first glance.

lib/pymedphys/tests/dicom/test_header_tweaks.py
```diff
@@ -280,3 +281,96 @@ def test_structure_name_based_RED_append():
     ]
 
     compare_dicom_cli(command, original_ds, expected_ds)
+
+
+@pytest.mark.pydicom
+def test_pretty_patient_name_all_params():
```
Matthew-Jennings, line 287: There has *got* to be a cleverer way to do this...

lib/pymedphys/_dicom/dose.py
```diff
@@ -314,3 +318,72 @@ def require_patient_orientation(ds, patient_orientation):
...
+
+    Returns
+    -------
+    pydicom.dataset.Dataset
+        A new DICOM RT Dose dataset whose dose is the sum of all doses
+        within `datasets`
+    """
+
+    if not len(datasets) >= 2:
+        raise ValueError("At least two datasets must be provided for comparison")
```
SimonBiggs, line 343: My preference would be for keeping the API aligned to other sum functions in the Python ecosystem where it allows a "sum" of one element:

```python
In [1]: sum([1])
Out[1]: 1

In [2]: import numpy as np

In [3]: np.sum([1])
Out[3]: 1
```

A benefit I can think of, potentially there is a variable amount of plans in a sequence and the total of all the plans is wanted as a return, even if it is just one plan.

---

To the point where I think it'd be worth having a unit test testing this edge case.

lib/pymedphys/_dicom/header.py
SimonBiggs: Maybe a new file within:

https://github.com/pymedphys/pymedphys/tree/main/lib/pymedphys/_dicom/utilities

Something like display? I'm not sure. I'm happy for where you think best :slightly_smiling_face:.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
```diff
@@ -138,190 +138,3 @@ def _save_dataset_to_downloads_dir(ds: "pydicom.dataset.Dataset"):
...
-        raise ValueError(
-            "Only DICOM RT Doses whose DoseTypes are 'PHYSICAL' or "
-            "'EFFECTIVE' are supported"
-        )
-    if any(ds.DoseType == "EFFECTIVE" for ds in datasets):
-        ds_summed.DoseType = "EFFECTIVE"
-    else:
-        ds_summed.DoseType = "PHYSICAL"
-    doses = np.array([dose_from_dataset(ds) for ds in datasets])
-    doses_summed = np.sum(doses, axis=0, dtype=np.float32)
```
SimonBiggs, line 318: So that I can see an in-line diff to what was previously reviewed, would you be able to split this into two PRs, one where the only change is this function moving into its new location and then the appropriate imports being fixed, and then the second PR (set to merge into this 'move-only-PR') be the one with the changes. That way we'll be able to review changes instead of having to review it as a whole.

SimonBiggs (commented):
> So I figured it _might_ be worth that extra push to move functions to the library? 
> 
> App can come later.

Awesome :smile:. I did a quick first pass, not an in-depth review though. The key request is for this to be split into two PRs, one where the only things changed is where the code is (but no actual code changed, except for corresponding imports). The second be a PR that merges into the first where the code is actually being changed. That way code that has already been reviewed doesn't need extra work :slightly_smiling_face:.

lib/pymedphys/_experimental/streamlit/apps/sum_doses.py
Matthew-Jennings: Sure :)

Matthew-Jennings:
See #1404 and #1405

PR #1410, opened 2021-02-05
Initial go at dose plotter.

Matthew-Jennings:


Matthew-Jennings:
Runs into similar data size limit as file downloading when using plotly to add all Heatmaps at once. May still be able to use st slider and build 2D heatmaps one at a time.

image

Matthew-Jennings:
App runs into similar size limit issue:

image

May be able to use st.slider and build plots one at a time to circumvent.

SimonBiggs:
> App runs into similar size limit issue:
> 
> image
> 
> May be able to use st.slider and build plots one at a time to circumvent.

A common pattern is to display coronal, sagittal and transverse slices about a point, and then allow users to click on different locations of the plots and have the other slices update to be at that location? That means not every Dose value needs to be sent to the user, just those of those slices...

SimonBiggs:
If plotly is your poison then the "on click" callback might be what you're after:

https://stackoverflow.com/questions/59487526/plotly-click-callback-for-a-heatmap

Matthew-Jennings:
> A common pattern is to display coronal, sagittal and transverse slices about a point, and then allow users to click on different locations of the plots and have the other slices update to be at that location? That means not every Dose value needs to be sent to the user, just those of those slices...

Thanks! Agree, I'll try for that first. But an eventual goal is to also plot ROIs along with the dose (with dose interpolated to the ROI z coords if needed). In that case, it would be much easier to start with transverse plotting only.

SimonBiggs:
@Matthew-Jennings the reason the diff here looks broken is because main hasn't been merged into the branches it is stacked on top of.

It can be quite painful to have to juggle stacks, so given I was quite happy with the other branches and they've been sitting around for a while without any further comments I'm going to merge them in and make this branch a little easier to see.

lib/pymedphys/_dicom/coords.py
```diff
@@ -150,30 +150,30 @@ def xyz_axes_from_dataset(
     if coord_system.upper() in ("FIXED", "IEC FIXED", "F"):
         x = x_dicom_fixed
         y = z_dicom_fixed
-        z = -np.flip(y_dicom_fixed)
```
Matthew-Jennings, line 153: So I'm embarrassed to say that I'm pretty sure all of these flips were unnecessary in `xyz_axes_from_dataset`. I figured it out whilst exploring plotting and noted some inconvenient relationships between fixed and patient-based coord systems.

I'll try to think up a compelling way to demonstrate that it's working completely correctly now (more for my own peace of mind than anything!). Viewing my new test should hopefully help.

The idea here is that we want, say, `y_axis[0]` to always correspond to `dose[0, i, j]` (IEC) or `dose[k, 0, j]` (DICOM). With the flips, we'll instead get `y_axis[0]` sometimes corresponding with `dose[-1, i, j]` (IEC) depending on the orientation.

- Another way to think about this is that, without flips, the indices of the axes that correspond to the origin (or closest to), x[i0], y[j0] and z[k0] will be associated with dose[`k0`, `i0`, `j0`] (DICOM). This is regardless of orientation and coordinate system. You would still get an axis _swap_ between DICOM and IEC or between S/P and DL/DR, but the index itself is still that of the origin voxel.

Matthew-Jennings: Incidentally, the flips were what necessitated some of the less unintuitive code in `xyz_axes_from_dataset()`. Removing flips provides an opportunity to rewrite to be clearer.

Matthew-Jennings:
> @Matthew-Jennings the reason the diff here looks broken is because main hasn't been merged into the branches it is stacked on top of.
> 
> It can be quite painful to have to juggle stacks, so given I was quite happy with the other branches and they've been sitting around for a while without any further comments I'm going to merge them in and make this branch a little easier to see.

Thanks, Simon!

It occurs to me that I didn't update the Release Notes! Are we tracking apps? Only those in library, or?

lib/pymedphys/tests/dicom/test_coords.py
```diff
@@ -42,54 +42,62 @@ def get_data_file(orientation_key):
...
+def test_axis_extraction_all_orients_and_coord_systems():
     r"""Run the xyz extraction test sequence for a given
     xyz extraction function"""
 
-    expected_xyz_path = download.get_file_within_data_zip(
-        "dicom_dose_test_data.zip", "expected_{}_xyz.json".format(coord_system.lower())
-    )
-
-    with open(expected_xyz_path) as fp:
-        expected_xyz = json.load(fp)
```
Matthew-Jennings, line 54: I'd still like to have a check against a baseline with this (maybe just for one coord system though - all others can be checked relative to this by the new code), but I'd need to resave the `json` and am hazy on where it is/how best to update it!

SimonBiggs:
> It occurs to me that I didn't update the Release Notes! Are we tracking apps? Only those in library, or?

We are trying to track apps also. Any update to the Release Notes would be massively appreciated. I'm going to try and do a release in the next couple of days before I go on leave, so any help in the docs there would be awesome 🙂.

SimonBiggs:
Also, just a heads up, I am going to purposefully avoid reviewing until you flag me. That way, once you've flagged for review you're sort of saying, "I'm okay to only change this with response to review comments now". That way each PR only goes through the review cycle once. Of course, you can achieve this by making a stack, but as you've seen, if they can be avoided that is ideal.

Matthew-Jennings:
I think we're out of luck :( https://discuss.streamlit.io/t/plotly-click-events-in-streamlit/1419

Matthew-Jennings:
> @Matthew-Jennings, also, just a heads up, I am going to purposefully avoid reviewing until you flag me. That way, once you've flagged for review you're sort of saying, "I'm okay to only change this with response to review comments now". That way each PR only initialises the review cycle once. Of course, you can achieve this by making a stack, but as you've seen, if they can be avoided that is ideal.

Understood!

SimonBiggs:
> I think we're out of luck :( [discuss.streamlit.io/t/plotly-click-events-in-streamlit/1419](https://discuss.streamlit.io/t/plotly-click-events-in-streamlit/1419)

Ahh supported, sshmorted, we can still do it 🙂. It's Python, and we're writing tests, we can do what we want 🙂.  

Within the callback save your data to a "session_state object" (see `index.py` for how to use that https://github.com/pymedphys/pymedphys/blob/f4a470e1a64474d3299753d61414b0aebca59efa/lib/pymedphys/_streamlit/index.py#L122), and then once the callback is complete, and all your data is saved in a session_state object, then call `st.experimental_rerun`. Then, just access the results further down in your script by calling that same session_state attribute.

Matthew-Jennings:
@SimonBiggs, in a discussion we had a while ago, we thought it'd probably be best for our dicom axis/coord extraction functions to default to IEC FIXED instead of DICOM as a coordinate system. We reasoned that FIXED is the most useful for plotting.

I thought that while I was updating coord stuff I'd make that change, and that last test is failing because of an old baseline set with the old default. There are a number of ways to address this (e.g. replace DicomDose.coords with some a coord system-specific solution), but the simplest for now would be to replace the baseline. How would I go about that with the current data download mechanism in place?

SimonBiggs:
> There are a number of ways to address this (e.g. replace DicomDose.coords with some a coord system-specific solution), but the simplest for now would be to replace the baseline. How would I go about that with the current data download mechanism in place?

Without being able to grok this PR at the moment my advice is going to be quite limited. Does the blame say who put the baseline in place? If the baseline was put in place by yourself, then I think you can make the call whether or not a baseline is the best way to go forward.

A huge benefit of baselines is that it is a bit of a "catch all". But that very benefit, is also its shortcoming. When a baseline is to be adjusted, it's hard to know "what it was catching". If there are other catches in place as well as baselines then that is beautiful. You get the best of both worlds. The functionality which needed to be tested individually is clear and transparent, and it's very clear what sort of "specification" it is enforcing on the code. With a baseline, it helps, because it is often real world data. It picks up the minutiae of reality and covers things that one may have forgotten to test. But, if we were leaning on the baseline to enforce a specification of how the code should react when it is utilised... if our baselines became our specification... then we need to be mighty careful when updating a baseline after changing code.

I wanted to say, that I absolutely love the following test:

https://github.com/pymedphys/pymedphys/blob/8a99a916ec900f599488b42c7b23a41bbb5796b1/lib/pymedphys/tests/dicom/test_dose.py#L96-L125

It is beautiful. 🙂.

So... if that baseline code covers some piece of functionality in a function that is exposed by the public API... and that's now changed... then that should probably be a PR all on its own, and we should try and get a handle on what code is utilising that function. If users of PyMedPhys have placed code within prototyping or site-specific and it is utilising that public function that is changing, it would probably be worth checking that this change hasn't broken assumptions they were making about the functionality of that function. If it does break those assumptions, then we need to step back and work out if it's worth it. At this stage of PyMedPhys' development, I'm for refining our API, but... breaking changes should stand on their own and be discussed for what they are. ... and should a test need to be changed to no longer break because code has changed, that is a serious flag that potentially we are talking about a breaking change.

---

But... all this was said by barely reading the code in this PR. This PR has certainly reached both "large" status and hit key components of the primary library, where it looks like it will take longer than 15 mins for me to individually read each line and attempt to understand what it is doing and the impacts that will have on the rest of PyMedPhys and beyond.

... in saying that ... since it is still draft and not flagged for review, I am also trying to *not* read the code, given it still may change significantly before a review is actually requested.

Matthew-Jennings:
@SimonBiggs and @sjswerdloff 

Okay, so I must admit that I've kinda gotten lost in the weeds diving into coords stuff here. I need to stop, step back and discuss with you and Stuart the way forward.

In summary, as I dove into dose plotting, it became apparent to me that our handling of coords and "axes" really is incomplete and quite confusing unless both coordinates and "axes" indicate their *coordinate system* and *patient orientation*. I really think this is super important for maintaining the correct relationship with a corresponding dosegrid extracted from DICOM.

I've been experimenting with namedtuples that indicate both coordinate system and patient orientation as can be seen here: https://github.com/pymedphys/pymedphys/pull/1410/files#diff-77b25ab342f25c20f8c450f0dd6053da0004b2b6cee4ed4c3a374bc740079704R23-R24

But I've gotten far too ahead of myself and I think we need to reach some sort of agreement before I proceed.

Matthew-Jennings:
(Very happy to split into multiple PRs, and not really asking you to review this entirely, but wanting to gauge your thoughts on Axes/Coords tuples at the outset)

sjswerdloff:
I have very strongly held opinions on coordinate systems, however those are focused on patient treatment and not on physics QA.

I will take a closer look at the coordinate system and axes issues here.
I will also do a deep dive in the code.
But probably not until the weekend.

Matthew-Jennings:
@sjswerdloff, on further thought and investigation, please don't spend any time reviewing this monster.

I believe that we can come up with a much neater solution for handling dose and coordinates. In particular, as @SimonBiggs once suggested, I really think [`xarray`](http://xarray.pydata.org/en/stable/index.html) can provide a beautiful solution. I've been prototyping the use of `xarray` for dose/coord manipulation in another branch. I'll open up a PR for that branch and close this behemoth.

Just note that, at present, we have semi-broken coords and axis extraction functions, along with dubious nomenclature. I quite like this stuff and I really think PyMedPhys could strongly benefit from having a nice solution to this. Indeed, this was one of the first things I worked on here!

PR #1411, opened 2021-02-05
Change `download` to accept file content instead of file path

SimonBiggs:
@Matthew-Jennings she's good to go :slightly_smiling_face:. Here is a newly pressed `download()` function that can also accept `BinaryIO` :slightly_smiling_face:. 

---

[EDIT]The below comments are now out of date, see https://github.com/pymedphys/pymedphys/pull/1411#issuecomment-774367070 [/EDIT]

I'm not so convinced with the API. I feel like people are going to want to write something like:

```python
download("a_file.txt", "Some awesome content")
```

Except that there won't work in it's current form. They would have to do:

```python
import io
download("a_file.txt", io.BinaryIO("Some awesome contents".encode()))
```

...quite convoluted.

So ideally, the function can be passed a string. ...but a string is currently assumed to be a path.

...maybe... the best approach actually would be to never let it be a path. But always be file contents. That way, if a path is actually wanted then the user would need to write:

```python
with open('path/to/file') as f:
    download(filename, f.read())
```

People are used to that pattern... it also has the added benefit of "failing fast", meaning if the filepath provided to `download` doesn't exist, it fails when `download` is called. But also, now the following would work:

```python
import tempfile
import pathlib

with tempfile.TemporaryDirectory() as dir:
	path = pathlib.Path(dir, 'a_file')
	... do stuff
    download(filename, path)

... more things
```

In the current design of `download` if the above was used as an approach even if the `download` code had some sort of sanity "check if file exists" code, it would still think everything was okay... until the user goes to click the link the file would no longer be there and it would error out.

Instead, by requiring the content of the file to always be passed to `download` even if the file no longer exists on disk when the user goes to click on it, still the user will get the version of the file as it existed when `download` was called within the streamlit app.

### DIFF
--- a/lib/pymedphys/_experimental/streamlit/apps/download.py
+++ b/lib/pymedphys/_experimental/streamlit/apps/download.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import io
 import pathlib
 
 from pymedphys._imports import streamlit as st
@@ -34,3 +35,6 @@ def main():
     filepath = THIS
 
     download(filename, filepath)
+
+    buffer = io.BytesIO("Some beautiful text!".encode())
+    download("a_text_file.txt", buffer)
--- a/lib/pymedphys/_experimental/streamlit/apps/download.py
+++ b/lib/pymedphys/_experimental/streamlit/apps/download.py
@@ -29,8 +29,12 @@ THIS = pathlib.Path(__file__).resolve()
 
 def main():
     st.write(
-        "This is a demo Streamlit app showing how to use the `download()` function"
+        "This is a demo Streamlit app showing how to use the "
+        "`download` function. The docstring for the "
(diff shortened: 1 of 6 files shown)


SimonBiggs:
Alright :smile:, I have now changed the API to only accept file contents. And the file contents can be in string form or bytes form. I have also added in a demo of saving to an in-memory file and downloading a matplotlib figure.

Matthew-Jennings:
Thanks mate, I probably won't get to take this for a spin until Monday, I'm sorry. But looks good!

SimonBiggs:
> Thanks mate, I probably won't get to take this for a spin until Monday, I'm sorry. But looks good!

No need to be sorry :slightly_smiling_face:

sjswerdloff (approved):

SimonBiggs:
Thanks for the review @sjswerdloff :slightly_smiling_face:.

Issue #1391, opened 2021-02-03
Inconsistent delivery data from Mosaiq


SimonBiggs:
Hi @dg1an3,

I thought you might have some ideas about this one. A while back I was extracting MU, Gantry and Col Angles as well as MLC and Jaw positions from Mosaiq. Intermittently I would get some weird disagreements. So... I made the following work-around:

https://github.com/pymedphys/pymedphys/blob/c5fa262fd5eb927dcbbe1fced3469f13ad7f1a96/lib/pymedphys/_mosaiq/delivery.py#L295-L313

Essentially the gist there is I retrieve the delivery information from Mosaiq twice, and then if the consecutive calls disagree with one another... I retrieve it again... until consecutive calls do agree. ...not a great solution...

I'm not sure if the issue is with Mosaiq, MS-SQL, or FreeTDS (the driver behind pymssql), pymssql, or PyMedPhys.

Here's the SQL code that the database is being queried with:

https://github.com/pymedphys/pymedphys/blob/c5fa262fd5eb927dcbbe1fced3469f13ad7f1a96/lib/pymedphys/_mosaiq/delivery.py#L249-L292

Anyway... figured if you're opening up the PyMedPhys Mosaiq code I should let you know about the skeletons before they fall out and scare you 🙂

---

...and while we're at it... here's probably the other main skeleton in that part of the code base:

https://github.com/pymedphys/pymedphys/blob/c5fa262fd5eb927dcbbe1fced3469f13ad7f1a96/lib/pymedphys/_mosaiq/delivery.py#L172-L196

When loading up the MLCs, in some cases there was a byte missing within the database... so... I tried to make a sensible interpretation to restore the data when loading it.

dg1an3:
Hi @SimonBiggs -- do you know what is changing between the first and second queries?  When I looked more closely at the table definition for TxFieldPoint, it occurred to me that the query:
```
SELECT 
    TxFieldPoint.[Index],
    TxFieldPoint.A_Leaf_Set, 
    TxFieldPoint.B_Leaf_Set, 
    TxFieldPoint.Gantry_Ang, 
    TxFieldPoint.Coll_Ang, 
    TxFieldPoint.Coll_Y1, 
    TxFieldPoint.Coll_Y2 
FROM TxFieldPoint 
WHERE 
    TxFieldPoint.FLD_ID = %(field_id)s
```

[...]

PR #1415, opened 2021-02-08
Add rowvers to mock txfield

dg1an3:
Attempting to include a rowversion in mock txfield, to test checking for changed rowversion

### DIFF
--- a/lib/pymedphys/_mosaiq/delivery.py
+++ b/lib/pymedphys/_mosaiq/delivery.py
@@ -261,7 +261,7 @@ def delivery_data_sql(connection, field_id):
     """
     txfield_results = api.execute(
         connection,
-        f"""
+        """
         SELECT
             TxField.Meterset,
             TxField.RowVers
@@ -281,7 +281,7 @@ def delivery_data_sql(connection, field_id):
     txfieldpoint_results = np.array(
         api.execute(
             connection,
-            f"""
+            """
         SELECT
             TxFieldPoint.[Index],
             TxFieldPoint.A_Leaf_Set,
--- a/lib/pymedphys/_mosaiq/delivery.py
+++ b/lib/pymedphys/_mosaiq/delivery.py
@@ -274,6 +274,9 @@ def delivery_data_sql(connection, field_id):
         },
     )
     assert len(txfield_results) == 1
+    txfield_results[0] = txfield_results[0][0], struct.unpack(
+        "Q", txfield_results[0][1]
+    )
 
     txfieldpoint_results = np.array(
         api.execute(
@@ -298,10 +301,11 @@ def delivery_data_sql(connection, field_id):
         )
     )
     assert len(txfieldpoint_results) >= 1
-
-    txfield_results[-1] = struct.unpack("Q", txfield_results[-1])
     for one_point in txfieldpoint_results:
+        # convert to list, so we can change last element
+        one_point = list(one_point)
         one_point[-1] = struct.unpack("Q", one_point[-1])
+        one_point = tuple(one_point)
 
     return txfield_results, txfieldpoint_results
 
--- a/lib/pymedphys/_mosaiq/delivery.py
+++ b/lib/pymedphys/_mosaiq/delivery.py
@@ -248,7 +248,7 @@ def collimation_to_bipolar_mm(mlc_a, mlc_b, coll_y1, coll_y2):
     return mlc, jaw
 
 
-def delivery_data_sql(connection, field_id, include_rowvers=False):
+def delivery_data_sql(connection, field_id):
     """Get the treatment delivery data from Mosaiq given the SQL field_id
 
     Args:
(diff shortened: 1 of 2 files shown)


dg1an3:
Hi @SimonBiggs -- I don't if you had looked any further at #1391; I was looking at how to add the rowversion to the "mock" data, and found (I think) that a BINARY-type column with a long int serial number acts similar to the SQL Server rowversion.

So I tried adding logic to capture the rowversion and log output if there is a difference.

lib/pymedphys/_mosaiq/delivery.py
```diff
@@ -281,36 +286,61 @@ def delivery_data_sql(connection, field_id):
...
-            reference_results = test_results
-            test_results = delivery_data_sql(connection, field_id)
+            reference_txfield_results = test_txfield_results
+            reference_txfieldpoint_results = test_txfieldpoint_results
+            test_txfield_results, test_txfieldpoint_results = delivery_data_sql(
+                connection, field_id, include_rowvers=True
+            )
 
-    return test_results
+    # TODO: remove rowverse from test_results
```
dg1an3, line 342: should the rowversion be removed before returning?

lib/pymedphys/_mosaiq/delivery.py
```diff
@@ -246,7 +248,7 @@ def collimation_to_bipolar_mm(mlc_a, mlc_b, coll_y1, coll_y2):
     return mlc, jaw
 
 
-def delivery_data_sql(connection, field_id):
+def delivery_data_sql(connection, field_id, include_rowvers=False):
```
SimonBiggs, line 251: The only place in the codebase where this is utilised is below within `fetch_and_verify_mosaiq_sql`. So you are free to adjust the API instead to just always `include_rowvers` in the return :slightly_smiling_face:.

Given it's only used internally, and only used within this module, it might as well also be dropped to be a private function.

lib/pymedphys/_mosaiq/delivery.py
```diff
@@ -281,36 +286,61 @@ def delivery_data_sql(connection, field_id):
             TxFieldPoint.Coll_Ang,
             TxFieldPoint.Coll_Y1,
             TxFieldPoint.Coll_Y2
+            {',TxFieldPoint.RowVers' if include_rowvers else ''}
```
SimonBiggs, line 289: The above comment about being able to return `RowVers` all the time should be able to simplify things here :slightly_smiling_face:

lib/pymedphys/_mosaiq/delivery.py
SimonBiggs: The only place this function here is utilised is within the following class method (which is within this module):

https://github.com/pymedphys/pymedphys/blob/fdb7fd1781416b3d31a69fb4af5af50933beb1c3/lib/pymedphys/_mosaiq/delivery.py#L384-L420

Everything in that function is using positive integer indexing so permanently allowing a new item to be included at the end shouldn't interfere. I think, remove all branching logic regarding `RowVers` and just include it all through.

---

Also, of note, the public API where all this code here is exposed is via:

```python
import pymedphys

delivery = pymedphys.Delivery.from_mosaiq(connection, field_id)
```

So as long as the output of that remains constant, we have quite a bit of flexibility within these internal functions here.

SimonBiggs (approved):
Hi @dg1an3,

This is awesome :slightly_smiling_face:, thank you. I have added a few comments to the code below, none of them are required changes, so feel free to merge once you're happy.

> Hi @SimonBiggs -- I don't if you had looked any further at #1391; I was looking at how to add the rowversion to the "mock" data, and found (I think) that a BINARY-type column with a long int serial number acts similar to the SQL Server rowversion.
>
> So I tried adding logic to capture the rowversion and log output if there is a difference.

I haven't yet had a look no.

SimonBiggs:
> I don't if you had looked any further at #1391

Also, I'm going on leave this coming Friday, so I'm going to be AWOL for two weeks. Chances are both on the lead up to this Friday any coding focus will be around the WLutz Arc, and after I get back there is usually a week or so where attention needs to be focused to items that came up while I was away.

So... long way of saying... I likely won't be assigning focus to #1391 myself in the near future.

lib/pymedphys/_mosaiq/delivery.py
```diff
@@ -259,58 +261,85 @@ def delivery_data_sql(connection, field_id):
     """
     txfield_results = api.execute(
         connection,
-        """
+        f"""
```
SimonBiggs, line 264:
```suggestion
        """
```

lib/pymedphys/_mosaiq/delivery.py
```diff
@@ -259,58 +261,85 @@ def delivery_data_sql(connection, field_id):
...
+        {
+            "field_id": field_id,
+        },
     )
 
     txfieldpoint_results = np.array(
         api.execute(
             connection,
-            """
+            f"""
```
SimonBiggs, line 280:
```suggestion
            """
```

PR #1419, opened 2021-02-09
Add session logic and unit tests

dg1an3:
with mock data.  There is a notebook that demonstrates this [`here`](https://github.com/pymedphys/pymedphys/blob/main/prototyping/mosaiq/mosaiq-sessions-protocols/demo_session_offsets.ipynb)

[NOTE] Recreating the branch on pymedphys/pymedphys, to see if it avoids the propagation failure that happened previously.


lib/pymedphys/_mosaiq/sessions.py
```diff
@@ -0,0 +1,186 @@
+# session and session offset calculator
```
SimonBiggs, line 1: Add in copyright 🙂

lib/pymedphys/_mosaiq/sessions.py
```diff
@@ -0,0 +1,186 @@
...
+
+from datetime import datetime, timedelta
+
+from pymedphys._imports import sklearn
+
+from pymedphys import mosaiq
```
SimonBiggs, line 7: This might cause a circular import down the line. Normally I make sure to call the most public API from the tests, but internally within PyMedPhys I usually call the internal APIs

lib/pymedphys/_mosaiq/sessions.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def cluster_sessions(tx_datetimes, interval=timedelta(hours=3)):
```
SimonBiggs, line 10:
```suggestion
def cluster_sessions(tx_datetimes: List[datetime], interval=timedelta(hours=3)):
```

lib/pymedphys/tests/mosaiq/create_mock_data.py
```diff
@@ -11,6 +14,20 @@
 sa_user = "sa"
 sa_password = "sqlServerPassw0rd"
 
+# vary the number of fractions a bit
+number_of_fractions = [20, 25, 30]
```
SimonBiggs, line 18: Just a nit-pick, more just preference.

```suggestion
NUMBER_OF_FRACTIONS = (20, 25, 30)
```

Making the top level constant use all-caps to symbolise constant, and utilise `tuple` for constants so it is clear that the variable won't be mutated.

lib/pymedphys/_mosaiq/sessions.py
```diff
@@ -0,0 +1,186 @@
...
+        {"sit_set_id": sit_set_id},
+    )
+
+    # cluster_sessions expects a sorted list, so extract
+    #   the Tx_DtTm value from each row
+    dose_hst_datetimes = [row[0] for row in result]
+    return cluster_sessions(dose_hst_datetimes)
+
+
+def session_offsets_for_site(connection, sit_set_id, interval=timedelta(hours=1)):
```
SimonBiggs, line 126: Given we now have the capacity, can we include a test that flexes this and `sessions_for_site` within this PR?

SimonBiggs (commented):
Hi @dg1an3,

To help my review, can we have a test that flexes the usage of `session_offsets_for_site`?

SimonBiggs:
For that test failure I believe sklearn just needs to have "mosaiq" added as a group.

pyproject.toml
```diff
@@ -88,8 +88,7 @@ pylinac = { version = "2.3.2", optional = true }  # groups = ["user", "dev", "do
 
 # Scikit-learn has an upper bound since there appears to be an incompatibility
 # between pylinac and scikit-learn version 0.24.0
-# Not needed as PyMedPhys dependency. Included only as a fix.
-scikit-learn = { version = "<0.24.0", optional = true }  # groups = ["user", "dev"]
+scikit-learn = { version = "<0.24.0", optional = true }  # groups = ["user", "dev", "mosaiq"]
```
dg1an3, line 91: is this all that needs to be changed, so that scikit-learn is available during the MosaiqDbTest?

SimonBiggs:
Hehe :) beat me to it :)

pyproject.toml
SimonBiggs: A propagate will be needed, but @pymedphys-bot should be able to fix that for you in a minute or so.

SimonBiggs: Also, is there a reason for the upper pinning?

SimonBiggs: Beautiful 🙂, the tests pass 🙂

dg1an3: would you like me to try removing the upper pinning and see if that causes any apparent problems?  are there tests involving pylinac that you think would flag if there were problems?

SimonBiggs: Yup, it should flag the issue in the tests.

dg1an3: ok I can try in a separate PR

SimonBiggs (approved):
Yup, beautiful 🙂.

LGTM.

PR #1424, opened 2021-02-10
Fix session notebook and try removing scikit-learn pin

dg1an3:
removing obsolete mosaiq session logic from prototyping/mosaiq
changed scikit-learn pin to allow any version


SimonBiggs:
Before merge run `poetry upgrade` as it is actually still stuck with the pins created with the previous version.

SimonBiggs:
Maybe this was the issue?

https://scikit-learn.org/stable/whats_new/v0.24.html#packaging

... ... I wish past Simon had better documented what the issue was ...

SimonBiggs (approved):

PR #1430, opened 2021-02-12
Add notebook exploring the use of xarray.DataArray as a dose object.

Matthew-Jennings:


sjswerdloff:
Do you want me to wait until you have addressed build failures?
In general I prefer to wait, unless the build failures are just the ones that are not required.
But if you need a quick look or know its going to be a slog and you want to have me look at a few commits at a time I can.
But PRs are really best handled in one review.  I don't like approving a PR and then seeing a dozen massive changes after that.

Matthew-Jennings:
> Do you want me to wait until you have addressed build failures?
In general I prefer to wait, unless the build failures are just the ones that are not required.
But if you need a quick look or know its going to be a slog and you want to have me look at a few commits at a time I can.
But PRs are really best handled in one review. I don't like approving a PR and then seeing a dozen massive changes after that.

Sorry, (at least one of) the tests were failing because I made small corrections to the library code that have corresponding, strict baseline tests that I can't easily update. I've abandoned those changes now - this probably isn't the place anyway.

I think the CI should now pass.

I've trying to design a dose/coord handling solution that would allow the user to effortlessly and correctly plot dose in whichever plane/slice in whichever coordinate system they choose for any orthogonal patient orientation - and for it all to be explicit and transparent. XArray appears to be an excellent tool to assist with this. 

I'm envisaging an XArray to be the basis of all dose handling and wanting to get your thoughts/expertise, basically. What woudl you like to see?

prototyping/xarray/xarray for dose.ipynb
```diff
@@ -0,0 +1,263 @@
+{
```
sjswerdloff, line 1: <p>While it might increase scope dramatically, it's useful to have an object representing the orientation be displayed as well (e.g. a "humaniform phantom outline" or a Cube with the S/I/A/P/L/R shown on the appropriate part of the surface).</p><p>And it's important to identify what variable comes in to play that causes one reference frame to no longer be the same as a parent reference frame (e.g. IEC 61217 (a.k.a. Table Top) ceases to be the same as IEC FIXED when the Couch Angle (Patient Support Angle) is no longer 0 degrees.  Robot tables in the room prefer to "think" in IEC FIXED.  But RT Plan is in terms of IEC 61217.  That these are not the same doesn't become apparent until the couch is no longer at 0 degrees.  </p><p>If your purpose is to visualise the dose to a water phantom *as if one were looking at the water phantom*, IEC FIXED is probably fine.  If your purpose is to visualise the dose to a water phantom *as if it was dose in the patient while looking at the patient on the table*, you may want to map to IEC 61217.  But if you want to view it as if it is on the images of the patient, then you may want to use DICOM (and not IEC PATIENT).</p><p>The moment you start expressing concern about the patient's orientation, you aren't really dealing with IEC FIXED any more.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymedphys/pymedphys/pull/1430/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

Matthew-Jennings: > While it might increase scope dramatically, it's useful to have an object representing the orientation be displayed as well (e.g. a "humaniform phantom outline" or a Cube with the S/I/A/P/L/R shown on the appropriate part of the surface).

Agree, this is just a start to demo XArray. I actually have taken a nice set of 8 CT scans of a coffee cup at each of the orthogonal orientations. In each scan, the coffee cup is oriented according to the scan orientation, where the cup opening represents the face direction and the handle represents the head direction. I have corresponding RTDOSES for these scans, and I'd hoped to use them for demoing exactly this. I'm just not sure of the latest protocol for test data in PyMedPhys. Where would I upload said doses? Are they going to be too large? Actually, their empty equivalents (PixelData wiped) are already in the testing data from a couple of years ago.

> And it's important to identify what variable comes in to play that causes one reference frame to no longer be the same as a parent reference frame (e.g. IEC 61217 (a.k.a. Table Top) ceases to be the same as IEC FIXED when the Couch Angle (Patient Support Angle) is no longer 0 degrees. Robot tables in the room prefer to "think" in IEC FIXED. But RT Plan is in terms of IEC 61217. That these are not the same doesn't become apparent until the couch is no longer at 0 degrees.

Excellent point, I hadn't considered table top rotation. I'd be in favour of replacing IEC FIXED with IEC 61217.

I expect that these objects are more likely to be used to view patient doses (as opposed to water phantom doses).

Matthew-Jennings: Let me elaborate a little on how I'd like to be able to view the dose here. It helps to conceptually separate the pixel grid from its corresponding coordinates. Let's say that the grid is indexed as `grid[k, i, j]`

Imagine we're viewing the dose to a patient lying on the table top in the transverse. The patient is HFS, the table top is 0, gantry at 0 (up). I'd like for the user to be looking at a grid that is depicted as though the user were looking at the patient from the far end of the couch (inf-end in this case). The key is: I'd like for this to *always* be the case, regardless of patient orientation. I.e., `grid[0, 0, 0]` is at the upper left corner of the image and farthest from the gantry, and in IEC Table Top, `k` increases with `y` , `i` _decreases_ with `z` and `j` increases with `x`.

I'd like to provide support for the user to view the grid with axes labelled according to the IEC Patient coordinate system as well as IEC Table Top, but in this case, they match anyway

Hopefully that makes sense.

Now, if we are to consider a new orientation, say FFP, my intention is for the grid _not to rotate according to the patient orientation_. I.e., the patient appears face down in the grid and `k` would now decrease with `z` in IEC Table Top. In IEC Table Top, this would mean that the axis labels would match (assuming same extent). But in IEC patient, some of the axis labels would *change in sign*. So, it's still as if we're looking from the far end of the couch, but we have the option to *label* the axes according to different coord systems; namely e.g. DICOM, IEC PATIENT, IEC TABLE TOP (IEC FIXED support would be complicated and I think fine to ignore).

Does that all make sense?

Incidentally, RayStation depicts the patient as I've described above but with axis labels in IEC Patient. But I have seen some viewers *rotate the grid with orientation!* I feel like this is so counterintuitive for most, it's almost a bug.

sjswerdloff: The preferred view you describe is IEC61217.  You are viewing it from the foot of the Table Top.

It makes sense to me.
That is the perspective of the clinician observing the patient, more or less.
It's the view a Radiologist takes of a CT.

prototyping/xarray/xdose.py
```diff
@@ -0,0 +1,198 @@
...
+XARRAY_DIMS_BY_COORD_SYSTEM_AND_ORIENT = {
+    "S": ["y", "z", "x"],
+    "P": ["y", "z", "x"],
+    "P dec": ["y", "x", "z"],
+    "D": ["z", "y", "x"],
+    "D dec": ["z", "x", "y"],
+}
+
+
+def xdose_from_dataset(ds, coord_system="SUPPORT", decimals_to_round_coords=-1):
```
SimonBiggs, line 18: I'm not convinced that this function should handle rounding at all. That seems like something the user can do with the object later if they desire...

prototyping/xarray/xdose.py
```diff
@@ -0,0 +1,198 @@
...
+        data=dose_values,
+        dims=xr_dims,
+        coords=xr_coords,
+        name="dose",
+        attrs={"units": ds.DoseUnits.title(), "coord_system": coord_system},
+    )
+
+
+def coords_from_dataset(
+    ds: "pydicom.dataset.Dataset", coord_system: str = "SUPPORT"
```
SimonBiggs, line 38: Can use the "Literal" type here for coord_system

prototyping/xarray/xdose.py
```diff
@@ -0,0 +1,198 @@
...
+    is_head_first = dataset_orientation_is_head_first(ds)
+
+    di = float(ds.PixelSpacing[0])
+    dj = float(ds.PixelSpacing[1])
+
+    col_range = np.arange(0, ds.Columns * di, di)
+    row_range = np.arange(0, ds.Rows * dj, dj)
+
+    if is_decubitus:
+        x_support = orientation[1] * position[1] + col_range  # type: ignore
```
SimonBiggs, line 121: Is there a way around this type ignore here? Maybe using type casting if need be?

prototyping/xarray/xdose.py
```diff
@@ -0,0 +1,198 @@
...
+    else:
+        x_support = orientation[0] * position[0] + col_range  # type: ignore
+        z_support = -(orientation[4] * position[1] + row_range)  # type: ignore
+
+    if is_head_first:
+        y_support = position[2] + np.array(ds.GridFrameOffsetVector)
+    else:
+        y_support = -position[2] + np.array(ds.GridFrameOffsetVector)  # type: ignore
+
+    if coord_system.upper() in ("SUPPORT", "IEC SUPPORT", "S"):
```
SimonBiggs, line 132: This case by case if/if else chain here can simplify down to a dictionary lookup.

prototyping/xarray/xdose.py
```diff
@@ -0,0 +1,198 @@
...
+    ds, coord_system: str = "SUPPORT", round_to_decimals: int = -1
+):
+    x, y, z = coords_from_dataset(ds, coord_system=coord_system)
+
+    xcoords = {"x": x, "y": y, "z": z}
+
+    if round_to_decimals >= 0:
+        return {
+            dim: np.round(xc, decimals=round_to_decimals) for dim, xc in xcoords.items()
+        }
```
SimonBiggs, line 185: See comment above, I think rounding here is making this function step outside its purview.

pyproject.toml
```diff
@@ -134,6 +134,7 @@ rope = { version = "*", optional = true }  # groups = ["dev"]
 doc8 = { version = "*", optional = true }  # groups = ["dev"]
 readme-renderer = { version = "*", optional = true }  # groups = ["dev"]
 tabulate = { version = "*", optional = true }  # groups = ["dev", "doctests"]
+xarray = "^0.16.2"
```
SimonBiggs, line 137:
```suggestion
xarray = { version = "*", optional = true }  # groups = ["dev", "user"]
```

SimonBiggs (commented):
This was just a quick pass through, mostly subjective code style items. A key change needed before merge though is how the dependency was introduced. I added a suggested change for that.

The rest of the code is all in prototyping, so feel free to merge as is if you'd like 🙂. Of course before having it land in the "stable library" tests would be needed.

prototyping/xarray/xdose.py
Matthew-Jennings: This doesn't complain now...

prototyping/xarray/xdose.py
SimonBiggs: Yup, `typing_extension` is the way to go until we drop support for Python 3.6/3.7

SimonBiggs (approved):
Given this is in prototyping, lets merge this one as is so as to not leave it hanging around :).

Matthew-Jennings:
> Given this is in prototyping, lets merge this one as is so as to not leave it hanging around :).

Sure! But what about those test failures? =/

SimonBiggs:
> > Given this is in prototyping, lets merge this one as is so as to not leave it hanging around :).
> 
> Sure! But what about those test failures (I don't understand them)? =(

Only tests that are flagged as "Required" need to block a PR merging into the `main` branch:

image

Other test failures can be addressed in targeted PRs and usually have nothing to do with your PR.

Matthew-Jennings:
Great!, Thanks @SimonBiggs !

PR #1431, opened 2021-02-15
Initial code for DICOM Send functionality

pchlap:
Hi @sjswerdloff 

I have finally managed to get my code together for the DICOM send part of this connect module. Would you consider giving it a first pass review? There are a few things I am unsure about so I was hoping to get your feedback. I could go through and flag those questionable areas if you like? But there isn't a massive amount of code so you might prefer just to look through it?

I saw that Simon is on leave for a little while, so of course happy to leave this until he is back to get some feedback from him before continuing with this.

Thanks
Phil


sjswerdloff (commented):
Pydicom. read_file(...) is strict about preamble.  I tend to use a different call that is less fussy.  I don't remember off hand but it's in the pseudonymisation (Streamlit based) app.

I think it might be confusing to have the sender have variables that appear to be self identifying used for the destination AE information. 
An SCU has its own host and AE Title, separate from an arbitrary destination. 

The current code mixes role and destination, but it does separate out the content. 


I will review more thoroughly later this week.

pchlap:
> Pydicom. read_file(...) is strict about preamble. I tend to use a different call that is less fussy. I don't remember off hand but it's in the pseudonymisation (Streamlit based) app.
> 
> I think it might be confusing to have the sender have variables that appear to be self identifying used for the destination AE information.
> An SCU has its own host and AE Title, separate from an arbitrary destination.
> 
> The current code mixes role and destination, but it does separate out the content.
> 
> I will review more thoroughly later this week.

Thanks @sjswerdloff 

Yes definitely happy to rework the module to make the separation between the sender and receiving host, port and AE Titles.

SimonBiggs:
> I saw that Simon is on leave for a little while, so of course happy to leave this until he is back to get some feedback from him before continuing with this.

I haven't had a look at this yet (still on leave) but please don't feel like you need to wait till I'm back 🙂

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -295,3 +297,131 @@ def test_dicom_listener_cli(test_dataset):
...
+            "listen",
+            str(TEST_PORT),
+            "-d",
+            str(receive_directory),
+            "-a",
+            str(scp_ae_title),
+        ]
+
+        with process(listener_command):
+            time.sleep(1)
```
SimonBiggs, line 328: Instead of `time.sleep` here, can there be a tangible thing waited for. For example, the listener command could print something to STDOUT when it's ready, and then this here could wait until that STDOUT item is seen.

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -295,3 +297,131 @@ def test_dicom_listener_cli(test_dataset):
...
+            dicom_sender = DicomSender(
```
SimonBiggs, line 329: @sjswerdloff what are your thoughts on naming here?

Given DICOM is an acronym should this be named `DICOMSender`?

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -295,3 +297,131 @@ def test_dicom_listener_cli(test_dataset):
...
+                host="127.0.0.1", port=TEST_PORT, ae_title=scp_ae_title
+            )
+            assert dicom_sender.verify()
+            dicom_sender.send([test_dataset])
+
+        dcm_files = receive_directory.glob("**/*.dcm")
+        dcm_file = next(iter(dcm_files))
```
SimonBiggs, line 336: Is the `iter` needed here?

```suggestion
        dcm_file = next(dcm_files)
```

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -295,3 +297,131 @@ def test_dicom_listener_cli(test_dataset):
...
+        test_directory = pathlib.Path(tmp_directory)
+        send_directory = test_directory.joinpath("send")
+        send_directory.mkdir()
+        send_file = send_directory.joinpath("test.dcm")
+        test_dataset.save_as(send_file)
+
+        # Need to write the preamble or else we pydicom throws an error when reading the
+        # file... This solution is from: https://github.com/pydicom/pydicom/issues/340
+        # TODO: Not entirely sure why this is needed. Is there a cleaner solution to
+        # this?
```
SimonBiggs, line 368: `dcmwrite` has utilities to automatically handle the preamble:

https://pydicom.github.io/pydicom/dev/reference/generated/pydicom.filewriter.dcmwrite.html

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -295,3 +297,131 @@ def test_dicom_listener_cli(test_dataset):
...
+            "-m",
+            "pymedphys",
+            "dicom",
+            "listen",
+            str(TEST_PORT),
+            "-d",
+            str(receive_directory),
+            "-a",
+            str(scp_ae_title),
+        ]
```
SimonBiggs, line 393: Likely time to refactor out this calling command into its own function that can be passed the items after "listen" as a function argument.

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -295,3 +297,131 @@ def test_dicom_listener_cli(test_dataset):
...
+            "localhost",
+            str(TEST_PORT),
+            str(send_file),
+        ]
+
+        # TODO: I don't love having to sleep in unit tests, but the CLI seems to need it
+        with process(listener_command):
+            time.sleep(1)
+            with process(sender_command):
+                time.sleep(1)
```
SimonBiggs, line 412: That's actually because this "with process" is not designed for use in this way. In these cases I would recommend calling `subprocess` yourself directly. The `check_call` command will wait until the process runs till completion.

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -295,3 +297,131 @@ def test_dicom_listener_cli(test_dataset):
...
+        assert ds.SeriesInstanceUID == test_dataset.SeriesInstanceUID
+        assert ds.StudyInstanceUID == test_dataset.StudyInstanceUID
+        assert ds.PatientID == test_dataset.PatientID
+        assert ds.Modality == test_dataset.Modality
+        assert ds.Manufacturer == test_dataset.Manufacturer
+
+        assert len(ds.BeamSequence) == 1
+        assert (
+            ds.BeamSequence[0].Manufacturer == test_dataset.BeamSequence[0].Manufacturer
+        )
```
SimonBiggs, line 427: This logic seems similar to the above test. Can it be refactored out into a `check_dicom_agrees` function?

SimonBiggs (changes requested):
@pchlap seeing this grow is quite exciting :slightly_smiling_face:. I also had a little bit of a peak at `platipy` and it seems to be coming along quite nicely!

My process of review generally has me reviewing tests first, so that's where I dove in. I have run out of time now though and I didn't manage to get to your actual code. There were a few key items within the tests though for change. So for now I will mark this as "changes requested". Once you've had a chance to address the issues in the tests I'll jump in, re-review those, and then review the actual code itself.

pchlap:
> @pchlap seeing this grow is quite exciting 🙂. I also had a little bit of a peak at `platipy` and it seems to be coming along quite nicely!
> 
> My process of review generally has me reviewing tests first, so that's where I dove in. I have run out of time now though and I didn't manage to get to your actual code. There were a few key items within the tests though for change. So for now I will mark this as "changes requested". Once you've had a chance to address the issues in the tests I'll jump in, re-review those, and then review the actual code itself.

Thanks for the great comments/feedback @SimonBiggs, I'll work through those a bit later in the week  :)

pchlap:
Hi @SimonBiggs 

I have managed to work my way through your comments on the test code. Would be interested to hear your feedback on what could be done better.

Thanks!

lib/pymedphys/tests/dicom/test_connect.py
```diff
@@ -360,68 +385,34 @@ def test_dicom_sender_cli(test_dataset):
...
+        sender_command = prepare_send_command(TEST_PORT, scp_ae_title, send_file)
+
+        with process(
+            listener_command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+        ) as listener_process:
+            stream_output = b""
+            for b in iter(lambda: listener_process.stdout.read(1), b""):
+                stream_output += b
+                if b"Listener Ready" in stream_output:
+                    break
```
SimonBiggs, line 405: Given this pattern is re-used a few times, this should probably be spun out into its own function... To be able to still utilise the "with logic" just with your additions it might be worth making a context manager for this section of code. There is a neat decorator in the Python standard library that can convert a defined function generator with a try finally block into a context manager.

I'm on sick leave at the moment, so I'm just on me phone and didn't get a chance to read the rest of your code.

pchlap: Thanks @SimonBiggs, I will look into that suggestion.

But in the meantime, rest up and get well soon. We'll follow up on the remainder of this when you're back in action.

Cheers
Phil

lib/pymedphys/_dicom/connect/send.py
```diff
@@ -0,0 +1,117 @@
...
+            for dataset in dcm_files:
+
+                if not isinstance(dataset, pydicom.Dataset):
+
+                    try:
+                        dataset = pydicom.read_file(dataset)
+                    except pydicom.errors.InvalidDicomError as ide:
+                        logging.warning(ide)
+                        logging.warning(
+                            "Unable to read DICOM file, skipping: %s", dataset
```
SimonBiggs, line 89: Could we opt for the fail fast philosophy here? Instead have this try except block happen at the highest level function (the CLI), and have this function just be in charge of sending the DICOM files it is provided. If those files are invalid, I think having the `send` method raise an error is appropriate.

SimonBiggs (approved):
@pchlap this is a great addition :slightly_smiling_face:. 

I'm not entirely better... but I did have a very long sleep today... so figured I'd jump on and complete this review :slightly_smiling_face:. 

I have a couple of comments below. If they could be implemented before merging that'd be awesome. I'm going to leave my approval here though so that you can merge once you're done.

lib/pymedphys/tests/dicom/test_connect.py
pchlap: @SimonBiggs @sjswerdloff Any preference here before I merge this PR?

SimonBiggs: Let's keep it as you have it @pchlap. And let's apply the upper camel case version of the following rules throughout the codebase in cases like these in the future:

https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined

SimonBiggs: That style guide linked above agrees with your naming approach.

SimonBiggs: Talk about bike shedding. Sorry about this @pchlap, when looking at pep-8 it says the following:

> When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError

So, let's follow pep-8 when there is ambiguity. So... DICOMSender it is... (If that's okay with you @pchlap

SimonBiggs: You know what... I had a bit more of a read... Let's go with that Google style guide above and not use PEP-8 for this case. See https://stackoverflow.com/a/2853609 for more musings.

So... Just to bounce back and forth one last time, let's stick with `DicomSender`. And as time goes on, let's aim to make the codebase align with that approach to acronyms for reasons of consistency.

SimonBiggs: ... moral of the story ... I can't even boat-shed effectively when sick ...

pchlap: Thanks @SimonBiggs, ok sounds good. Will leave it as is. I will merge this PR and will follow up soon (hopefully later this week) with a DicomQueryRetriever class :)

Issue #1432, opened 2021-02-15
pymedphys experimental dicom listen only binds to localhost


didymo:
The command:
pymedphys experimental dicom listen -d ~/test_data -a PMPSTORE_SCP 10401

only binds to port 10401 on localhost.

sudo netstat -na | sudo grep 10401
tcp        0      0 127.0.0.1:10401         0.0.0.0:*               LISTEN   

sudo nmap  workstationipnumber
Starting Nmap 7.80 ( https://nmap.org ) at 2021-02-16 08:39 AEDT
Nmap scan report for Lion.lan (workstationipnumber)
Host is up (0.0000020s latency).
Not shown: 996 closed ports
PORT    STATE SERVICE
22/tcp  open  ssh
25/tcp  open  smtp
80/tcp  open  http
443/tcp open  https

Nmap done: 1 IP address (1 host up) scanned in 0.14 seconds

there are no firewalls etc etc to block this.

sjswerdloff:
@pchlap I've only used it on localhost... is this working in your environment across hosts?
Any suggestion?

sjswerdloff:
I see that https://github.com/pymedphys/pymedphys/blob/ae9c2bc7e1da09855bff60c10504dd8e389ab032/lib/pymedphys/_dicom/connect/listen.py#L196
isn't including passing in a host, the CLI doesn't offer an argument for specifying which IP address (on the current host) to bind to, and defaults to binding on localhost (127.0.0.1).

At: https://stackoverflow.com/questions/166506/finding-local-ip-addresses-using-pythons-stdlib#166589
I found the following for finding the one with default routing:
```
import socket
def get_ip():
    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    try:
        # doesn't even have to be reachable
        s.connect(('10.255.255.255', 1))
        IP = s.getsockname()[0]
    except Exception:
        IP = '127.0.0.1'
    finally:
        s.close()
    return IP
```
I don't think I can get to this until the weekend, but I'd be happy to review then or ask for your review if I'm the one who gets to it first.

pchlap:
Thanks @sjswerdloff
I will have a go at solving this today. I'll let you know how I go, if I can get something together for you to review then you can pick it apart over the weekend :)

SimonBiggs:
Fixed in #1434

PR #1434, opened 2021-02-18
Allow setting the bind IP address for the listen CLI

pchlap:
Hi @sjswerdloff 

this is an attempt at a fairly minimal fix to #1432. My thinking here was that if we bind to 0.0.0.0 by default then that would usually be the desired behaviour. In case the user wants to bind to a specific IP address, they can now do this with a command line option.

I think this should work, note that I haven't been able to test this on Windows. I also don't know how to make a unit test which confirms the resolution of this issue.

This further highlights the issue you raised in #1431, that there is no separation between the SCU and SCP IP/host. I'd be happy to look into how to best do that here, or possibly leaven that for the larger #1431 pull request.

Looking forward to hearing what you think.

### DIFF
--- a/lib/pymedphys/_dicom/connect/listen.py
+++ b/lib/pymedphys/_dicom/connect/listen.py
@@ -193,7 +193,10 @@ def listen_cli(args):
 
     # Start the listener
     dicom_listener = DicomListener(
-        port=args.port, ae_title=args.aetitle, storage_directory=args.storage_directory
+        host=args.bind,
+        port=args.port,
+        ae_title=args.aetitle,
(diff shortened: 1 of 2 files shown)


sjswerdloff (approved):

sjswerdloff:
Testing can be done by getting the IP address for the current host and using that for the command line for the listener (SCP) as well as for the client (SCU), and again only for the client (to confirm the default works) 

You could confirm that by doing that only for the client in the previous (dev?) release and I would expect it to fail.

I encourage trying these out locally and then create a PR to update the automated tests.

PR #1435, opened 2021-02-23
adding mean and localization offset functions and tests

dg1an3:
- functions to assist with the trend anomaly model
    * mean_session_offset_for_site
    * localization_offset_for_site
- unit tests for the above functions
- added type hints to function definitions, because they were getting kind of confusing
- added random number generators to pass to mock data generation, so we can have deterministic mock data for testing purposes


SimonBiggs:
Ahh beautiful 🙂, will this mean that I don't need to pass timezone info through to functions that already have a Mosaiq connection argument?

It was always a little silly to require both the Mosaiq connection object and the Mosaiq timezone info, when the latter should be pulled from the former.

dg1an3:
@SimonBiggs maybe use of the term "localization" isn't clear here -- this is the Mosaiq "localization offset", which is the correction for a systematic setup error that is applied after trending.  I don't know why Mosaiq calls it a "localization offset", other than to make things more confusing than they need to be...

SimonBiggs:
@dg1an3 when you're keen for me to review give me a heads up by clicking the following:

image

Might be worth also flagging @sjswerdloff and @Matthew-Jennings in case they get the opportunity to get to it first :slightly_smiling_face:

lib/pymedphys/_mosaiq/sessions.py
```diff
@@ -144,7 +149,7 @@ def sessions_for_site(connection: Connection, sit_set_id: int):
 
 def session_offsets_for_site(
     connection: Connection, sit_set_id: int, interval=timedelta(hours=1)
-):
+) -> Iterator[Tuple[int, Optional[np.ndarray]]]:
```
SimonBiggs, line 152:
```suggestion
) -> Iterator[Tuple[int, Optional["np.ndarray"]]]:
```

lib/pymedphys/tests/mosaiq/create_mock_data.py
```diff
@@ -307,16 +339,25 @@ def create_mock_treatment_fields(site_df=None):
...
+):
     """for a given site and set of tx fields, generate treatment session data
     (Dose_Hst and Offset) for randomly chosen treatment interval
+
+    Parameters
+    ----------
+    site_df : [type], optional
+        [description], by default None
+    txfield_df : [type], optional
+        [description], by default None
```
SimonBiggs, line 353: Missing bits 🙂

lib/pymedphys/tests/mosaiq/test_session_logic.py
```diff
@@ -106,22 +120,23 @@ def test_session_offsets_for_site(
...
-                    assert session_offset[0] > previous_session_offset_when
-                assert abs(session_offset[1]) <= 10.0
-                assert abs(session_offset[2]) <= 10.0
-                assert abs(session_offset[3]) <= 10.0
+            if np.any(session_offset):
+                assert np.max(np.abs(session_offset)) <= 10.0
+
+            mean_session_offset = mean_session_offset_for_site(connection, sit_set_id)
+            assert np.any(mean_session_offset)
+            assert np.max(np.abs(mean_session_offset)) <= 10.0
```
SimonBiggs, line 136: Would we be able to have another test here that creates a static, non-random, well defined set of offsets for a new site ID and just verifies that those exact numbers are pulled correctly with `localization_offset_for_site` and that `mean_session_offset_for_site` appropriately returns the mean as expected?

SimonBiggs (approved):
LGTM 🙂. A few comments for small adjustments before merge, but feel free to merge once those are addressed.

SimonBiggs:
I was just setting up the release notes and wanting to give a shout out to all your Mosaiq database related work within this release... and then... I realised... you're not on the team list yet! Would you like to be?

Also, at some point I'd be keen to catch up and say hi. I know that you and @sjswerdloff know each other well. Would there be interest in potentially catching up for beverages with you both over a video chat one "evening"?

dg1an3:
Hi @SimonBiggs sure I would appreciate being added to the teams list!  We can do a virtual get together with Stuart - let me know when would be a good time (you are in the Southern Hemisphere as well, correct?)

SimonBiggs:
> I would appreciate being added to the teams list!

Awesome 🙂, would you be able to make PR that adds yourself? Make the change on the README file that is found within `lib/pymedphys/docs`. Propagate will then copy the changes to everywhere else it needs to be 🙂

PR #1438, opened 2021-03-02
Use Xarray to demo new coords extraction function.

Matthew-Jennings:
Add demo of IEC SUPPORT xdose plotting


prototyping/xarray/xdose.py
```diff
@@ -221,3 +222,102 @@ def _validate_coord_system(coord_system):
                 "\n".join(VALID_COORD_SYSTEMS)
             )
         )
+
+
+def plot_xdose_tcs_at_point(xdose_to_plot, point, coord_system="S"):
```
SimonBiggs, line 227: To encourage the writing of tests during this experimentation stage did you want to be putting these modules within experimental?

prototyping/xarray/xdose.py
```diff
@@ -221,3 +226,112 @@ def _validate_coord_system(coord_system):
                 "\n".join(VALID_COORD_SYSTEMS)
             )
         )
+
+
+def plot_xdose_tcs_at_point(xdose_to_plot, point):
```
SimonBiggs, line 231: Given this is within the `xdose` module, probably could get away with not having xdose in the name. For example, when xdose is imported it would look like this:

```python
import xdose

xdose.plot_tcs_at_point( ... )
```

SimonBiggs:
Some neat things coming to xarray syntax in Python 3.10... :slightly_smiling_face: 

https://www.python.org/dev/peps/pep-0637/#use-cases

SimonBiggs:
Given this is all in prototyping etc, could we get this one merged (I'm aiming to clean up the dangling PRs)

SimonBiggs (approved):

PR #1443, opened 2021-03-02
[Utility PR] Changes between 0.35.0 and 0.36.0

SimonBiggs:


lib/pymedphys/_dicom/connect/listen.py
```diff
@@ -193,7 +193,10 @@ def listen_cli(args):
 
     # Start the listener
     dicom_listener = DicomListener(
-        port=args.port, ae_title=args.aetitle, storage_directory=args.storage_directory
+        host=args.bind,
```
SimonBiggs, line 196: @pchlap before I make the `0.36.0` release would we be able to refine this API to address the commend over at

https://github.com/pymedphys/pymedphys/commit/1854ba1d3f6f6e6a6a6b02863f64c63eac8c5fde#r47727955

PR #1448, opened 2021-03-05
add dose constraint checks to chartchecks

rembishj:
Added dose constraint checks to the transfer check application.


SimonBiggs:
@rembishj did you want a hand working out what the following is asking?

image

rembishj:
> @rembishj did you want a hand working out what the following is asking?
> 
> image

@SimonBiggs some insight would be appreciated. I looked into how to use enumerate, and was not really seeing the benefit of it in this situation. But it's very possible that I also just overlooked something simple. 

For the nested loops, I wasn't quite sure how to tackle that problem as it was within the main function for the streamlit transfer check. Just a bunch of different parts creating the widgets.

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -70,185 +58,405 @@ def main():
...
-                    "rx"
-                ].values[0],
-            )
+    display_results = display_results.style.apply(colour_results, axis=1)
+    st.dataframe(display_results.set_precision(2), height=1000)
+
+
+def get_structure_aliases():
+    cwd = os.getcwd().replace("\\", "/")
+    file_path = cwd + "/lib/pymedphys/_experimental/chartchecks/ALIASES.csv"
```
SimonBiggs, line 215: cwd won't always be the root of pymedphys 🙂, so this'll bite you later. If you were to create a file within `_experimental/chartchecks` that had the following function that should help:

```python
def get_aliases():
    file_path = pathlib.Path(__file__).parent.joinpath('ALIASES.csv')
    return pd.read_csv(file_path)
```

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -70,185 +58,405 @@ def main():
...
+    alias_df = pd.read_csv(file_path)
+    for i in range(len(alias_df.keys())):
+        df_list = alias_df.iloc[0][i][1:-1].split((","))
+        formatted_df_list = []
+        for item in df_list:
+            formatted_df_list.append(item.replace("'", "").strip(" "))
+        alias_df.iloc[0][i] = formatted_df_list
```
SimonBiggs, line 222: Regarding the pylint message, instead of using enumerate as it suggested, I would recommend to swap to using iteritems. See:

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.iteritems.html

Then instead of calling `alias_df.iloc[0][i]` utilise the column content returned by that above method.

Also... as a side note, I would recommend no replacing the dataframe inplace (`alias_df.iloc[0][i] = formatted_df_list`) instead, create a new dataframe that has the formatting be undergone. This is more a decision of taste, but I find in-place changing of dataframes can get one in trouble down the line.

I still do sometimes do in-place replace dataframes. But it can make it quite hard to reason about what is going on. An example of what can happen... you could accidently be changing the output of a cached streamlit function for example.

SimonBiggs (commented):
Looks like a lot of great work here 🙂 There's a lot going on so I'm not going to dive in and review everything here. However, I did provide a comment that I hope helps with a strategy to fix those pylint messages 🙂

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
rembishj: I knew this was going to be a problem whenever I was writing it, but wasn't sure the best way to go about it! Should be taken care of now. 😄  I've left the path for the database set to my personal computer though, since it contains patient mrns and I do not want that committed.

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
rembishj: I've gone ahead and switched the Aliases to json instead of csv, and this actually simplified things to where I no longer have this problem! I also fixed the other enumerate issues.

Any recommendations on the too many nested-blocks issue?

SimonBiggs: > Any recommendations on the too many nested-blocks issue?

I'm just on my phone, so can't see the actual code itself easily. But if you've got too much going on in a function I'd recommend spinning out some sections of the code into separate functions.

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -70,199 +61,500 @@ def main():
...
+    show_table = st.checkbox("View complete Mosaiq table.")
+    if show_table:
+        st.subheader("Mosaiq Table")
+        st.dataframe(mosaiq_table, height=1000)
+
+
+def add_alias(dvh_calcs, ALIASES):
+    define_alias = st.checkbox("Define a new structure alias")
+    if define_alias:
+        add_new_structure_alias(dvh_calcs, ALIASES)
```
SimonBiggs, line 447: All caps names are usually used to denote "constants", so I don't think that fits here.

SimonBiggs (commented):
A few comments :slightly_smiling_face:.

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -70,199 +61,500 @@ def main():
...
+    When exporting the DICOM, only the RP is needed for the data transfer check. Once you have that, you can select it where prompted and the application
+    will run.
+
+    To evaluate the DVH, you will need to import the RD and RS files as well.
+    """
+    )
+
+    files = get_patient_files()
+
+    if "rp" in files:
```
SimonBiggs, line 475:
```suggestion
    if not "rp" in files:
        st.stop()
```

Then the nesting of all the below can be reduced by 1.

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -70,199 +61,500 @@ def main():
...
+        if len(selected_label) != 0:
+            show_field_rx(dicom_table, selected_label)
+            check_for_field_approval(mosaiq_table, field_selection, connection)
+            show_comparison_of_selected_fields(dicom_field_selection, results)
+            show_fx_pattern_and_comments(mosaiq_table, field_selection)
+
+        show_dicom(dicom_table)
+        show_mosaiq(mosaiq_table)
 
         if "rs" in files and "rd" in files:
```
SimonBiggs, line 509:
```suggestion
        if not "rs" in files or not "rd" in files:
            st.stop()
```

Then all the below indenting can be reduced by 1.

lib/pymedphys/_experimental/streamlit/apps/transfer_check.py
```diff
@@ -70,199 +61,500 @@ def main():
...
+            dd_input: pydicom.FileDataset = pydicom.dcmread(files["rd"], force=True)
+            ds_input: pydicom.FileDataset = pydicom.dcmread(files["rs"], force=True)
+
             show_dvh = st.checkbox("Create DVH Plot")
             if show_dvh:
```
SimonBiggs, line 514: Split `show_dvh` out into its own function. Then go:

```python
if st.checkbox("Create DVH Plot"):
    show_dvh(ds_input, dd_input, etc, ...)
```

SimonBiggs (commented):
A few comments specifically targeted at simplifying the nested if statements within `main()`.

ALIASES.json
```diff
@@ -0,0 +1 @@
...,"r optic n"]},"Oral_cavity":{"0":["oral_cavity","oral cavity"]},"Orbit_L":{"0":["orbital_l","orbital l","l orbit","lt orbit","left_orbit","left orbit","orbit_left","orbit left","eye_l","l eye"]},"Orbit_R":{"0":["orbital_r","orbital r","r orbit","rt orbit","right_orbit","right orbit","orbit_right","orbit right","eye_r","r eye"]},"Parotid_L":{"0":["parotid_l","left parotid","l parotid","l_parotid","lt parotid"]},"Parotid_R":{"0":["parotid r","parotid_r","right parotid","r parotid","r_parotid","rt parotid","rt_parotid"]},"Parotid_unilateral":{"0":["parotid_unilateral"]},"Penile_bulb":{"0":["penilebulb","penile bulb","penile_bulb","bulb"]},"Pharyngeal_constrictors":{"0":["pharyngeal_constrictors"]},"Rectum":{"0":["rectum"]},"Sigmoid":{"0":["sigmoid"]},"Small_bowel":{"0":["small_bowel","small bowel","bowel_small","bowel small"]},"Spinal_cord":{"0":["spinalcord","cord","spinal cord","s cord"]},"Spinal_cord_PRV":{"0":["spinal cord plus 5mm","spinalcord plus 5mm","spinal cord + 5mm","spinalcord + 5mm"]},"Stomach":{"0":["Stomach"]},"Temporal_lobes":{"0":["temporal_lobes","temporal lobe"]},"Tongue":{"0":["tongue"]},"Total_Lung":{"0":["lungs","total lung-gtv","total lung-itv","total lung"]}}
```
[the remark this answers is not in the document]
rembishj: No it was not, I will delete this from there!

lib/pymedphys/_experimental/chartchecks/ALIASES.json
```diff
@@ -0,0 +1 @@
...,"r optic n"]},"Oral_cavity":{"0":["oral_cavity","oral cavity"]},"Orbit_L":{"0":["orbital_l","orbital l","l orbit","lt orbit","left_orbit","left orbit","orbit_left","orbit left","eye_l","l eye"]},"Orbit_R":{"0":["orbital_r","orbital r","r orbit","rt orbit","right_orbit","right orbit","orbit_right","orbit right","eye_r","r eye"]},"Parotid_L":{"0":["parotid_l","left parotid","l parotid","l_parotid","lt parotid"]},"Parotid_R":{"0":["parotid r","parotid_r","right parotid","r parotid","r_parotid","rt parotid","rt_parotid"]},"Parotid_unilateral":{"0":["parotid_unilateral"]},"Penile_bulb":{"0":["penilebulb","penile bulb","penile_bulb","bulb"]},"Pharyngeal_constrictors":{"0":["pharyngeal_constrictors"]},"Rectum":{"0":["rectum"]},"Sigmoid":{"0":["sigmoid"]},"Small_bowel":{"0":["small_bowel","small bowel","bowel_small","bowel small"]},"Spinal_cord":{"0":["spinalcord","cord","spinal cord","s cord"]},"Spinal_cord_PRV":{"0":["spinal cord plus 5mm","spinalcord plus 5mm","spinal cord + 5mm","spinalcord + 5mm"]},"Stomach":{"0":["Stomach"]},"Temporal_lobes":{"0":["temporal_lobes","temporal lobe"]},"Tongue":{"0":["tongue"]},"Total_Lung":{"0":["lungs","total lung-gtv","total lung-itv","total lung"]}}
```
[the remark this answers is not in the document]
rembishj: It looks like pandas.to_json already has it's own option for indent. I've gone ahead and set that to 4, and I believe it's achieving the same results you were looking for? Let me know if there's something specific about the json.dump approach you'd prefer

lib/pymedphys/_experimental/chartchecks/compare.py
```diff
@@ -252,3 +253,20 @@ def specific_patient_weekly_check_colour_results(val):
...
         return ["background-color: #EE6363"] * len(val)
+
+
+def constraint_check_colour_results(val):
+    if val["Type"] != "Average Score" and val["Type"] != "Total Score":
+        diff = val["Dose [Gy]"] - val["Actual Dose [Gy]"]
+        limit = val["Dose [Gy]"] / 10
+        if val["Actual Dose [Gy]"] > val["Dose [Gy]"]:
+            return ["background-color: #EE6363"] * len(val)
+        elif 0 < diff < limit:
```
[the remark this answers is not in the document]
rembishj: I'm not sure I'm following here... I have a total of 3 possible outcomes here, so do I not need the elif?

lib/pymedphys/_experimental/chartchecks/weekly_check_helpers.py
```diff
@@ -159,3 +164,50 @@ def plot_couch_deltas(delivered):
...
+
+    columns = list(dataframe_column_to_sql_reference.keys())
+    select_string = "SELECT " + ",\n\t\t    ".join(
+        dataframe_column_to_sql_reference.values()
+    )
+
+    sql_string = (
+        select_string
+        + """
+                    From Image, Ident
```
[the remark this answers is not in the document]
rembishj: yeah I don't know what happened there... lol

lib/pymedphys/_experimental/chartchecks/helpers.py
```diff
@@ -409,3 +415,26 @@ def get_all_treatment_history_data(connection, mrn):
...
     return treatment_history
+
+
+def get_structure_aliases():
+    file_path = pathlib.Path(__file__).parent.joinpath("ALIASES.json")
+    return pd.read_json(file_path)
+
+
+def add_new_structure_alias(dvh_calcs, alias_df):
+    file_path = pathlib.Path(__file__).parent.joinpath("ALIASES.json")
```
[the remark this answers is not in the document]
rembishj: Could you explain a bit what you mean by this? Define the variable once outside of the functions, and then call it as an argument in each?

lib/pymedphys/_experimental/chartchecks/weekly_check_helpers.py
```diff
@@ -159,3 +164,50 @@ def plot_couch_deltas(delivered):
         xaxis=dict(tickmode="linear", tick0=0, dtick=1),
     )
     st.plotly_chart(deltas_fig, use_container_width=True)
+
+
+def get_patient_image_info(patient):
+    connection = _pp_mosaiq.connect("PRDMOSAIQIWVV01.utmsa.local")
```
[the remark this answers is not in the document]
rembishj: @SimonBiggs I definitely agree. I could spare some time next week if there's a day that works well for you!

lib/pymedphys/_experimental/chartchecks/ALIASES.json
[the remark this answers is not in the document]
SimonBiggs: Nothing specific :slightly_smiling_face:

lib/pymedphys/_experimental/chartchecks/compare.py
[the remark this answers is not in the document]
SimonBiggs: The specific section that won't work in some cases is `0 < diff < limit`.

> do I not need the elif?

No harm in having it, just that can line 264 ever run if the following is true?

`val["Actual Dose [Gy]"] > val["Dose [Gy]"]`

lib/pymedphys/_experimental/chartchecks/helpers.py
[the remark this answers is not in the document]
SimonBiggs: You could go for either:

```python
ALIAS_PATH = pathlib.Path(__file__).parent.joinpath("ALIASES.json")
```

or

```python
def get_alias_path():
    return pathlib.Path(__file__).parent.joinpath("ALIASES.json")
```

Downside of the top one is that it will run that logic on import, meaning if it takes some time to run then it slows down all of PyMedPhys when it is imported. Downside of the second one, it's a little bit more verbose to write.

lib/pymedphys/_experimental/chartchecks/weekly_check_helpers.py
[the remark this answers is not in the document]
SimonBiggs: I'm not sure if next week will work. I'll have to let you know. I'm currently on sick leave :face_with_thermometer:, likely next week will have a lot in it. Anyway, will let you know.

pyproject.toml
```diff
@@ -134,6 +134,7 @@ rope = { version = "*", optional = true }  # groups = ["dev"]
 doc8 = { version = "*", optional = true }  # groups = ["dev"]
 readme-renderer = { version = "*", optional = true }  # groups = ["dev"]
 tabulate = { version = "*", optional = true }  # groups = ["dev", "doctests"]
+fsspec = "^0.8.7"
```
SimonBiggs, line 137: Out of interest, what was this for?

rembishj: Streamlit wasn't wanting to play nice with the indent parameter of pandas.to_json. It would work outside of streamlit, but within the app I would get an error message saying I needed that package. 

Also, am I doing things wrong when adding to poetry? I just did: 
`poetry add <package>`

Just trying to make sure I don't mess anything up in the future

SimonBiggs: > but within the app I would get an error message saying I needed that package

Would you be able to provide a print out of that error message? I've never seen that one before.

> Also, am I doing things wrong when adding to poetry?

For normal poetry usage that would be fine. But we are using it a little bit differently so as to support optional imports. Instead of using `poetry add` I would recommend instead just inserting a line within `pyproject.toml` that looks like:

```toml
fsspec = { version = "*", optional = true }  # groups = ["user", "dev"]
```

And then place it under the appropriate section under https://github.com/pymedphys/pymedphys/blob/4b983b84dad4cc5bb67ac673f273d9c77a2a11d2/pyproject.toml#L37-L39

Then run:

```
poetry install -E dev
```

---

When you are adding dependencies though it is a bit of a big deal. So, I'd recommend trying to reserve the dependency addition for its own PR, and have a reasonable justification for the addition. Hence why I'm trying to be sure that we really do need `fsspec` and can't get around it some other way.

rembishj: image

rembishj: Here is the error message I get when trying to run without fsspec. This only was happening when running in streamlit. I was able to write the dataframe to json outside of streamlit with no issues.

SimonBiggs: Yup, awesome. Makes sense. That's a pandas optional dependency. Let's merge this PR and then could you make a fresh PR where you just add that new dependency by itself?

SimonBiggs (approved):
LGTM :slightly_smiling_face:. I'll merge this now :slightly_smiling_face:.

PR #1471, opened 2021-03-28
[STACK base] Config work for server move

SimonBiggs:
A range of configuration focused changes to make it so that more of the applications draw from the central config. These changes occurred when needing to move some key parts of our server infrastructure.

Most changes are to site-specific or experimental, however, there was a change to the `pymedphys trf orchestrate` API, as it now pulls from the central `config.toml` file. I also made some changes to meterset map to handle an edge case where a Linac doesn't yet have any backup files available.

Also, when running `pymedphys gui` the docs (if they've been built) are now accessible at `localhost:8051/docs/index.html`


SimonBiggs:
Given this one has been hanging around without comment I'm going to merge it in.

PR #1486, opened 2021-04-12
[ARCHIVED] [STACK base] Create Mosaiq Mimic data, fix `pymedphys.Delivery.from_mosaiq`

SimonBiggs:
Pull dummy data for Mosaiq Patient IDs. The goal here is to create a set of CSV files for the demo Patient IDs `989898` and `979797`. I am pulling out the relevant tables that apply to that patient and are used by the `Delivery.from_mosaiq` method.

Then, I'll write a test for that method, show that it is currently failing, fix it, and hopefully never break it again because it will then be undergoing automated testing.

@dg1an3 I'd be keen for your thoughts on my approach here.

---

~~If the request is made, I'd be happy to split this up into smaller PRs.~~ I'm going to split this one up into smaller PRs and wait for reviews before merge.


lib/pymedphys/tests/mosaiq/test_mimicked_db.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@pytest.mark.mosaiqdb
+def test_get_patient_name(
+    create_mimic_db_with_tables,
+):  # pylint: disable = unused-argument
+
+    with _connect.connect(database=mimics.DATABASE) as connection:
+        name = helpers.get_patient_name(connection, PATIENT_ID)
+        assert name == "PHYSICS, Mock"
```
SimonBiggs, line 43: Hi @rembishj,

Some cool news, this PR is able to be used to test the `helpers` functions calling from a testing MSSQL database! 🙂 🥳

SimonBiggs:
Hi @rembishj,

This PR should fix `pymedphys.Delivery.from_mosaiq`. If you merge this branch into your current working branch then you should be able to use that to get the MLC positions.

lib/pymedphys/tests/mosaiq/test_mimicked_db.py
rembishj: Awesome! I'll probably need to meet with you to explain everything that's going on here 😄

rembishj (approved):
LGTM! 😄

rembishj:
> Hi @rembishj,
> 
> This PR should fix `pymedphys.Delivery.from_mosaiq`. If you merge this branch into your current working branch then you should be able to use that to get the MLC positions. Here's an example of how you can use that above object to verify MLCs and other parameters:

So I've merged this branch and am attempting to use the `from_mosaiq()` function, it's working for VMAT fields but not for any of the step-and-shoot fields. Should it be?

SimonBiggs:
> it's working for VMAT fields but not for any of the step-and-shoot fields. Should it be?

When you say step and shoot, do you specifically mean not dMLC?

rembishj:
> When you say step and shoot, do you specifically mean not dMLC?

Yes, that is correct. So the field still has multiple control points with varying MLC positions but is not dynamic. I think it would still be nice to grab the MLC positions for each of these control points to compare to the DICOM.

SimonBiggs:
> I think it would still be nice to grab the MLC positions for each of these control points to compare to the DICOM.

Yes certainly. I think I know why it might not work though... We haven't done step and shoot IMRT for almost 5 years... So I wouldn't have seen it.

If you're happy to just directly read the MLC positions from the Mosaiq SQL query, here's a function that can decode those bytes into MLC positions:

https://github.com/pymedphys/pymedphys/blob/59c8c2818b09069156fae9c68b09dd7de9199c2f/lib/pymedphys/_mosaiq/delivery.py#L212-L236

PR #1494, opened 2021-04-15
[STACK 1] Create Mosaiq table extractor, add csv based testing data

SimonBiggs:


lib/pymedphys/_experimental/streamlit/apps/mosaiq_to_csv.py
```diff
@@ -0,0 +1,251 @@
...
+
+LIB_ROOT = pathlib.Path(__file__).parents[3]
+TEST_DATA_DIR = LIB_ROOT.joinpath("tests", "mosaiq", "data")
+
+ALLOWLIST_TABLE_NAMES = [
+    "Ident",
+    "Patient",
+    "TxField",
+    "TxFieldPoint",
+    "Site",
```
sjswerdloff, line 41: I would include the SiteSetup table as well.
I'm not sure what the goal is here, but "stuff that goes in a DICOM Plan" includes stuff that comes from SiteSetup

sjswerdloff: Speaking of which... what is the goal here (docstring for the file, docstring for the public methods, and no, it doesn't matter that this is in _experimental, it is really helpful to give us reviewers a hand in understanding what is supposed to happen and why, even if that's the only time anyone ever reads the docstring.

sjswerdloff (approved):
I'm not sure about my position to confirm the variable types in the schema but those tend to be stable.
What tends to not be stable is additional elements as MOSAIQ has new releases.
So you might want to have something that identifies what version of MOSAIQ you based your schema knowledge on (a comment in the code?). and you might want to have the CSVs indicate what version of MOSAIQ was queried to populate. the CSVs

lib/pymedphys/_experimental/streamlit/apps/mosaiq_to_csv.py
SimonBiggs: > it is really helpful to give us reviewers a hand in understanding what is supposed to happen and why, even if that's the only time anyone ever reads the docstring

Yup, fair call. I'll run through and do that and link to the changes.

SimonBiggs: > I would include the SiteSetup table as well.
> I'm not sure what the goal is here, but "stuff that goes in a DICOM Plan" includes stuff that comes from SiteSetup

In this case I was only actually including tables on an "as-needs" basis. The whole purpose here is to be able to create the following testing suite:

https://github.com/pymedphys/pymedphys/blob/89f4e76e7e579de9e376e59de5f719342e513c9b/lib/pymedphys/tests/mosaiq/test_mimicked_db.py#L39-L163

So tables were only included on an as-needs basis. Given how easy extraction of a new table and its corresponding types is with this `mosaiq_to_csv` tool when there is a PyMedPhys function written that utilises calls to `SiteSetup`, that can be added at that time.

SimonBiggs:
> I'm not sure about my position to confirm the variable types in the schema but those tend to be stable.
What tends to not be stable is additional elements as MOSAIQ has new releases.
So you might want to have something that identifies what version of MOSAIQ you based your schema knowledge on (a comment in the code?). and you might want to have the CSVs indicate what version of MOSAIQ was queried to populate. the CSVs

If it's okay, I think I'd prefer to leave the Mosaiq version specification when testing across multiple Mosaiq versions becomes the desired purpose.

For now, this here is primarily for regression testing with changes to the Mosaiq sections of the code base. I certainly can see though that it would be desirable to have this tables have multiple versions of Mosaiq and the PyMedPhys code run through each Mosaiq version. That would be quite awesome.

... not today though. For now... these regression tests were non-existent... so just making them exist is enough for this project.

SimonBiggs:
@sjswerdloff I have added in docstrings for the pertinent functions. See the diff-comparison to your recent review:

https://github.com/pymedphys/pymedphys/compare/4dc99dd44c074b8dc1f5920f75086506df4654f1..fa738326da92abc0adbb0f752ec772f99e419838

Hope that helps...

dg1an3 (approved):

PR #1497, opened 2021-04-16
[STACK 1] Add passing mimic tests

SimonBiggs:
Some tests required some bug fixes. These are the tests that didn't only needed very minor library changes.

### DIFF
--- a/lib/pymedphys/tests/mosaiq/test_mimicked_db.py
+++ b/lib/pymedphys/tests/mosaiq/test_mimicked_db.py
@@ -1,0 +1,144 @@
+# Copyright (C) 2021 Cancer Care Associates
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from pymedphys._imports import numpy as np
+from pymedphys._imports import pytest
+
+import pymedphys
+import pymedphys.beta
+from pymedphys._mosaiq import helpers
+
+from . import _connect
+from .data import mimics
+
+PATIENT_ID = 989898
+FIELD_ID = 88043
+A_TREATMENT_DATETIME = "2020-04-27 08:03:28.513"
+MACHINE_ID = "2619"
+FIELD_NAME = "3ABUT"
+TIMEZONE = "Australia/Sydney"
+FIRST_NAME = "MOCK"
+LAST_NAME = "PHYSICS"
+FULL_NAME = f"{LAST_NAME}, {FIRST_NAME.capitalize()}"
+QCL_LOCATION = "Physics_Check"
+AN_UNCOMPLETED_QCL_DUE_DATETIME = "2021-05-21 23:59:59"
+QCL_COMPLETED_DATETIMES = ["2021-04-14 09:11:30.387", "2021-04-14 09:11:35.383"]
+
+
+@pytest.fixture(name="connection")
+def connection_base():
+    """ will create the test database, if it does not already exist on the instance """
+    mimics.create_db_with_tables()
+    return _connect.connect(database=mimics.DATABASE)
+
+
+@pytest.fixture(name="trf_filepath")
+def trf_filepath_base():
+    data_paths = pymedphys.zip_data_paths("metersetmap-gui-e2e-data.zip")
+    date = A_TREATMENT_DATETIME.split(" ")[0]
+
+    filtered_paths = [
+        path
+        for path in data_paths
+        if date in str(path) and str(FIELD_ID) in str(path) and path.suffix == ".trf"
+    ]
+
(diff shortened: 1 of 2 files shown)


lib/pymedphys/tests/mosaiq/test_mimicked_db.py
```diff
@@ -0,0 +1,144 @@
...
+
+PATIENT_ID = 989898
+FIELD_ID = 88043
+A_TREATMENT_DATETIME = "2020-04-27 08:03:28.513"
+MACHINE_ID = "2619"
+FIELD_NAME = "3ABUT"
+TIMEZONE = "Australia/Sydney"
+FIRST_NAME = "MOCK"
+LAST_NAME = "PHYSICS"
+FULL_NAME = f"{LAST_NAME}, {FIRST_NAME.capitalize()}"
```
sjswerdloff, line 34: Why not capitalize the surname?

lib/pymedphys/tests/mosaiq/test_mimicked_db.py
```diff
@@ -0,0 +1,144 @@
...
+    data_paths = pymedphys.zip_data_paths("metersetmap-gui-e2e-data.zip")
+    date = A_TREATMENT_DATETIME.split(" ")[0]
+
+    filtered_paths = [
+        path
+        for path in data_paths
+        if date in str(path) and str(FIELD_ID) in str(path) and path.suffix == ".trf"
+    ]
+
+    return filtered_paths[0]
```
sjswerdloff, line 58: This assumes ordering in the zip is explicitly what you need or can accept.
Do you have a preferred first item / ordering?

lib/pymedphys/tests/mosaiq/test_mimicked_db.py
```diff
@@ -0,0 +1,144 @@
...
+        path for path in data_paths if path.name == f"{PATIENT_ID}_{FIELD_NAME}.dcm"
+    ]
+
+    return filtered_paths[0]
+
+
+@pytest.mark.mosaiqdb
+def test_get_patient_name(connection):
+    name = helpers.get_patient_name(connection, PATIENT_ID)
+    assert name == FULL_NAME
```
sjswerdloff, line 75: Is it worth doing a capitalization insensitive compare of some kind?

lib/pymedphys/tests/mosaiq/test_mimicked_db.py
```diff
@@ -0,0 +1,144 @@
...
+
+    treatments = helpers.get_treatments(connection, start, end, MACHINE_ID)
+    assert np.datetime64(A_TREATMENT_DATETIME) in treatments["start"].tolist()
+
+
+@pytest.mark.mosaiqdb
+def test_trf_identification(connection: pymedphys.mosaiq.Connection, trf_filepath):
+    delivery_details = pymedphys.beta.trf.identify(connection, trf_filepath, TIMEZONE)
+    assert delivery_details.field_id == FIELD_ID
+    assert delivery_details.first_name == FIRST_NAME
```
sjswerdloff, line 105: Is capitalization important?

sjswerdloff (approved):

lib/pymedphys/tests/mosaiq/test_mimicked_db.py
SimonBiggs: Weirdly "capitalize" actually means make the first letter be capital and the rest be lower case.

So in this case FULL_NAME will be "PHYSICS, Mock".

This essentially a choice based on the current API.

lib/pymedphys/tests/mosaiq/test_mimicked_db.py
SimonBiggs: Yeah, I do actually. Some of those TRF files are partial deliveries. Only the first one is a full delivery. Potentially I should be explicitly pulling the first one by name. I'll do that I think.

Cheers

lib/pymedphys/tests/mosaiq/test_mimicked_db.py
SimonBiggs: Potentially, it's just the API at the moment of that function by definition is trying to "pretty print" the name. So, it's sort of testing that as well?

lib/pymedphys/tests/mosaiq/test_mimicked_db.py
SimonBiggs: See above.

lib/pymedphys/tests/mosaiq/test_mimicked_db.py
SimonBiggs: Thanks for this @sjswerdloff. Fixed in c33c7f7c46070ca4ae5e48758cbd8787f7074940

lib/pymedphys/tests/mosaiq/test_mimicked_db.py
SimonBiggs: Fixed in 43edbc618acd60f55eae27d30944342b18b9fb4f

lib/pymedphys/tests/mosaiq/test_mimicked_db.py
SimonBiggs: I'll leave this one, as (at the moment) that specific case is a result of the `helpers.get_patient_name` API.

SimonBiggs:
Thanks for this review @sjswerdloff 🙂. Given this is approved I'll merge in. Let me know if you have any further recommendations stemming from the above adjustments based on your recommendations.

PR #1498, opened 2021-04-16
[STACK 2] Refactor out "get_qcls_by_date"

SimonBiggs:


dg1an3 (approved):

lib/pymedphys/_mosaiq/helpers.py
```diff
@@ -225,16 +225,17 @@ def get_qcls_by_date(connection, location, start, end):
...
             Chklist.Act_DtTm,
-            Chklist.Instructions,
-            Chklist.Notes,
+            Com_Staff.Last_Name,
+            Com_Staff.First_Name,
             QCLTask.Description
-        FROM Chklist, Staff, QCLTask, Ident, Patient
+        FROM Chklist, Staff as Rsp_Staff, Staff as Com_Staff, QCLTask, Ident, Patient
         WHERE
             Chklist.Pat_ID1 = Ident.Pat_ID1 AND
```
sjswerdloff, line 233: Is this an alternative approach to a JOIN?

sjswerdloff (approved):

lib/pymedphys/_mosaiq/helpers.py
SimonBiggs: Yeah... it's the "old style INNER JOIN". These should be refactored to utilise ANSI style JOINs instead. My SQL mentor back in the day utilised this style, so I followed suite. But I have been told a few times that it's not ideal.

Bit of a chicken and the egg problem though. I haven't wanted to refactor these until they were being regressions tested. Happy for this sort of thing to be adjusted once all the tests are in.

sjswerdloff: This might be a left join.
At some point refactoring to ANSI SQL would be good for maintainability

SimonBiggs: > At some point refactoring to ANSI SQL would be good for maintainability

Yup, I agree.

SimonBiggs: Flagged in:

https://github.com/pymedphys/pymedphys/issues/1499

PR #1500, opened 2021-04-20
[STACK 1] Fix `Delivery.from_mosaiq`, remove db repeat

SimonBiggs:
@dg1an3 here I removed that db-repeat that I was doing. I highly suspect it was because I hadn't undergone the ordering appropriately as you highlighted. It does align with the bug I was seeing. And, now we have a test to verify that it is appropriately and consistently pulling from MSSQL. I'd rather that test throw an error if it doesn't match baseline on first db query. So, I removed the retry loop. Thanks for helping me go through the process of pulling out that nasty thorn.

### DIFF
--- a/lib/pymedphys/_mosaiq/delivery.py
+++ b/lib/pymedphys/_mosaiq/delivery.py
@@ -333,13 +333,12 @@ class DeliveryMosaiq(DeliveryBase):
     @classmethod
     def from_mosaiq(cls, connection, field_id):
         total_mu, tx_field_points = delivery_data_sql(connection, field_id)
+        tx_field_points_index = tx_field_points["Index"].to_numpy(dtype=float)
 
-        cumulative_percentage_mu = tx_field_points["Index"].to_numpy(dtype=float)
-
-        if np.shape(cumulative_percentage_mu) == ():
+        if np.shape(tx_field_points_index) == ():
             mu_per_control_point = [0, total_mu]
         else:
-            cumulative_mu = cumulative_percentage_mu * total_mu / 100
+            cumulative_mu = tx_field_points_index / tx_field_points_index[-1] * total_mu
             mu_per_control_point = np.concatenate([[0], np.diff(cumulative_mu)])
 
         monitor_units = np.cumsum(mu_per_control_point).tolist()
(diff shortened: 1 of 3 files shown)


lib/pymedphys/_mosaiq/delivery.py
```diff
@@ -273,128 +264,77 @@ def delivery_data_sql(connection, field_id):
...
-
-    @classmethod
-    def _from_mosaiq_base(cls, connection, field_id):
-        txfield_results, txfieldpoint_results = fetch_and_verify_mosaiq_sql(
-            connection, field_id
-        )
-
-        total_mu = np.array(txfield_results[0]).astype(float)
-        cumulative_percentage_mu = txfieldpoint_results[:, 0].astype(float)
+        cumulative_percentage_mu = tx_field_points["Index"].to_numpy(dtype=float)
```
sjswerdloff, line 337: If this is from TxFieldPoints You need to be careful to renormalize against the last  value.  Some conventional plans can use 1.0 as the final cumulative meterset weight.  While you aren't trying to deal with Particle Therapy those can be absolute meterset or something rather arbitrary. This is probably OK for ELEKTA linacs and Monaco as the TPS, but formally, it's not guaranteed to be %

sjswerdloff (approved):
Consider renormalizing TxFieldPoints [index] column contents

dg1an3 (approved):

lib/pymedphys/_mosaiq/delivery.py
SimonBiggs: Heya @msobolewski,

Can I give a shout-out to the value of peer review (and the value of having @sjswerdloff on board) for this one. Every time I've seen this that last point has always been 1.0, so I assumed that is how it was. However, potentially a Monaco upgrade (maybe) or a different TPS, would have completely thrown that assumption asunder and we would have been reporting total MU/cumulative MU as way off.

---

Thanks @sjswerdloff for this 🙂, massively appreciated.

SimonBiggs: Fixed in 4da5a90a4dbd3e6d58e2660dfb7b2beeb0fa76a2

PR #1502, opened 2021-04-21
[STACK 1] Investigate plausibility of iso-cube based WLutz

SimonBiggs:
Project goal
------------

Make it so that WLutz can run every morning utilising the "warm-up" beams of the Linac. The workflow adjustment means that morning warm-up + daily QA doesn't actually take any longer, but every day, for every gantry angle, for every beam the field centre to kv based couch shift location deviation is being recorded.

Timeline
--------

Enough time so that I am not making any commits during my last week (last two weeks?) of work. I finish up on the 18/5.

This PR
-------

Creates tools to readily investigate how the WLutz algorithm needs to be adjusted to detect an air cavity within the daily kV QA cube instead of a BB.

The below zip contains three images extracted from an arc delivered to this cube, one from each of our clinical energies (6 MV, 10 MV, and 6 MV FFF):

> [img.zip](https://github.com/pymedphys/pymedphys/files/6354850/img.zip)

By running `poetry run pymedphys gui` within this branch and then going to the new iView Image WLutz app (<http://localhost:8501/?app=iview-image-wlutz>), the `.jpg` files within the above zip can be dragged and dropped into the app to observe the algorithm in action.

Up Next
-------

Implement an option within the current WLutz Arc tool to utilise these algorithm adjustments.

### DIFF
--- a/lib/pymedphys/_experimental/streamlit/apps/iview_image_viewer.py
+++ b/lib/pymedphys/_experimental/streamlit/apps/iview_image_viewer.py
@@ -1,0 +1,63 @@
+# Copyright (C) 2021 Cancer Care Associates
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from pymedphys._imports import libjpeg, plt
+from pymedphys._imports import streamlit as st
+
+from pymedphys._streamlit import categories
+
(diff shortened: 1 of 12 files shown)


sjswerdloff (approved):
Might want to mention what that rare encoding is, in case that changes or if someone wants to try to do the same elsewhere.  Not needed if it's documented in the transform function itself...

SimonBiggs:
> Might want to mention what that rare encoding is, in case that changes or if someone wants to try to do the same elsewhere. Not needed if it's documented in the transform function itself...

Good call. It's the 1993 Lossless JPEG encoding that never made it into the standard libjpeg libraries.

lib/pymedphys/_experimental/streamlit/apps/iview_image_viewer.py
```diff
@@ -0,0 +1,62 @@
...
+
+CATEGORY = categories.ALPHA
+TITLE = "iView Image Viewer"
+
+
+def main():
+    """A viewer for the back-end iView `.jpg` files.
+
+    The backend iView `.jpg` files are encoded in a format that is
+    rarely supported within modern software. As such, opening these
```
SimonBiggs, line 31:
```suggestion
    rarely supported within modern software (1993 Lossless JPEG). 
    As such, opening these
```

PR #1516, opened 2021-04-30
Update the site-specific centre config

SimonBiggs:


sjswerdloff (commented):
Other than failing the cypress test this looks fine and doesn't affect any others.
But it is best to have an explanation for the failure before approving...

SimonBiggs:
> Other than failing the cypress test this looks fine and doesn't affect any others.
> But it is best to have an explanation for the failure before approving...

We actually didn't used to test Cypress on Windows. I added it in because I noticed it failed sometimes even though it passed on Ubuntu. I don't know why it passed on Ubuntu but not Windows. ...having the test just shows us now that it is failing.

...so that's as far as I know at the moment... Also not sure how much I will investigate it ...

sjswerdloff:
Please create issue for investigating cypress on windows failure

sjswerdloff (approved):

PR #1541, opened 2021-08-06
Fix for pydicom update

SimonBiggs:
See https://github.com/pydicom/pydicom/pull/1458#issuecomment-893891181 and https://github.com/pydicom/pydicom/pull/1462#issue-705102054


sjswerdloff (approved):

sjswerdloff:
Just to confirm...
Pydicom added validation of some sort involving pixel data and the DICOM element Bits Stored that happens automatically (perhaps when adding the pixel data itself?) but some of the PyMedPhys tests were not including the Bits Stored element.

So no change to the PyMedPhys library code itself...
But if a user of PyMedPhys is constructing an object that has pixel data but not the Bits Stored element, they will see a validation error if they have the most recent pydicom?

And Bits Stored is presumably type 1 (mandatory and must have a value)...

SimonBiggs:
> Just to confirm...

Spot on

PR #1544, opened 2021-08-09
Run updates

SimonBiggs:
Need a "with updates" version of the Mosaiq DB tests so that we have pre-warning of DB tests failing due to dependencies.


SimonBiggs:
Superseded by https://github.com/pymedphys/pymedphys/pull/1547

PR #1547, opened 2021-08-18
Update pywin32 and limit upgrade of pymsssql

SimonBiggs:
When updating all of the dependencies there was an issue with pymssql:

https://github.com/pymedphys/pymedphys/runs/3276441619

In the pymssql changelog for the most recent patch release (2.2.1 -> 2.2.2) the underlying database library changed:

https://github.com/pymssql/pymssql/blob/master/ChangeLog.rst#version-222---2021-07-24---mikhail-terekhov

I suspect this is the cause for the database CI failures. For now I am pinning the pymssql dependency.

There is also a CVE issue with pywin32, so I am pinning that to be >=301.

---

Due to the tests failing on the latest pymssql release I would like to create a pymedphys release once this merges with this pinning so that new pymedphys installs don't break when utilising the mosaiq module. ...gee I am so grateful for the Mosaiq testing suite. Thank you @dg1an3!

---

I am tracking the CI updates failure over at https://github.com/pymedphys/pymedphys/issues/1549. That can be addressed separately from this PR.

---

@pchlap, just to help with the readability here. The only "human changes" were the version pinnings within `pyproject.toml`. Every other line of code change occurred when running `poetry run pymedphys dev propagate` afterwards.


pchlap (approved):
Thanks for the explanation @SimonBiggs, all seems fine to me. I don't have a thorough understanding of how poetry works through to be honest. So probably time for me to change that!

SimonBiggs:
Thanks @pchlap :slightly_smiling_face:

Issue #1549, opened 2021-08-18
Running poetry update gets attribute error


SimonBiggs:
When running poetry update I get the following error:

```
  AttributeError

  'EmptyConstraint' object has no attribute 'allows'

  at ~/.poetry/lib/poetry/_vendor/py3.9/poetry/core/version/markers.py:279 in validate
      275│ 
      276│         if self._name not in environment:
      277│             return True
      278│ 
    → 279│         return self._constraint.allows(self._parser(environment[self._name]))
      280│ 
      281│     def without_extras(self):
      282│         return self.exclude("extra")
      283│ 
```

I haven't been able to get to the bottom of what is causing the issue.

PR #1552, opened 2021-08-23
Update to Poetry 1.1.8 + run dep updates

SimonBiggs:
Fix for the following:

https://github.com/pymedphys/pymedphys/issues/1549

As well as running a dep update in preparation for a pymedphys release

---
## Update

In the end, I removed the pymssql pinning, it turns out is was sqlalchemy that was causing the pain. So sqlalchemy has been pinned to be < 1.4:

https://github.com/pymedphys/pymedphys/blob/6b5618c7064a3af451e5525f1b345a85848e960a/pyproject.toml#L57-L58

sqlalchemy is primarily used to support the testing suite, so, given the tests still pass with pymssql upgraded, creating a release is not so urgent.


sjswerdloff (approved):

SimonBiggs:
Thanks @sjswerdloff :slightly_smiling_face:

PR #1553, opened 2021-09-01
Update install docs

SimonBiggs:
After running through a PyMedPhys install with someone I have improved the getting started docs.


lib/pymedphys/docs/lib/howto/get-started.rst
```diff
@@ -6,21 +6,61 @@ Install Python
...
+
+If python isn't already installed this will open the Windows store. At this
+point you can click the "Get" button. It will ask you to sign in, but you can
+skip the sign in step. It will still install.
+
+Once installed, open a new command prompt and type:
+
+.. code:: bash
+
+    python
```
Matthew-Jennings, line 29: No biggie, but could instead suggest a command that doesn't require a subsequent exit? Something like `python --version`?

lib/pymedphys/docs/lib/howto/get-started.rst
```diff
@@ -33,31 +73,31 @@ following in a terminal or command prompt:
 
     pip install pymedphys[user]
 
+This command can be copied into a command prompt on Windows with a "Right Click"
+of the mouse.
```
Matthew-Jennings, line 77: Maybe make clear the right clicking = pasting? E.g.

"You can copy this command directly into a Windows command prompt. You can _Right Click_ to paste."

SimonBiggs:
Yup, I like those suggestions.

lib/pymedphys/docs/lib/howto/get-started.rst
```diff
@@ -33,31 +73,31 @@ following in a terminal or command prompt:
...
+
 You may need to open and close your terminal if you have only just installed
 Python. The ``[user]`` option is needed to install pymedphys with its
 "batteries included" so-to-speak. It will go and install a range of
 dependencies which you may need during your use of pymedphys.
 
+SSL Issues
```
pchlap, line 78: Within our network, we not only have SSL issues but also need to specify a proxy server for pip to use (I'm guessing this is common on health networks). This can be done via the command line, but hard to provide specific guidance because the user would need to know the proxy server in their network. Best to just leave it out here? Or if you like I could prepare a snippet to put in here with the command and explaining they need to find out the proxy server.

lib/pymedphys/docs/lib/howto/get-started.rst
```diff
@@ -6,21 +6,61 @@ Install Python
...
-    When installing Anaconda make sure to install it for your user only, and
-    tick the option "add to path".
+Open up a command prompt and type:
 
-.. image:: ../../img/add_anaconda_to_path.png
+.. code:: bash
+
+    python
+
+If python isn't already installed this will open the Windows store. At this
```
pchlap, line 21: I'm curious, does this way of installing python require admin rights? Just thinking that for our clinical physicists this could be the first stumbling block...

pchlap (approved):
This is a great concise simple guide for getting started.

lib/pymedphys/docs/lib/howto/get-started.rst
SimonBiggs: > Or if you like I could prepare a snippet to put in here with the command and explaining they need to find out the proxy server.

@pchlap that would be amazing. Having an example here that someone can use to fill in the blanks for their own system would be significantly better than having no guidance at all.

pchlap: Will do, I'll add it directly to the branch of this pull request? Or rather a fresh branch?

lib/pymedphys/docs/lib/howto/get-started.rst
SimonBiggs: I'm actually not sure... I suspect updating the path length beyond 260 would certainly need admin rights... and this Windows Store method installs Python in a "very nested" directory. So much so that Windows store method with a 260 character path limit means that some of pymedphys' dependencies can't install.

I do notice though that this max path limit is recommended within the Python install docs:

https://docs.python.org/3/using/windows.html#removing-the-max-path-limitation

lib/pymedphys/docs/lib/howto/get-started.rst
SimonBiggs: Implemented :slightly_smiling_face:

lib/pymedphys/docs/lib/howto/get-started.rst
SimonBiggs: Implemented :slightly_smiling_face:

lib/pymedphys/docs/lib/howto/get-started.rst
SimonBiggs: If it's just a small addition happy to have it included directly into this PR. But really whichever you prefer.

pchlap: Cool, I've popped something into this PR. Hopefully that's suitable?

lib/pymedphys/docs/lib/howto/get-started.rst
pchlap: Yep, fair enough. Let's leave it for now. But I'll try to run through this at some point with one of our clinical physicists and see how far they get. Perhaps we'll just have to add at the top that this guide assumes they are working on a device where they have admin rights.

SimonBiggs: > But I'll try to run through this at some point with one of our clinical physicists and see how far they get

That'd be amazing

> Perhaps we'll just have to add at the top that this guide assumes they are working on a device where they have admin rights.

I think if there is a simple approach that doesn't need admin rights we should opt to go that way. It certainly is possible (I didn't have admin rights back at RCCC, I did get the path limit bumped by IT though).

lib/pymedphys/docs/lib/howto/get-started.rst
```diff
@@ -95,3 +95,21 @@ pypi.org and files.pythonhosted.org:
...
+
+The following command specifies the proxy server for pip. Ensure you insert
+your username and password used to authenticate on the proxy server, along with
+the host and port of the proxy server. If you are unsure of the host and port
+to use in your environment, reach out to a network administrator to obtain
+these:
+
+.. code:: bash
+
+    pip install --proxy=http://username:password@host:port --trusted-host pypi.org --trusted-host files.pythonhosted.org pymedphys[user]
```
SimonBiggs, line 115: Yup, perfect. Thanks @pchlap.

SimonBiggs:
@Matthew-Jennings given you left a comment I'll wait until you leave an approval before merging. :slightly_smiling_face:, but I think this is now ready for merge.

Matthew-Jennings (approved):
LGTM

lib/pymedphys/docs/lib/howto/get-started.rst
SimonBiggs: Hi @pchlap,

When you do get someone to run through it, it appears the Windows 10 store install approach doesn't add python scripts to the path. So the pymedphys CLI won't be available. PyMedPhys can still be called by running `python -m pymedphys --help`. 

It might be worth us having a "test to make sure the pymedphys CLI is working" section, and if it's not, know that you can use `python -m pymedphys etc...` or you can use pymedphys directly by adding Scripts to the path (provide instructions on how to do this).

Issue #1556, opened 2021-10-08
Building a PyOxidizer binary install of PyMedPhys


SimonBiggs:
I plan to create an executable for the PyMedPhys CLI (which also includes the GUI). This will install without admin privileges on Windows. I will also create an install for MacOS and Linux. I have the option to manually sign the release executable with my Radiotherapy AI enterprise license declaring that I have overseen that particular build and guaranteeing that it isn't modified before installation after the signing.

When this executable is installed an option is provided to automatically have PyMedPhys added to the PATH.

I will start with Windows, and I will detail the key steps within this issue so that others can follow along and replicate if need be. Of note, this is essentially bundling a Streamlit application into a standalone executable, so those who are trying to do this, following my steps here may be helpful. This in particular has been requested over in the [Streamlit forums](https://discuss.streamlit.io/t/using-pyinstaller-or-similar-to-create-an-executable/902/32?u=simonbiggs).

I am going to detail here the process I go through (warts and all). I figure it might be beneficial for someone to have a read through and see my process. This is heavily inspired by how @simonw documents his work on datasette-app, which helped me through some issues recently:

https://github.com/simonw/datasette-app/issues/22

Also, if one of my desires for PyMedPhys is that it becomes a way that people can learn how to code, it helps to be able to read someone struggling through these concepts.

---

I'm going to label this issue as "teaching", and this is the approach I will take. My primary audience is people who are wanting to learn, so I am going to step into "teacher mode", and write from that frame of reference. I recognise that writing out an issue like this will cost a reasonable amount of time. It is my hope that others can benefit from it. Please let me know if you do find this beneficial as whether or not I flesh out more code contributions like this will be directly dependent on that feedback.

---

To see this latest code from this work, see https://github.com/pymedphys/pymedphys/pull/1557

SimonBiggs:
Created an initial commit for the pyoxidizer build file:

https://github.com/pymedphys/pymedphys/blob/87ee7b6e0bdbc67f2c93e15345b14338f8c5ea90/pyoxidizer.bzl#L1-L78

Also added PyOxidizer as a dev dependency:

https://github.com/pymedphys/pymedphys/blob/4f198f1d8944a4dcfeee9bb8daf63e565d981499/pyproject.toml#L152

On a first pass I got the following error:

```
error[PYOXIDIZER_PYTHON_EXECUTABLE]: adding PythonExecutable to FileManifest

Caused by:
    0: building Python executable
    1: building executable with Rust project
[...]

PR #1557, opened 2021-10-08
Working on PyOxidizer binary

SimonBiggs:
See https://github.com/pymedphys/pymedphys/issues/1556 for details. (Fixes https://github.com/pymedphys/pymedphys/issues/1556)


SimonBiggs:
Hi @sjswerdloff, @Matthew-Jennings, and @pchlap,

I detailed what this PR is doing over at https://github.com/pymedphys/pymedphys/issues/1556. In the end I hit a bit of a dead end. But in the process an Icom listener test was written and a few refactoring occurred, as well as some initial steps for creating a PyMedPhys binary utilising PyOxidizer (as opposed to the previous approach).

Nothing core in here, but figured there is still enough going on that it is worth a review before merge.

lib/pymedphys/_icom/patients.py
```diff
@@ -126,6 +126,10 @@ def update_data(self, ip, data):
 
                 return
 
+            # Every iCOM data item has an index, this check here is
+            # making sure that the currently received data item is
```
sjswerdloff, line 130: Is monotonic increasing sufficient or is exact increment by one necessary?

lib/pymedphys/tests/icom/test_icom_cli.py
```diff
@@ -0,0 +1,64 @@
...
+from pymedphys._root import LIBRARY_ROOT
+
+
+def test_icom_cli():
+    icom_server_process = multiprocessing.Process(target=_mock_icom_server)
+    icom_server_process.start()
+
+    env = os.environ.copy()
+    if getattr(sys, "frozen", False):
+        env["PATH"] = str(LIBRARY_ROOT.parents[1]) + os.pathsep + env["PATH"]
```
sjswerdloff, line 34: It would be helpful to explain what this parents[1] is pointing to and why.  I get the gist of it, but...

pyoxidizer.bzl
```diff
@@ -0,0 +1,84 @@
...
+
+    files.install("dist", replace = True)
+
+    return files
+
+def make_exe_installer(exe):
+    files = FileManifest()
+    files.add_python_resource(".", exe)
+
+    installer = WiXInstaller("pymedphys", "PyMedPhys-0.38.0-dev4-Setup.msi")
```
sjswerdloff, line 63: Any way to avoid hardcoded value?

sjswerdloff (approved):

lib/pymedphys/_icom/patients.py
SimonBiggs: I'm not sure... I don't have access to a Linac anymore... But, this listener, with the monotonic increasing check, has been humming along without issue for ~2 years.

lib/pymedphys/tests/icom/test_icom_cli.py
SimonBiggs: Yup, I completely agree. That could be much clearer. Made it explicit in:

https://github.com/pymedphys/pymedphys/pull/1557/commits/ae3309c44cadd52d0e20a6422c2f9a09ba29f161

Thanks @sjswerdloff :slightly_smiling_face:

pyoxidizer.bzl
SimonBiggs: I will need to think of something. This file isn't actually using Python, instead it is using [Skylark](https://docs.bazel.build/versions/main/skylark/language.html). I'll need to learn how to import files etc.

I won't fix the hard coding in this pass. But for maintainability it does need to be one of those things that is fixed. Worst case scenario... I could write it in with the `pymedphys dev propagate` call. But... more ideally... skylark could load from `pyproject.toml`... ... but I suspect "import toml" is out of the question.

Anyway... a can to be kicked down the road a bit.

There is talk of changing this config to Python proper:
https://github.com/indygreg/PyOxidizer/issues/444

Potentially if I wait long enough this hard-coding problem might be able to be solved for us by that :slightly_smiling_face:.

PR #1558, opened 2021-10-09
[RFR] Fix pyright false negatives

SimonBiggs:


SimonBiggs:
Thanks @sjswerdloff :)

PR #1559, opened 2021-10-10
Python 3.10 and updating dependencies

SimonBiggs:
Still awaiting wheels (as to be expected)


SimonBiggs:
I learnt something from:

https://stackoverflow.com/a/65589331/3912576

One can explicitly set poetry python versions by calling:

```
poetry env use /full/path/to/python
```

See the docs:

https://python-poetry.org/docs/managing-environments/#switching-between-environments

I use [asdf](https://github.com/asdf-vm/asdf) to dynamically manage the versions of all the programming languages I use. So, to achieve this temporary swap to Python 3.10.0 locally for testing this PR I did:

```
asdf install python 3.10.0
asdf global python 3.10.0
```

```
$ which python
/home/simon/.asdf/shims/python
```

```
poetry env use /home/simon/.asdf/shims/python
```

Matthew-Jennings:
I've been exploring this myself today.

So I think `scipy 1.6.1` might be incompatible with python 3.10. `scipy` does have later versions that are explicitly stated to be python 3.10 compatible. I just can't work out where `scipy` is being pinned to `1.6.1`. I've tried `pipdeptree` to investigate further but cannot find a dependency that is applying an upper bound to `scipy`. The only clue I have is that the `requirements-**.txt` files in `pymedphys` pin to `scipy=1.6.1`, but I think this is a "result" rather than a "cause".

@SimonBiggs, any idea what I'm missing?

SimonBiggs:
> @SimonBiggs, any idea what I'm missing?

Yup, do a ctrl+f for scipy within the poetry.lock file and it will tell you exactly which dependency has the upper bound :)

SimonBiggs:
Also, there may not be one. It may be able to be updated just by running `poetry update`

Matthew-Jennings:
I can't find a pin in `poetry.lock`, but I do see that it's set to `scipy-1.6.1`. When I run `poetry update`, it doesn't update to past `1.6.1`, but if I run a `pip install scipy` in the `venv`, it does collect `scipy=1.8.1`!

Not sure what's going on... I had a cheeky go at updating the lock file manually, but it looks tedious... I'd have to update all the wheels with their hashes, and don't want to go to the effort if it's not going to work (e.g. something is still invisibly - to me at least - forcing `scipy=1.6.1`).

Still not super comfortable with `poetry`, so not sure if my understanding is correct.

SimonBiggs:
> I had a cheeky go at updating the lock file manually, but it looks tedious...

Yeah... That lock file is not designed for manual editing...

Matthew-Jennings:
> > I had a cheeky go at updating the lock file manually, but it looks tedious...
> 
> Yeah... That lock file is not designed for manual editing...

Do you know of any other leads to investigate why poetry won't update past `scipy=1.6.1`, beyond:

- `poetry.lock`
- `pyprojecty.toml`, and
- Google search for bugs

?

SimonBiggs:
Any of the pins could be a culprit:

https://github.com/pymedphys/pymedphys/blob/c9ff95f002e8497351b966cf130993474bc1f0bf/pyproject.toml#L73-L101

And if that version of scipy was the last one to support Python 3.7 then
the 3.7 lower bound is also a likely candidate for the issue.

Matthew-Jennings:
Removing all upper pins from `pyproject.toml` still doesn't make poetry update `scipy` when running `poetry uipdate`.

Does poetry respect `setup.py` as well?

SimonBiggs:
> Does poetry respect `setup.py` as well?

Nope, it ignores setup.py. Just toml and lock

SimonBiggs:
Hi @nlambriICH and @Matthew-Jennings if you're interested in how I went about fixing the Mosaiq DB tests, see:

https://github.com/pymedphys/pymedphys/issues/1656

Cheers :),
Simon

Matthew-Jennings:
Great job!!

PR #1565, opened 2021-10-30
An initial workaround for dataframe unable to be styled

SimonBiggs:
An initial workaround for:

https://pymedphys.discourse.group/t/daily-winston-lutz-app-error/190/2

This is until more information is able to be provided regarding the data scenario in use when this error occurred.


SimonBiggs:
This is a small change within an experimental module. For now it is short-circuiting review.

PR #1571, opened 2021-11-10
Fix pixel spacing bug

cedricbelanger:
Hello,

Here is the follow up of the discussion about possible issues in the gamma index calculations https://pymedphys.discourse.group/t/gamma-index-calculation-possible-issues/191/2.

I cloned the latest version of the repository. After investigations, the bug is in the way the pixel spacing of the rtdose file is handled. Here is an illustration of the problem and the solution. First, I used a dummy plan in Oncentra Prostate and calculated the 3D dose in the TPS. I added some reference points to help for the visualisation. Here is an illustration of the isodose lines and the points (white star markers): 
points_ocp

Next, I tried to recreate this image based on the rtdose file that was exported from Oncentra Prostate. Here is the code for it : 

```
import numpy as np
import pymedphys
import pydicom
import matplotlib.pyplot as plt

p1 = 10*np.array([-0.67, -2.375, -1.05])
p2 = 10*np.array([-0.103, -2.642, -1.05])
p3 = 10*np.array([2.012, -2.792, -1.05])
rtdose = pydicom.read_file("RTDOSE.dcm", force=True)
zyx, dose = pymedphys.dicom.zyx_and_dose_from_dataset(rtdose)
z, y, x = zyx
isodose_levels = {	
	"lime": 11.75,
	"midnightblue": 14.25,
	"red": 15.0,
	"yellow": 18.75,
	"cornflowerblue": 22.5,
	"white": 30
}
fig, ax = plt.subplots()
ax.invert_yaxis()
ax.plot(p1[0], p1[1], "w*")
ax.plot(p2[0], p2[1], "w*")
ax.plot(p3[0], p3[1], "w*")
index = np.argmin(np.abs(z - p1[2]))
ax.pcolormesh(x, y, dose[index], shading='nearest')
ax.contour(x, y, dose[index], levels=list(isodose_levels.values()), colors=list(isodose_levels.keys()))
plt.show()
```

The image produced was : 

bug

As seen in the image, there is a shift in the position of the isodose lines and the 3D dose relative to the points. By changing 
``` 
di = float(ds.PixelSpacing[0])
dj = float(ds.PixelSpacing[1])
```
with
```
di = float(ds.PixelSpacing[1})
dj = float(ds.PixelSpacing[0])
```
produced the following image : 

bug_fixed

Now the points are visually correctly placed. This is because PixelSpacing[0] is associated with the rows and PixelSpacing[1] is associated with the columns as described in the DICOM documentation https://dicom.innolitics.com/ciods/cr-image/cr-image/00280030#:~:text=Pixel%20Spacing%20Attribute%20%E2%80%93%20DICOM%20Standard%20Browser%20Physical,Section%C2%A010.7.1.3.%20Required%20if%20the%20image%20has%20been%20calibrated.

In Oncentra Prostate, the pixel spacing of the 3D dose is not cubic (i.e. x and y pixel spacing can be different), as it would be expected in general. I cannot share the DICOM file because of ethical rules at my institution, and creating a dummy plan with oncentra is not that trivial because I need to have the US image and contours, but I suggest to implement tests with pixel spacing that are not the same in x and y to ensure that this situation is handled correctly. 

Cédric


SimonBiggs:
> I cannot share the DICOM file because of ethical rules at my institution,

Might you be allowed to share an npz file of just the pixel array? And accompany it with the pixel spacing values, and other co-ordinate related values?

sjswerdloff:
> > I cannot share the DICOM file because of ethical rules at my institution,
> 
> Might you be allowed to share an npz file of just the pixel array? And accompany it with the pixel spacing values, and other co-ordinate related values?

I'll see if I can get my hands on a sample phantom RT Dose from Oncentra (no promises, so the npz file may be required).

sjswerdloff (commented):
given that there was confusion on this matter, it might be helpful to have a comment that explains what's going on, or even just a link to the specific part of the DICOM standard.
http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_10.7.html#sect_10.7.1.3

sjswerdloff:
> given that there was confusion on this matter, it might be helpful to have a comment that explains what's going on, or even just a link to the specific part of the DICOM standard.
> http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_10.7.html#sect_10.7.1.3

It seems that test_structure_dose_mask() is now failing, and I believe it may be that the same incorrect assumption/interpretation in the working code that is under repair was *also* in the test.

SimonBiggs:
> It seems that test_structure_dose_mask() is now failing, and I believe it may be that the same incorrect assumption/interpretation in the working code that is under repair was _also_ in the test.

:+1: Yup, that does seem to be the case.

---

I believe the hung windows tests are not due to this PR:

image

I can set aside ~45 mins today between 1:15 and 2 pm. I'll see if I can fix that bug in the CI suite to make this PR a bit easier.

---

Also, I think we should change the variable naming from di/dj to "row_spacing and column_spacing" to make this issue a bit clearer and be more in-line with the DICOM standard's naming.

cedricbelanger:
@SimonBiggs  I can generate the .npz file tomorrow if you still need it. I will explicit the changes with i and j also to follow the dicom standard. Regarding the np.arange, do you want to change 

```
col_range = np.arange(0, ds.Columns * di, di)
row_range = np.arange(0, ds.Rows * dj, dj)
```
with
```
col_range = np.array([i * di for i in range(ds.Columns)])
row_range = np.array([i * dj for i in range(ds.Rows)])
```
to make sure that there is no bug when the pixel spacing has a lot of decimal digits.

cedricbelanger:
@SimonBiggs 

Here are the files : 

[npy_files.zip](https://github.com/pymedphys/pymedphys/files/7517228/npy_files.zip)

Cédric

SimonBiggs:
> Regarding the `np.arange`, do you want to change

Yup, definitely. PyMedPhys almost needs a "search and replace" for np.arange at some point. I find using something like `np.array(range(...))` is a good direct swap in that doesn't have the floating point issues that `arange` has.

SimonBiggs:
> Here are the files

To write a test that mimics your plot above we'll also need the x, y, and z values for those dose points and their dose values.

Then, I reckon we create a dummy DICOM file that loads up that npz file into itself as well as the respective pixel and z values and then we use the load zyx dose function as well as [scipy.interpolate.RegularGridInterpolator](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RegularGridInterpolator.html) to create a dose interpolation of the grid. Then lastly, we input the x, y, z values of the known dose points and verify that the interpolation is sufficiently close to the known dose value.

What do you think?

cedricbelanger:
@SimonBiggs That might work. However, I have some concerns with this test. We need to quantify sufficiently what is close to the dose values, because for now I choosed dose points that are in high dose gradients regions (easier for visualization of the displacement/distortion). Hence, small displacements, even with the right coordinates, might have effect of 1-10% on the dose (from what I observed with the gamma test). The brachy plan is not ideal for this test because of these dose gradients. Furthermore, linear interpolation is not sufficient to describe the dose gradients close to the source (mainly because of the 1/r^2 + the geometry of the source which is a line). We might want to create points that are in regions with low gradients perhaps?

SimonBiggs:
I think importantly we just need to define a buffer percent that the
agreement must be within. If we made the buffer be equivalent to what you
used as a contour line on those images that will be enough to differentiate
if it is making the previous mistake or it has been fixed.

What are your thoughts?

SimonBiggs:
Or, maybe, we just verify that the interpolated dose for all those points
is larger than the value you used for that contour line? That way it's
essentially testing whether or not "it is within that contour" that you
drew on those images.

cedricbelanger:
I agree with your suggestions. I will generate the dummy RTDOSE file and get the dose values for these points. Visually it is true that these points seem to be out of the 200% isodose lines, so this can be easily be verified with what you proposed. We can also use some function to verify is the point falls within the contour or not (matplotlib have this function or shapely as well).

SimonBiggs:
> We can also use some function to verify if the points fall within the contour or not (matplotlib have this function or shapely as well).

Ahh good call. Yup, you can feel free to go down whichever approach you prefer.

cedricbelanger:
@SimonBiggs Here is the dicom file, which contains no patient information.

[RTDOSE_test.zip](https://github.com/pymedphys/pymedphys/files/7522662/RTDOSE_test.zip)

The code to run the test : 


```
import numpy as np
import pymedphys
import pydicom
import matplotlib.pyplot as plt
from matplotlib.path import Path

rtdose = pydicom.read_file("RTDOSE_test.dcm", force=True)
zyx, dose = pymedphys.dicom.zyx_and_dose_from_dataset(rtdose)
test_points = []
for p in rtdose.ROIContourSequence:
	test_points.append(p.ContourData)
test_points = np.array(test_points)
prescription = 100*rtdose.DoseGridScaling
z, y, x = zyx
index = np.argmin(np.abs(z - test_points[0][2]))
cs = plt.contour(x, y, dose[index], levels=[2*prescription], colors=['r'])
is_inside_count = 0
for p in test_points:
	for contour in cs.allsegs[0]:
		path = Path(contour)
		if path.contains_point(p[:2]):
			is_inside_count += 1
assert(is_inside_count == 3)
plt.close('all')
```

The test passes with the change, but fails with the previous version.

Cédric

SimonBiggs:
Can you upload your dose file as a PR to the following repo:

https://github.com/pymedphys/data

Then copy the download URL (including the commit hash) to the following file in this repo within this PR:

https://github.com/pymedphys/pymedphys/blob/main/lib/pymedphys/_data/urls.json

That will then allow you to use the following function which downloads the data on the fly if it isn't already available locally:

```python
import pymedphys

path_to_downloaded_file = pymedphys.data_path("the-name-you-gave-the-file-within-the-urls-file.dcm")
```

Then, that test you wrote above, if you make the following replacement:

```python
# change this
rtdose = pydicom.read_file("RTDOSE_test.dcm", force=True)

# to this
path_to_downloaded_file = pymedphys.data_path("the-name-you-gave-the-file-within-the-urls-file.dcm")
rtdose = pydicom.read_file(path_to_downloaded_file, force=True)
````

And then place that test script within a function called `test_non_square_pixels` within the following file:

https://github.com/pymedphys/pymedphys/blob/main/lib/pymedphys/tests/dicom/test_coords.py

That will then make your test run on every commit to PyMedPhys within the online CI testing :slightly_smiling_face:

File: lib/pymedphys/_dicom/coords.py [hunk 1]
```diff
@@ -132,8 +132,8 @@ def xyz_axes_from_dataset(
     di = float(ds.PixelSpacing[0])
     dj = float(ds.PixelSpacing[1])
```
SimonBiggs, line 133: Let's rename di and dj like the following:

```suggestion
    column_spacing = float(ds.PixelSpacing[1])
```

cedricbelanger: @SimonBiggs I uploaded the rtdose file in a new PR. I wait for the approval of the PR before doing the rest in this PR?

Cédric

SimonBiggs:
> I wait for the approval of the PR before doing the rest in this PR?

Yup, I've approved and merged that in now. Here's the URL:

https://github.com/pymedphys/data/raw/a9f530bcf9ceeb73fe6e1583ac060252b3ef9c96/rtdose_non_square_pixels.dcm

Also, before making any further changes to this PR can you make sure to merge in pymedphys' main branch. I have fixed some of the lingering issues with the CI suite:

```
git remote add upstream https://github.com/pymedphys/pymedphys.git
git fetch upstream main
git merge upstream/main
```

SimonBiggs:
Also, just re-highlighting @sjswerdloff comment above:

> given that there was confusion on this matter, it might be helpful to have a comment that explains what's going on, or even just a link to the specific part of the DICOM standard. http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_10.7.html#sect_10.7.1.3

Let's make sure to have a comment that includes that above link with this section of code.

cedricbelanger:
@SimonBiggs I updated the url and added the test. Not sure tho about the imports for matplotlib and the declaration of the function.

Thank you for your help,

Cédric

lib/pymedphys/tests/dicom/test_coords.py
```diff
@@ -18,6 +18,7 @@
 
 from pymedphys._imports import numpy as np
 from pymedphys._imports import pydicom, pytest
+from pymedphys._imports import matplotlib
```
SimonBiggs, line 21:
```suggestion
from pymedphys._imports import plt
```

lib/pymedphys/tests/dicom/test_coords.py
```diff
@@ -135,3 +136,24 @@ def test_coords_in_datasets_are_equal():
...
+    rtdose = pydicom.read_file(path_to_downloaded_file, force=True)
+    zyx, dose = pymedphys.dicom.zyx_and_dose_from_dataset(rtdose)
+    test_points = []
+    for p in rtdose.ROIContourSequence:
+        test_points.append(p.ContourData)
+    test_points = np.array(test_points)
+    prescription = 100*rtdose.DoseGridScaling
+    z, y, x = zyx
+    index = np.argmin(np.abs(z - test_points[0][2]))
+    cs = matplotlib.pyplot.contour(x, y, dose[index], levels=[2*prescription], colors=['r'])
```
SimonBiggs, line 151:
```suggestion
    cs = plt.contour(x, y, dose[index], levels=[2*prescription], colors=['r'])
```

lib/pymedphys/tests/dicom/test_coords.py
```diff
@@ -135,3 +136,24 @@ def test_coords_in_datasets_are_equal():
...
+    is_inside_count = 0
+    for p in test_points:
+        for contour in cs.allsegs[0]:
+            path = matplotlib.Path(contour)
+            if path.contains_point(p[:2]):
+                is_inside_count += 1
+    assert(is_inside_count == 3)
+    matplotlib.pyplot.close('all')
```
SimonBiggs, line 159:
```suggestion
    plt.close('all')
```

SimonBiggs:
Yup, awesome.



Can you wrap the section in a try finally wrapper, that way if the code fails it still closes the figures:

```python
    cs = matplotlib.pyplot.contour(x, y, dose[index], levels=[2*prescription], colors=['r'])
    is_inside_count = 0
    for p in test_points:
        for contour in cs.allsegs[0]:
            path = matplotlib.Path(contour)
            if path.contains_point(p[:2]):
                is_inside_count += 1
    assert(is_inside_count == 3)
    matplotlib.pyplot.close('all')
```

to something like:

```python
    try:
        cs = matplotlib.pyplot.contour(x, y, dose[index], levels=[2*prescription], colors=['r'])
        is_inside_count = 0
        for p in test_points:
            for contour in cs.allsegs[0]:
                path = matplotlib.Path(contour)
                if path.contains_point(p[:2]):
                    is_inside_count += 1
    finally:
        matplotlib.pyplot.close('all')

    assert(is_inside_count == 3)
```

SimonBiggs:
Also, having a bit of a think about this, sometimes tests a run in parallel with each other. There is a chance that this test could be running at the same time as another test that is also using matplotlib. The "close all" could result in some weird intermittent test failures as a result.

Would you be able to create a matplotlib figure/axis object and build the contour off of that object? And then make sure to close just that figure at the end?

SimonBiggs:
Also, with the imports, just so you can see how matplotlib is imported internally, here is where those packages are coming from:

https://github.com/pymedphys/pymedphys/blob/f95c98edbed5a8f417834eb4c455834ca1d20af4/lib/pymedphys/_imports/imports.py#L40-L46

lib/pymedphys/tests/dicom/test_coords.py
```diff
@@ -135,3 +136,24 @@ def test_coords_in_datasets_are_equal():
     ds2.ImageOrientationPatient = [1.0, 0, 0, 0, 1, 0]
     ds2.GridFrameOffsetVector = [0, -1, -2]
     assert not coords.coords_in_datasets_are_equal([ds1, ds2])
+
+def test_non_square_pixels():
```
SimonBiggs, line 140:
```suggestion
@pytest.mark.pydicom
def test_non_square_pixels():
```

cedricbelanger:
@SimonBiggs All right I will finish this tomorrow I have to go!

SimonBiggs:
Regarding the decorator, so that you can know what that is doing, the upstream pydicom library runs a subset of pymedphys tests:

https://github.com/pydicom/pydicom/blob/1b307e9e2309302113a72f9dbc9e05bc9d7a13a7/.github/workflows/master-push.yml#L151-L152

This decorator means that this test you're writing will also be run within pydicom's testing suite whenever they merge pull requests.

SimonBiggs:
> @SimonBiggs All right I will finish this tomorrow I have to go!

Completely okay :) See you :)

lib/pymedphys/_dicom/coords.py
```diff
@@ -111,6 +111,14 @@ def xyz_axes_from_dataset(
        Carlo (DOSXYZnrc and VMC++) verifications of DICOM compatible
        radiotherapy treatment plans", arXiv:1406.0014, Table 1,
        https://arxiv.org/ftp/arxiv/papers/1406/1406.0014.pdf
+
+    Extra notes
+    -----------
+    The ordering to unpac the pixel spacing values from PixelSpacing have
```
SimonBiggs, line 117:
```suggestion
    The ordering to unpack the pixel spacing values from PixelSpacing have 
```

lib/pymedphys/_dicom/coords.py
```diff
@@ -111,6 +111,14 @@ def xyz_axes_from_dataset(
        Carlo (DOSXYZnrc and VMC++) verifications of DICOM compatible
        radiotherapy treatment plans", arXiv:1406.0014, Table 1,
        https://arxiv.org/ftp/arxiv/papers/1406/1406.0014.pdf
+
+    Extra notes
+    -----------
+    The ordering to unpac the pixel spacing values from PixelSpacing have 
+    importance when dealing with non square pixels. For more informations
```
SimonBiggs, line 118:
```suggestion
    importance when dealing with non square pixels. For more information 
```

lib/pymedphys/tests/dicom/test_coords.py
```diff
@@ -135,3 +136,24 @@ def test_coords_in_datasets_are_equal():
...
+        test_points.append(p.ContourData)
+    test_points = np.array(test_points)
+    prescription = 100*rtdose.DoseGridScaling
+    z, y, x = zyx
+    index = np.argmin(np.abs(z - test_points[0][2]))
+    cs = plt.contour(x, y, dose[index], levels=[2*prescription], colors=['r'])
+    is_inside_count = 0
+    for p in test_points:
+        for contour in cs.allsegs[0]:
+            path = matplotlib.Path(contour)
```
SimonBiggs, line 155:
```suggestion
            path = matplotlib.path.Path(contour)
```

cedricbelanger:
@SimonBiggs Sorry for the delay, I comitted the new changes. I didn't sucessfully install poetry yet because of certificate certificate problems.

Cédric

SimonBiggs:
> I didn't sucessfully install poetry yet because of certificate certificate problems.

Something that can really help with certificate issues is the following package:

https://files.pythonhosted.org/packages/c1/43/916ef701ae481e04ed87af26fd54fa3906d44239f32d7f43be2fcbf54896/python_certifi_win32-1.6-py2.py3-none-any.whl

If you download that wheel file and then install it with:

```
cd path/to/where/you/downloaded/the/file
pip install python_certifi_win32-1.6-py2.py3-none-any.whl
```

That then patches pip to use your Windows Certificate Store. So if your organisation has any certificates installed there, it'll use them.

SimonBiggs:
Hi @cedricbelanger,

I am going to merge this into a non-main branch within the pymedphys repo in order for me to do the last few touch-ups before merge.

I'll ping you on the new pull request.

SimonBiggs:
Here is a link to the new PR:

https://github.com/pymedphys/pymedphys/pull/1575

PR #1572, opened 2021-11-11
Ignore hung test on Windows for now

SimonBiggs:
Ideally this fix would get to the bottom of why `pymedphys` can't be found on the path when running this test on Windows. But for now this fix is sufficient as this is already being testing on Linux.

When doing more work on the binary I'll need to remove this ignore.

### DIFF
--- a/lib/pymedphys/tests/icom/test_icom_cli.py
+++ b/lib/pymedphys/tests/icom/test_icom_cli.py
@@ -21,10 +21,15 @@ import subprocess
 import sys
 import tempfile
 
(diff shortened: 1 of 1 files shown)


SimonBiggs:
This fixes the hung test... but there are a range of tests within PyMedPhys which have deteriorated... I'm not immediately sure how to fix those. Will need to give it some thought.

sjswerdloff (approved):
As long as this change isn't somehow responsible for the errors in the various build/checks on this commit

PR #1573, opened 2021-11-12
Tests and CI fix + General Maintenance

SimonBiggs:
# Overview

So here are the items that were changed in this PR:

* Removed the comments section within the documentation
  * This was supported by extending the sphinx theme
  * My extension no longer worked with the new package versions
  * I don't currently have the bandwidth to maintain a divergence from the current docs themes
  * At some point in the future I could merge my comments addition upstream within the jupyter-book repo, and that way all version updates still will support the discourse comments

* Updated black past the previous major version which it was stuck on, and was causing a lot of weird issues
* Updated poetry to the latest version (I don't think this fixed anything, but it doesn't hurt)
* Introduced a hard 60 minute limit on all test runs so that we don't lock up all of our CI runners in 4 hour "black holes" as was happening before
* Made it so that a Python 3.7 logging patch if attempted to be applied twice simply just doesn't apply it the second time (instead of raising an error)
  * Hopefully we can remove this patch altogether soon and drop support for Python 3.7
* Updated all packages to their latest versions


SimonBiggs:
Ahh, it was an old black version... :slightly_smiling_face:, a recent upgrade must have got us stuck on the highest minor version, however there was a next major version available which didn't have the same issues.

SimonBiggs:
Arghh... running updates fixes the black issue... but breaks the docs build :/

lib/pymedphys/_vendor/patchlogging.py
```diff
@@ -46,7 +46,7 @@ def apply_logging_patch():
         )
 
     if _patch_applied:
-        raise ValueError("This patch has already been applied.")
```
sjswerdloff, line 49: This reverses behaviour and I will assume that this is intentional but I thought I would mark it just in case

sjswerdloff (approved):

lib/pymedphys/_vendor/patchlogging.py
SimonBiggs: Yup good call marking it :) yup it was intentional. I think I was overzealous on the first pass. It didn't need to raise an error, instead just not re-apply the patch.

PR #1578, opened 2021-11-24
forced save to dcm standard

crcrewso:
This forces the pinnacle converter to save every result file according to dicom standards instead of the previous attempt to maintain the source file version. This produced errors as pinnacle source files are often binary and structure sets/plan files should be ASCII

### DIFF
--- a/lib/pymedphys/_experimental/pinnacle/image.py
+++ b/lib/pymedphys/_experimental/pinnacle/image.py
@@ -226,5 +226,5 @@ def convert_image(image, export_path):
             export_path, f"{image.image['Modality']}.{imageds.SOPInstanceUID}.dcm"
         )
 
-        imageds.save_as(output_file)
+        imageds.save_as(output_file, False)
         image.logger.info("Exported: %s to %s", file, output_file)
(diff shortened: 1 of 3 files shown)


lib/pymedphys/_experimental/pinnacle/image.py
```diff
@@ -226,5 +226,5 @@ def convert_image(image, export_path):
             export_path, f"{image.image['Modality']}.{imageds.SOPInstanceUID}.dcm"
         )
 
-        imageds.save_as(output_file)
+        imageds.save_as(output_file, False)
```
SimonBiggs, line 229: Would this be able to call False as a keyword argument? Ends up being easier to read if it is quite obvious what False is being assigned to.

crcrewso: Good Idea, I'll submit the amendment

SimonBiggs (approved):
Thanks for this @crcrewso, I'm happy with this change, it makes other systems reading the DICOM files far happier :slightly_smiling_face:.

I'll let @pchlap give his approval and merge though given he's the code owner around this portion of the repository.

pchlap (approved):
Looks good to me too. Thanks a lot @crcrewso!

Issue #1170, opened 2020-11-11
Thoughts on anonymisation/pseudonymisation


Matthew-Jennings:
Hi @SimonBiggs and @sjswerdloff,

I'm sorry I wasn't able to spare much time when @sjswerdloff was making lovely changes to anonymise. I'm hoping to belatedly get up to speed, both on the backend and progress with the app. 

A relatively pertinent issue we have at our department at the moment is the constant need our therapists and admin staff have for us to anonymise DICOM files for clinical trials. I'd love to have an app ready made for them to safely use, and figure it'd be best in PyMedPhys, especially considering the progress @sjswerdloff has aleady made.

So I guess I have a few questions:

1. What is the status of GUI development? Can I help?
2. I've noticed that the `is_anonymised_xxx()` functions are in `core`. Should they be in `api`? I did intend for them to be exposed..
3. How has the default replacement strategy changed? What are your thoughts on its current status? Can it be improved? E.g. I've only very recently noticed DICOM's De-identification Method Attribute, which I don't think we set at the moment (or at least, never used to). Can we aim to claim conformance to a DICOM Application Level Confidentiality Profile?

sjswerdloff:
Hi @Matthew-Jennings 

  While the GUI is very simplistic and is limited to pseudonymisation, it is functional.
There is plenty of room for making it more sophisticated, and help is always appreciated.
I imagine it would be nice to have identifying-key exclusion available through the GUI.  Whatever is needed by real world users should drive the issues/feature request list.

  Note that I created another very simple GUI based on PyQT: https://github.com/sjswerdloff/qtpseudo.git
and I made an effort to provide a setup.py and setup_macos.py for people to generate a standalone binary for ease of deployment.  Using Simon's current naming convention, I would call it Raw, or Experimental if one wanted to be generous...

  I don't know if I would expose is_anonymised_xxx().  Whether something is anonymised is a judgement call.
I think that warrants a completely separate discussion.

  I don't think we changed the default replacement strategy for anonymise, although we did address identifiers in nested sequences.  Note that UIDs are addressed in pseudonymise, but not anonymise.

  I had not seen the De-identification Method attribute until a day or so ago.
That counts as a very useful small enhancement.  

[...]

PR #1580, opened 2021-11-28
Add pepper and epoch_jitter to pseudonymisation: See issue #1170 and initial comment from sjswerdloff on Nov 11, 2020

sjswerdloff:
incorporated pepper, which is a salt-like  addition to the hash
incorporated epoch_jitter, a number of days offset from the hardcoded epoch_start, to provide some uncertainty
regarding the date offset

first looking in the config.toml for these in a new pseudo  section (e.g.)
[pseudo]
pepper =  "stringrepresentationofrandomtext"
epoch_jitter = 999

if the section or keys are not found in the config.toml in the config_directory
suitable values are created based on OS provided pseudo-randomization routines
the top level config dict is updated, and then written to the config.toml file
Note that this inherently (within the toml package) converts literal strings to basic strings so backslashes become less pretty

### DIFF
--- a/lib/pymedphys/_experimental/pseudonymisation/__init__.py
+++ b/lib/pymedphys/_experimental/pseudonymisation/__init__.py
@@ -7,4 +7,4 @@ WARNING = """
     details.
     """
 
-warnings.warn(WARNING)
+# warnings.warn(WARNING)
--- a/lib/pymedphys/_experimental/pseudonymisation/strategy.py
+++ b/lib/pymedphys/_experimental/pseudonymisation/strategy.py
@@ -59,7 +59,7 @@ def _get_pepper():
         _config = get_config()
         pseudo_config = _config["pseudo"]
         pepper = pseudo_config["pepper"].encode("ASCII")
-    except:
+    except (KeyError, OSError) as e:
         pepper = secrets.token_urlsafe(32).encode("ASCII")
         _config_dir = get_config_dir()
         config_path = _config_dir.joinpath("config.toml")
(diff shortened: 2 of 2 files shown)


sjswerdloff:
The purpose of reading from the config.toml is so that the pepper and epoch_jitter will be consistent for the user if they pseudonymise additional data from the same patient (otherwise the pseudonymised patient name and ID will be inconsistent, and that prevents the use from accumulating  data on the same patient).
If multiple users in the same facility are pseudonymising data for the same patient, they can re-use  the values for pepper and epoch_jitter.

It is important for security purposes that the pepper and epoch_jitter be kept away from the pseudonymised data.  
Potential further improvements include:

1. securing the pepper and epoch_jitter  themselves (e.g. via encryption,  perhaps using the same technique used in pymedphys for SQL  database username/password)
2. Using a more expensive hashing technique such as bcrypt or scrypt (however, this would be at the expense of substantially longer psuedonymisation).

lib/pymedphys/_experimental/pseudonymisation/strategy.py
```diff
@@ -31,6 +34,95 @@
...
+    Look for pseudo_pepper value in the .pymedphys configuration file
+    If not present, generate a pseudorandom pepper value, attempt to save it to the config file
+    provide back either the value previously stored or just generated.
+
+    Returns
+    -------
+    Bytes representing the pepper (which were stored as ASCII)
+
+    """
+    pepper = "mynotveryprotectivepeppertest".encode("ASCII")
```
SimonBiggs, line 55:
```suggestion

```

lib/pymedphys/_experimental/pseudonymisation/__init__.py
```diff
@@ -7,4 +7,4 @@
     details.
     """
 
-warnings.warn(WARNING)
+# warnings.warn(WARNING)
```
SimonBiggs, line 10:
```suggestion

```

SimonBiggs (approved):
Just a couple of lines that I believe can be removed :slightly_smiling_face:.

Thanks for this @sjswerdloff :slightly_smiling_face: 

Happy Hannukah!

Issue #1597, opened 2021-12-13
CT Iso is being ignored (Pinnacle Conversion)


crcrewso:
https://github.com/pymedphys/pymedphys/blob/3ccfb2d2a32749e11e5fff9e1b5fd2aa03b0713c/lib/pymedphys/_experimental/pinnacle/rtstruct.py#L520

I'm in the middle of troubleshooting an issue, but I have one structure set where the ct center and the iso center are both non-zero. the structure set is offset by quite a bit and after some testing I've been able to prove to myself that the ct center value in the roi file isn't being used.

SimonBiggs:
@crcrewso would you be interested in submitting a PR for a fix? @pchlap would you be able to review this one?

crcrewso:
> @crcrewso would you be interested in submitting a PR for a fix? @pchlap would you be able to review this one?

When I get a fix of course I'll kick it back for review. Though I doubt I'm doing it the 'right' way.

SimonBiggs:
> Though I doubt I'm doing it the 'right' way.

All good, the "right way" is often far more apparent after the road has been travelled.

SimonBiggs:
> Looking over this code I'm wondering if there should be some refactoring

I would recommend that it is verified that the relevant section of code is covered by tests before any refactoring is embarked upon.

> it's hard to see the workflow

I'm all for having the code be readable as the number one priority. Keen for that to be improved.

> and there's very little class abstraction.

If by this you mean it's not very [DRY](https://media.pragprog.com/titles/tpp20/dry.pdf) then I am very keen for that to be fixed. However, if the argument is to make it more object oriented I'd like to refer to the following:

> ... unlike Java, Python does not impose object-oriented programming as the main programming paradigm. It is perfectly viable for a Python project to not be object-oriented, i.e. to use no or very few class definitions, class inheritance, or any other mechanisms that are specific to object-oriented programming languages.
>
> Moreover, as seen in the modules section, the way Python handles modules and namespaces gives the developer a natural way to ensure the encapsulation and separation of abstraction layers, both being the most common reasons to use object-orientation. Therefore, Python programmers have more latitude as to not use object-orientation, when it is not required by the business model.
>
> There are some reasons to avoid unnecessary object-orientation. Defining custom classes is useful when we want to glue some state and some functionality together. The problem, as pointed out by the discussions about functional programming, comes from the “state” part of the equation.
>
[...]

PR #1599, opened 2021-12-15
Structure set Iso Bug fixed

crcrewso:
Referenced in Issue #1597 

Strucutre sets created were not being shifted by the difference between CT center and Isocenter. This fix is tested to work against patient setups without immobilization devices. 

Known issue - if the patient center is offset from the CT center structure set may not match
I consider this improvement sufficient to upstream because currently all patients' structure sets are produced outside the patient ct volume.


SimonBiggs:
Might you be able to create a small test covering this also? I'll let @pchlap manage this PR though given he's the code owner for this part of the code base 🙂.

SimonBiggs:
You may also need to make sure you have pre-commit installed on your machine as per the overview at:

https://docs.pymedphys.com/contrib/setups/setup-win.html

crcrewso:
> Might you be able to create a small test covering this also? I'll let @pchlap manage this PR though given he's the code owner for this part of the code base slightly_smiling_face.

I'd be happy to, though I'm not sure what that test would look like? Once I get this edge case sorted I'll generate a couple anon versions of the patients that triggered me going down this rabbit hole for the project to use.

SimonBiggs:
> Once I get this edge case sorted I'll generate a couple anon versions of the patients that triggered me going down this rabbit hole for the project to use.

Yup, I think that'd be perfect. Essentially, if the input is an anonned version of the data that flagged it in the first place, then the test just points the relevant functions to that anonned data and makes sure that the relevant component of desired output is as required by the fix.

SimonBiggs:
Here's an example test that @pchlap recently wrote for another PR that is currently open:

https://github.com/pymedphys/pymedphys/blob/057a35de6ca759d612f59f3679ba83fcec5d332f/lib/pymedphys/tests/experimental/pinnacle/test_pinnacle_cli.py#L145-L176

lib/pymedphys/_experimental/pinnacle/rtstruct.py
```diff
@@ -296,7 +299,18 @@ def read_roi(ds, plan):
                     points[1] = round(points[1], 5)
                     points[2] = round(points[2], 5)
 
+                curr_points[0] = curr_points[0] \
+                    - (1*plan.ct_center[0]) \
+                    - (0*plan.iso_center[0])
```
pchlap, line 304: last part of these equations is multiplying zero * iso_center... Meaning this would always be zero and could be removed for readability?

lib/pymedphys/_experimental/pinnacle/rtstruct.py
```diff
@@ -296,7 +299,18 @@ def read_roi(ds, plan):
                     points[1] = round(points[1], 5)
                     points[2] = round(points[2], 5)
 
+                curr_points[0] = curr_points[0] \
+                    - (1*plan.ct_center[0]) \
```
pchlap, line 303: A test is failing at this point with "IndexError: list index out of range" indicating that ct_center is empty at least for the test dataset.

Would we always expect there to be a ct_center point?

crcrewso:
CT center and ISO center are user generated. Will find systemic values instead

PR #1615, opened 2022-02-09
Updating lock

pchlap:
Attempting to fix breaking update test


SimonBiggs:
Yup, it appears this was a numpy issue. Fixed now.

SimonBiggs:
Okay... not fixed. I wonder if simply upper-pinning the libjpeg might fix things?

PR #1640, opened 2022-05-27
Replace prints/writes with logging in gamma_shell

Matthew-Jennings:


Matthew-Jennings:
@SimonBiggs, do the `quiet`s need to stay for backwards compatibility? A major feature of logging in in-built suppression based on levels...

lib/pymedphys/_gamma/implementation/shell.py
```diff
@@ -329,13 +320,11 @@ def gamma_loop(options: GammaInternalFixedOptions):
...
     while distance <= options.maximum_test_distance:
-        if not options.quiet:
-            sys.stdout.write(
-                "\rCurrent distance: {0:.2f} mm | "
-                "Number of reference points remaining: {1}".format(
-                    distance, np.sum(to_be_checked)
-                )
-            )
+        logging.debug(
+            "\rCurrent distance: %.2f mm | " "Number of reference points remaining: %i",
```
SimonBiggs, line 324:
```suggestion
            "Current distance: %.2f mm | " "Number of reference points remaining: %i",
```

lib/pymedphys/_gamma/implementation/shell.py
```diff
@@ -450,13 +439,11 @@ def calculate_min_dose_difference(options, distance, to_be_checked, distance_ste
...
 
-    if not options.quiet:
-        sys.stdout.write(
-            " | Points tested per reference point: {} | RAM split count: {}".format(
-                num_points_in_shell, num_slices
-            )
-        )
-        sys.stdout.flush()
+    logging.debug(
+        " | Points tested per reference point: %i | RAM split count: %i",
```
SimonBiggs, line 443:
```suggestion
        "Points tested per reference point: %i | RAM split count: %i",
```

lib/pymedphys/_gamma/implementation/shell.py
```diff
@@ -44,7 +44,6 @@ def gamma_shell(
     skip_once_passed=False,
     random_subset=None,
     ram_available=DEFAULT_RAM,
-    quiet=False,
```
SimonBiggs, line 47: Can we deprecate this for now and not remove it just yet. We need a deprecation message that flags that it will be removed in a future version, and that informs the user how to replicate the old functionality with the new logging approach.

SimonBiggs:
> @SimonBiggs, do the `quiet`s need to stay for backwards compatibility? A major feature of logging in in-built suppression based on levels...

I am torn here (and annoyed at old me's design decisions). We can't remove the parameter without deprecation, as that will cause code that calls that parameter to break. Potentially, we can keep the parameter in the API, and have it return a deprecation message which details it will be removed in a future version?

lib/pymedphys/_gamma/implementation/shell.py
Matthew-Jennings: Sure! So just leave it as a function arg but with no functionality? Or leave in the functionality (i.e. wrap the logs in `if not quiet`)?

Matthew-Jennings:
> > @SimonBiggs, do the `quiet`s need to stay for backwards compatibility? A major feature of logging in in-built suppression based on levels...
> 
> I am torn here (and annoyed at old me's design decisions). We can't remove the parameter without deprecation, as that will cause code that calls that parameter to break. Potentially, we can keep the parameter in the API, and have it return a deprecation message which details it will be removed in a future version? (as well as instruction on setting the logging level for various 'quiet' options...)

I am very happy to defer to your preference here. I guess I'd point out that we're still strictly not at version 1. I understand that with significant use (as this module has!), a breaking change causes problems, but this warning is very clear on the PyMedPhyhs landing page. Moreover:
- it's a relatively _small_ change, and
- it's only an issue for those applications that wish to use the latest version of `pymedphys`, right? But deployed applications,. if following best practices, should be pinning. So... I dunno mate. Seems to me to be a small thing, and the benefit of innocation while we're still in beta trumps backwards compatibility in my mind.

That said, I may be missing things and I'd say it's entirely your decision. Your code!

Matthew-Jennings:
@SimonBiggs, sorry, I should do a better job of carefully reading your comments.

> I am torn here (and annoyed at old me's design decisions). We can't remove the parameter without deprecation, as that will cause code that calls that parameter to break. Potentially, we can keep the parameter in the API, and have it return a deprecation message which details it will be removed in a future version? (as well as instruction on setting the logging level for various 'quiet' options...)

That suggestion sounds reasonable. I can work on that.

SimonBiggs:
If you'd like to utilise it, we have inbuilt tooling to handle deprecation messages:

https://github.com/pymedphys/pymedphys/blob/f95c98edbed5a8f417834eb4c455834ca1d20af4/lib/pymedphys/_vendor/deprecated.py

Matthew-Jennings:
@SimonBiggs, oh, I already had a go...

To roughly explain:
- I tried setting `quiet` to `None` by default to enable checking whether the user supplied `quiet` directly. Maybe I didn't need to do this and maybe it raises other issues. Not sure...
- I emit a warning in any function/method where `quiet` is actually used (rather than passed through). Not sure if this is best either.
  - Do I need to account for scenarios where individual worker functions are called by users? Drawback is multiple instances of the same warning...

SimonBiggs:
> Drawback is multiple instances of the same warning...

If you use the warnings functionality in Python it only ever displays the warning the first time it occurs, and then ignores it subsequently.

lib/pymedphys/_gamma/implementation/shell.py
```diff
@@ -309,6 +316,15 @@ def from_user_inputs(
 
 
 def gamma_loop(options: GammaInternalFixedOptions):
+
+    # TODO: warning here too or no?
```
SimonBiggs, line 320: Ahh, I see what you mean (ignore my other comment).

We don't expose gamma loop directly via the documented pymedphys API, so, I think we can skip needing that notice here.

lib/pymedphys/_gamma/implementation/shell.py
```diff
@@ -97,8 +98,7 @@ def gamma_shell(
         The number of bytes of RAM available for use by this function. Defaults
         to 0.8 times your total RAM as determined by psutil.
     quiet : bool, optional
-        Used to quiet informational printing during function usage. Defaults to
-        False.
+        (Deprecated but maintained for now for backwards compatibility)
```
SimonBiggs, line 101: Maybe a note around what logging parameters can be passed to achieve the desired verbosity?

lib/pymedphys/_gamma/implementation/shell.py
Matthew-Jennings: So just in `gamma_shell()` then?

SimonBiggs: Yup, that's the only function here that we have exposed as a part of the PyMedPhys public API:

https://github.com/pymedphys/pymedphys/blob/ea9bfe6d6555a2a6d73da4fa6c0290dc0f644f98/lib/pymedphys/__init__.py#L6

SimonBiggs:
> ... and maybe it raises other issues. Not sure...

Doesn't make in new issues. All good, an appropriate approach. It probably wasn't needed, but certainly doesn't hurt.

lib/pymedphys/_gamma/implementation/shell.py
Matthew-Jennings: I've now added something.

lib/pymedphys/_gamma/implementation/shell.py
SimonBiggs: In the future, if there is an issue of wanting to avoid multiple warning, another approach would have been to put the warning inside a separate function, and then just call that function everywhere where relevant. I believe that would make it so that it only ends up emitting once.

But, I don't think that approach is needed here.

Matthew-Jennings: Ah yes, that makes sense.

lib/pymedphys/_gamma/implementation/shell.py
```diff
@@ -97,8 +98,12 @@ def gamma_shell(
         The number of bytes of RAM available for use by this function. Defaults
         to 0.8 times your total RAM as determined by psutil.
     quiet : bool, optional
-        Used to quiet informational printing during function usage. Defaults to
-        False.
+        Deprecated but maintained for now for backwards compatibility.
+        `gamma_shell` now utilises the `logging` module. You can set
```
SimonBiggs, line 102:
```suggestion
        `pymedphys.gamma` now utilises the `logging` module. You can set
```

SimonBiggs: And maybe add a link to the following?

https://docs.python.org/3/howto/logging.html

lib/pymedphys/_gamma/implementation/shell.py
```diff
@@ -97,8 +98,12 @@ def gamma_shell(
...
+        your desired verbosity by setting the corresponding logging
+        level. Basic information is given for the `info` level.
+        Additional information using for benchmarking or troubleshooting
+        performance is provided for the `debug` level.
```
SimonBiggs, line 106: Beautiful, thanks @Matthew-Jennings

SimonBiggs (approved):
LGTM :) last addition would be to add a line to the release changelog so that we don't forget to mention this on the next release :)

Matthew-Jennings:
> LGTM :) last addition would be to add a line to the release changelog so that we don't forget to mention this on the next release :)

In `pymedphys\lib\pymedphys\docs\release-notes.md`, right? Should I replicate the structure of 0.38.0 under "Unreleased"?

SimonBiggs:
> In `pymedphys\lib\pymedphys\docs\release-notes.md`, right? Should I replicate the structure of 0.38.0 under "Unreleased"?

Yup, that's right. Regarding the structure, up to you, but yes, put it under the heading of "Unreleased".

lib/pymedphys/docs/release-notes.md
```diff
@@ -7,6 +7,13 @@ All notable changes to this project will be documented in this file.
 This project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### New features and enhancements
+
+* The Gamma tool now utilises the `logging` module. The `quiet`parameter for
```
Matthew-Jennings, line 14: @SimonBiggs, do you think it's necessary to elaborate here? `logging` _is_ in the standard library and a decent case could be made that it's basic functionality is assumed knowledge.

SimonBiggs (changes requested):
Tests currently failing wrt to logging

SimonBiggs (approved):
LGTM :)

PR #1641, opened 2022-05-27
JOSS paper Reword as per feedback

Matthew-Jennings:


SimonBiggs (approved):

SimonBiggs:
By the way, for your reference, here is the link to the built paper after your changes:

https://github.com/pymedphys/pymedphys/actions/runs/2394647836

Matthew-Jennings:
> By the way, for your reference, here is the link to the built paper after your changes:
> 
> https://github.com/pymedphys/pymedphys/actions/runs/2394647836

Thanks! Reviewing that, I've actually noticed that the affiliations are out of order. Moreover, Elekta are represented twice (once as Pty Ltd and once as AB)

@dg1an3 and @sjswerdloff, do both of these affiliation versions need to be represented? (No worries if yes - i.e. they are strictly separate - just want to know how best to resolve the out-of-order affiliations).

PR #1645, opened 2022-05-28
Try python 3.9 & poetry 1.1.13 for library CI

Matthew-Jennings:


Matthew-Jennings:
@SimonBiggs, well, the docs built, I think - plus my local experience...

SimonBiggs (approved):

PR #1675, opened 2022-06-15
Docs reorganisation

Matthew-Jennings:
See #1665

You can see the changes made to the docs by this PR here: [https://docs.pymedphys.com/en/docs-reorg/](https://docs.pymedphys.com/en/docs-reorg/)

-------------------------------

Proposed structure (draft):

- Overview
- Users Guide
   - Getting started
      - Quick start guide
      - Using the Application
      - Using the CLI
      - Using the Library
   - How-to guides
   - Technical reference
   - Background information
- Contributors Guide
   - Getting started
      - Your first PR
      - How your PR will be reviewed
      - Joining the team!
   - Tips and tricks

PR #1688, opened 2022-06-17
Add pre-download phase before building docs

nlambriICH:
As noted in #1675 we should be pre-downloading within the docs command.

### DIFF
--- a/lib/pymedphys/_dev/docs.py
+++ b/lib/pymedphys/_dev/docs.py
@@ -56,6 +56,8 @@ def build_docs(args):
         shutil.copy(original_path, target_path)
 
     for file_name in FILES_TO_PRE_DOWNLOAD:
+        # Implemented to remove the downloading prompts from appearing
+        # within the online doc notebooks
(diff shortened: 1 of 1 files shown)


lib/pymedphys/_dev/docs.py
```diff
@@ -51,6 +55,9 @@ def build_docs(args):
     for original_path, target_path in FILE_COPY_MAPPING:
         shutil.copy(original_path, target_path)
 
+    for file_name in FILES_TO_PRE_DOWNLOAD:
```
SimonBiggs, line 58:
```suggestion
    for file_name in FILES_TO_PRE_DOWNLOAD:
        # Implemented to remove the downloading prompts from appearing
        # within the online doc notebooks
```

SimonBiggs (approved):
Thanks @nlambriICH.

I just added one commit suggestion so that our future selves can have some hints as to why that belongs there (since it doesn't appear to be doing anything...)

Issue #1686, opened 2022-06-17
Gamma speed-up howto - random subset


Matthew-Jennings:
Hi @nlambriICH and @SimonBiggs,

I've just happened to notice that the example used to demonstrate the speed-up potential of using `random-subset` is not very compelling. I.e., for a 10% sample rate (which is arguably fairly aggressive), the howto only demonstrates a 20% reduction in gamma calculation time - hardly worth it.

I think we should use a 3D example. In my previous experience, use of the default gamma parameters to compare a full 3D planned dose distribution on an 8 GB RAM workstation required many minutes of calculation time. I'm almost certain that calculating for a full factor of 10 fewer points should result in a much better time improvement. 

In the existing howto, a 2D slice is used for deonstration. It's possible that interpolation doesn't even dominate the calculation time in this scenario, and the need for speeding up gamma is largely not there.

SimonBiggs:
> required many minutes of calculation time

Were you using `max_gamma` when that happened?

Matthew-Jennings:
> > required many minutes of calculation time
> 
> Were you using `max_gamma` when that happened?

We're talking about a long time ago, but my best recollection is that both `max_gamma` and `random_subset` _independently_ improved calculation times markedly, with `random_subset` proving the more effective (which I think would make sense for dosegrids that are expected to match relatively well).

nlambriICH:
Yup, I used a 2D slice to avoid dealing with large computation times which increase really fast.

@Matthew-Jennings I'm kinda busy right now... Might you be interested to test it locally using the whole 3D dose (or a subset)?

Matthew-Jennings:
> Yup, I used a 2D slice to avoid dealing with large computation times which increase really fast.
> 
> @Matthew-Jennings I'm kinda busy right now... Might you be interested to test it locally using the whole 3D dose (or a subset)?

Sorry @nlambriICH & @SimonBiggs, I'll be travelling and out of action, development-wise, for the next two weeks. Should still be able to review though.

I don't think this is particularly urgent or anything, just a proposed improvement upon an otherwise lovely howto.

nlambriICH:
I did some testing using the 10 slices above and below the maximum dose mismatch of the 3D distribution (and shifting the evaluation dose as in the how-to).

The following are the computation times I obtained on my machine:

1. Using default gamma parameters: 9min 39s
2. Using `max_gamma=1.01`: 7min 11s
[...]

PR #1695, opened 2022-07-04
Gamma speed up: more realistic calculations

nlambriICH:
As noted in #1686, the gamma speed up how-to can be improved with more realistic computations.

Now, the examples show calculations on (small) 3D dose distributions with a pass rate of 90.9%.

On my workstation, I obtained the following computation times (all with `lower_percent_dose_cutoff=1`):

- Using the default gamma params: 12min 4s
- Using `max_gamma=1.01`: 11min 48s
- Using `random_subset` to sample 1/10th of the points: 6min 20s

I think that now the "aggressive" choice of sampling 1/10th of the points is acceptable when compared to the time gain.

### DIFF
--- a/.github/workflows/library.yml
+++ b/.github/workflows/library.yml
@@ -374,8 +374,6 @@ jobs:
 
           git status
           git diff-index --quiet HEAD --
-          REPO1: ${{ github.event.pull_request.head.repo.full_name }}
-          REPO2: ${{ github.event.pull_request.base.repo.full_name }}
 
       ## Pyright
 
--- a/.github/workflows/library.yml
+++ b/.github/workflows/library.yml
@@ -374,6 +374,8 @@ jobs:
 
           git status
           git diff-index --quiet HEAD --
+          REPO1: ${{ github.event.pull_request.head.repo.full_name }}
+          REPO2: ${{ github.event.pull_request.base.repo.full_name }}
 
       ## Pyright
 
--- a/lib/pymedphys/docs/users/howto/gamma/speed-up.ipynb
+++ b/lib/pymedphys/docs/users/howto/gamma/speed-up.ipynb
@@ -339,16 +339,6 @@
    ]
   },
   {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": []
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": []
-  },
-  {
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
@@ -379,7 +369,7 @@
    "source": [
     "Let's see how these parameters affect the gamma calculation time.\n",
     "\n",
-    "For simplicity, we restrict ourself on the z-slice with the highest disagreement.\n",
+    "For simplicity, we restrict ourself on the 10 z-slices above and below the z-slice with highest disagreement.\n",
     "\n",
     "We also add a small shift to the evaluation dose distribution to obtain a worse gamma agreement which will make the computation slower:"
    ]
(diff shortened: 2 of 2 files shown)


SimonBiggs:
It still is quite slow though... Thanks for highlighting that, and doing this @nlambriICH.

Might you have any interest in digging into the code and seeing if replacing numpy with Jax could seriously improve this?

I wonder if there would be value in having this example there, but commented out with estimated time costs within the comments? But still keep your old approach as what actually runs. At the end of the day, a tutorial that takes 20mins to run isn't ideal...

What do you think?

Also, @nlambriICH, I think it'd be worth catching up on a video call at some point, getting to know you, and getting to know what it is that you'd be keen to help make fly within PyMedPhys :).

Cheers,
Simon

SimonBiggs:
> At the end of the day, a tutorial that takes 20mins to run isn't ideal...

Actually, I see that you're way ahead of me :). You are restricting it to a subset so that the tutorial runs in a timely manner :). Great call :).

Matthew-Jennings:
> It still is quite slow though... Thanks for highlighting that, and doing this @nlambriICH.

Yep. And the choosing a random 1/10th subset only reduces the time by a factor of two? Seems unexpected. Could this be a memory issue?

SimonBiggs:
> Could this be a memory issue?

As a first pass guess, it's that we're losing out on some of numpy's array speedups.

Running a line profiler over both approaches will make it clear where the relative difference truly is though. Rule #1 of speeding up code... Don't guess, measure 🙂

Matthew-Jennings:
> Running a line profiler over both approaches will make it clear where the relative difference truly is though. Rule 1 of speeding up code... Don't guess, measure

Indeed! And we have before, but perhaps not with enough scrutiny. Pretty confident it was [the RegularGridInterpolator() calls](https://github.com/pymedphys/pymedphys/blob/17c9d93f4f9deef51003f0f86773ac506cd62693/lib/pymedphys/_gamma/implementation/shell.py#L271) that dominated the compute time.

nlambriICH:
> Might you have any interest in digging into the code and seeing if replacing numpy with Jax could seriously improve this?

Yes sure I'd be glad to do that, although it would be something completely new to me, so I agree that a team effort as suggested by @Matthew-Jennings is needed in this case.

> I think it'd be worth catching up on a video call at some point

Sure at some point we can meet virtually :)

> You are restricting it to a subset so that the tutorial runs in a timely manner :)

The computation times I reported were from the subset, so it's my workstation that is really slow...

SimonBiggs:
> I see that the notebook execution failed (possibly due to a timeout in a cell execution). 

Yup, I agree. Timeout would be the likely cause. We can increase that timeout limit, but, I suspect a tutorial/how-to should be something that runs quickly.

https://jupyterbook.org/en/stable/content/execute.html#setting-execution-timeout

> I propose we leave the how-to as is while we understand if JAX can speed up the compuatation. Do you agree? @Matthew-Jennings @SimonBiggs

Yup, I agree. Feel free to close the PR. I think this has flagged that the PyMedPhys gamma as it stands right now in many cases does take way to long to run.

SimonBiggs:
> Sure at some point we can meet virtually :)

When works best for you?

nlambriICH:
> > Sure at some point we can meet virtually :)
> 
> When works best for you?

How about this friday morning? I'm in UTC+1, I don't know what your timezone is 😄. You can send me a teams/zoom invite as you like, and see if it fits both of us

SimonBiggs:
> You can send me a teams/zoom invite as you like, and see if it fits both of us

Beautiful :slightly_smiling_face:. I have done just that. Feel free to adjust the day/time if that doesn't work for you :slightly_smiling_face:

nlambriICH:
> Beautiful 🙂. I have done just that. Feel free to adjust the day/time if that doesn't work for you 🙂

Great! 😃

PR #1710, opened 2022-07-26
Reduce dev scope

SimonBiggs:
Some dependency packages don't install on M1 macs.

So that people can still develop on an M1, scope the `dev` extras to just development packages. We can then give specific instructions on how to selectively install the extras for an M1 machine.


sjswerdloff (approved):
Other than merge conflict LGTM

Issue #1651, opened 2022-05-30
Expand DICOM anonymisation to conform to DICOM confidentiality profiles? 


Matthew-Jennings:
Is this a worthwhile endeavour? Any idea on how much the current anonymiser is used and/or how much interest there is in a python-based anonymiser that is nicely integrable into other applicaitions (e.g. AI/big-data applications)

SimonBiggs:
Potentially, a good first step to get a good idea of the value of embarking on something like this would be to add some usage statistics into PyMedPhys?

At the end of the day, we want to make sure we make maintenance decisions based on the magnitude of use of any given API, and we want to make sure that feature expansion hits places that have need. Here's an example of how streamlit informs users that they are collecting usage statistics:

https://github.com/streamlit/streamlit/blob/d44b162909fb8adcae463172c78000029e5d2fef/lib/streamlit/credentials.py#L64-L80

And it would only make sense to have the usage stats be able to be easily disabled via the config file.

crcrewso:
@Matthew-Jennings Short version, I think a dependable fixed API and CLI interface would be great if its focused only on the 'many patients' use cases. 

The long version: I had looked at putting in some real effort to use pymedphys to replace dicomcleaner, an old JRE app that opened up the can of security worms associated with maintaining a JRE install. We ended up using [Aliza-MS](https://www.aliza-dicom-viewer.com/) which has turned out so well that I felt the need to pay for the supported version just out of gratitude. For simple one-off anonymizations it works really well, and can be deployed pretty much any way you would need.

SimonBiggs:
Heya @Matthew-Jennings, happy for you to make the judgement call here. When using the only metrics we have at the moment I suspect that putting your time and work into shoring up the speed and validity of the `pymedphys.gamma` might be time better spent.

Matthew-Jennings:
> Heya @Matthew-Jennings, happy for you to make the judgement call here. When using the only metrics we have at the moment I suspect that putting your time and work into shoring up the speed and validity of the `pymedphys.gamma` might be time better spent.

Agree.

PR #1719, opened 2022-08-05
Restore electron factor API

SimonBiggs:
When this API was removed it was assumed it wasn't in use apart from one clinic. But then, in the months following a validation paper came out:

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8598145/

Given the API was in use, it is going to be re-exposed mostly as it was (with one API spelling error correction).

---

As an aside, this is an example where we should be collecting de-identified usage stats so that we can make informed maintenance decisions -- https://github.com/pymedphys/pymedphys/issues/1651#issuecomment-1140601707

### DIFF
--- a/lib/pymedphys/docs/users/ref/lib/electronfactors.rst
+++ b/lib/pymedphys/docs/users/ref/lib/electronfactors.rst
@@ -1,0 +1,28 @@
+#######################
+Electron Cutout Factors
+#######################
+
+*******
+Summary
+*******
+
+.. automodule:: pymedphys.electronfactors
+    :no-members:
+
+
+
+***
+API
+***
+
+.. autofunction:: pymedphys.electronfactors.parameterise_insert
+
+.. autofunction:: pymedphys.electronfactors.spline_model
+
+.. autofunction:: pymedphys.electronfactors.calculate_deformability
+
+.. autofunction:: pymedphys.electronfactors.spline_model_with_deformability
+
+.. autofunction:: pymedphys.electronfactors.calculate_percent_prediction_differences
+
+.. autofunction:: pymedphys.electronfactors.visual_alignment_of_equivalent_ellipse
--- a/lib/pymedphys/docs/users/ref/lib/index.rst
+++ b/lib/pymedphys/docs/users/ref/lib/index.rst
@@ -13,5 +13,6 @@ itself. Below is the documentation for a range of these libraries.
     gamma
     mosaiq
     metersetmap
+    trf
     electronfactors
     experimental/index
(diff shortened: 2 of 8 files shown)


lib/pymedphys/docs/users/ref/lib/index.rst
```diff
@@ -13,5 +13,5 @@ itself. Below is the documentation for a range of these libraries.
     gamma
     mosaiq
     metersetmap
-    trf
+    electronfactors
```
SimonBiggs, line 16:
```suggestion
    trf
    electronfactors
```

Matthew-Jennings (approved):
LGTM

> As an aside, this is an example where we should be collecting de-identified usage stats so that we can make informed maintenance decisions

That only works for use through streamlit, right? Are there additional solutions for users/app devs who programmatically call the API? New area for me.

SimonBiggs:
> That only works for use through streamlit, right?

Not at all, we could have an opt-out warning message that displays on first use of PyMedPhys that says whenever a function is utilised a counter is added to a log file. Also, if a function is run and an error is raised at the same time an error counter is incremented.

Periodically, these logs of function utilisation counts are then uploaded to a PyMedPhys store for use by maintainers to make library maintenance decisions and to provide as evidence within grant funding proposals and other similar activities.

lib/pymedphys/docs/users/ref/lib/electronfactors.rst
```diff
@@ -0,0 +1,28 @@
...
+.. automodule:: pymedphys.electronfactors
+    :no-members:
+
+
+
+***
+API
+***
+
+.. autofunction:: pymedphys.electronfactors.parameterise_insert
```
sjswerdloff, line 18: Why two different spellings, and will that impact the autofunction documentation?

SimonBiggs: Thanks Stuart :) The old spelling was a spelling mistake. So in restoring the API I took the opportunity to fix the spelling mistake in the API. But didn't do it in the docs. I will do that in a new PR.

Thank you for picking that up in review!

SimonBiggs: ...okay... so it looks like both spellings are okay. ... I will revert to the original API spelling to minimise issues.

SimonBiggs: So, in my early Sunday morning stupor, I committed straight to `main`. I need to make it so that's not possible for me to do that. Nevertheless, I fixed the issue in the following commit:

https://github.com/pymedphys/pymedphys/commit/28006b38066de913d46c3d221288d8c28edb8044

Also, that spelling mistake did raise a warning in the docs. We need to get the docs to the point where it has now warnings, and then we can have new warnings be raised as an error.

So, two process problems highlighted here, first, Simon able to commit to main, second, docs not raising errors within the CI.

Issue #1660, opened 2022-06-10
error message in gamma analysis for a feet-in DICOM


guanfada:
@SimonBiggs 
For most of cases I tested using gamma analysis, the patient's CT was head-in.
But for a special case with a tumor in the leg, the patient's CT was feet-in, and then I got the below "ascending" error. 
Any ideas about this? 
Thanks

   exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "E:/PyCharmVenv/gamma 3D TBB and TBC/001/gamma3DForDcm001.py", line 102, in <module>
    **gamma_options)
  File "E:\PyCharmVenv\venv\lib\site-packages\pymedphys\_gamma\implementation\shell.py", line 128, in gamma_shell
    quiet,
  File "E:\PyCharmVenv\venv\lib\site-packages\pymedphys\_gamma\implementation\shell.py", line 263, in from_user_inputs
    fill_value=np.inf,
  File "E:\PyCharmVenv\venv\lib\site-packages\scipy\interpolate\interpolate.py", line 2475, in __init__
    "ascending" % i)
ValueError: The points in dimension 0 must be strictly ascending

SimonBiggs:
That is interesting. It's an error message from the trilinear interpolation being utilised. But the fix would be to make it so that should a given axis be descending then within the gamma function that axis and the corresponding axis within the dose grid would just need to be flipped prior to passing it to the interpolation function.

@guanfada, might you be able to make some dummy data and create a function that loads up that dummy set and then tries to run the gamma function. We can then include that within the test suite to show the bug:

https://github.com/pymedphys/pymedphys/blob/ff7e8c54af62c85468a967f7f5abee862b0a590b/lib/pymedphys/tests/gamma/test_gamma_shell.py#L116-L127

And from there, once we have a failing test, we can take the steps required to make a fix :slightly_smiling_face:.

@Matthew-Jennings, might you be interested in providing support here?

Matthew-Jennings:
> @Matthew-Jennings, might you be interested in providing support here?

Sure!

---------------

@guanfada: Just to flesh out @SimonBiggs' request a little:

Our preferred process for a bugfix like this is to employ [Test-driven Development (TDD)](https://en.wikipedia.org/wiki/Test-driven_development). When we identify a bug, we:
   1. write a **_minimal_** test in our automated test suite that replicates the bug. 
   2. run the test suite and ensure that this new test fails
   3. write the necessary library code to fix the bug
[...]

PR #1730, opened 2022-09-04
add test feet-in-supine in test_gamma_shell.py

guanfada:
@SimonBiggs @Matthew-Jennings 
regarding issue [1660](https://github.com/pymedphys/pymedphys/issues/1660)
I added a test function to test the fee-in-supine DICOM

### DIFF
--- a/lib/pymedphys/tests/gamma/test_gamma_shell.py
+++ b/lib/pymedphys/tests/gamma/test_gamma_shell.py
@@ -20,6 +20,8 @@
 
 from pymedphys._imports import numpy as np
 
+import pydicom
+
 import pymedphys
 import pymedphys._utilities.createshells
 
@@ -216,16 +218,18 @@ def test_distance_0_gives_1_point():
 def test_if_descending_coordinate_in_DICOM():
     """Testing if the coordinate along an axis in DICOM is descending
     0020,0037  Image Orientation (Patient):
-    head-in supine,  1\0\0\0\1\0  (all ascending in DICOM)
-    head-in prone,  -1\0\0\0\-1\0 (descending x, descending y in DICOM)
-    feet-in supine, -1\0\0\0\1\0  (descending x, descending z in DICOM)
-    feet-in prone,   1\0\0\0\-1\0 (descending y, descending z in DICOM)
+    head-in supine,  1\\0\\0\\0\\1\\0  (all ascending in DICOM)
+    head-in prone,  -1\\0\\0\\0\\-1\\0 (descending x, descending y in DICOM)
+    feet-in supine, -1\\0\\0\\0\\1\\0  (descending x, descending z in DICOM)
+    feet-in prone,   1\\0\\0\\0\\-1\\0 (descending y, descending z in DICOM)
     if descending detected, flip the axis; otherwise, below error:
-    scipy\interpolate\interpolate.py", line 2475, in init "ascending" % i)
+    scipy\\interpolate\\interpolate.py", line 2475, in init "ascending" % i)
     ValueError: The points in dimension 0 must be strictly ascending
     """
     reference_filepath = pymedphys.data_path("RD.TBB_feet_in_supine.dcm")
     evaluation_filepath = pymedphys.data_path("RD.TBC_feet_in_supine.dcm")
+    # reference_filepath = pymedphys.data_path("original_dose_beam_4.dcm")
+    # evaluation_filepath = pymedphys.data_path("logfile_dose_beam_4.dcm")
 
     reference = pydicom.dcmread(reference_filepath, force=True)
     evaluation = pydicom.dcmread(evaluation_filepath, force=True)
@@ -239,11 +243,6 @@ def test_if_descending_coordinate_in_DICOM():
 
     (z_ref, y_ref, x_ref) = axes_reference
     (z_eval, y_eval, x_eval) = axes_evaluation
-
-    np.shape(z_ref)
-    np.shape(y_ref)
-    np.shape(x_ref)
-    np.shape(dose_reference)
 
     gamma_options = {
         "dose_percent_threshold": 3,
--- a/lib/pymedphys/tests/gamma/test_gamma_shell.py
+++ b/lib/pymedphys/tests/gamma/test_gamma_shell.py
@@ -227,7 +227,7 @@ def test_if_descending_coordinate_in_DICOM():
     reference_filepath = pymedphys.data_path("RD.TBB_feet_in_supine.dcm")
     evaluation_filepath = pymedphys.data_path("RD.TBC_feet_in_supine.dcm")
 
-    reference = pydicom.read_file(str(reference_filepath), force=True)
+    reference = pydicom.dcmread(reference_filepath, force=True)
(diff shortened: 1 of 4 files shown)


SimonBiggs:
@Matthew-Jennings, might you be able to take this one?

guanfada:
@Matthew-Jennings @SimonBiggs 
How can I implement the test on my local computer for this test_gamma_shell.py to generate the error (as expected)?

SimonBiggs:
@guanfada, I'll let @Matthew-Jennings take this one (spreading the maintainer load around)

lib/pymedphys/tests/gamma/test_gamma_shell.py
```diff
@@ -211,3 +211,55 @@ def test_distance_0_gives_1_point():
...
+    feet-in supine, -1\0\0\0\1\0  (descending x, descending z in DICOM)
+    feet-in prone,   1\0\0\0\-1\0 (descending y, descending z in DICOM)
+    if descending detected, flip the axis; otherwise, below error:
+    scipy\interpolate\interpolate.py", line 2475, in init "ascending" % i)
+    ValueError: The points in dimension 0 must be strictly ascending
+    """
+    reference_filepath = pymedphys.data_path("RD.TBB_feet_in_supine.dcm")
+    evaluation_filepath = pymedphys.data_path("RD.TBC_feet_in_supine.dcm")
+
+    reference = pydicom.read_file(str(reference_filepath), force=True)
```
Matthew-Jennings, line 230:
```suggestion
    reference = pydicom.dcmread(reference_filepath, force=True)
```
Preferred syntax for reading files (see [here](https://pydicom.github.io/pydicom/dev/old/transition_to_pydicom1.html?highlight=read_file#changed-function-and-variable-names) )

Also fairly sure wrapping in `str()` is not required. `pydicom.dcmread()1 handles `PathLike` objects.

lib/pymedphys/tests/gamma/test_gamma_shell.py
```diff
@@ -211,3 +211,55 @@ def test_distance_0_gives_1_point():
...
+    evaluation = pydicom.read_file(str(evaluation_filepath), force=True)
```
Matthew-Jennings, line 231:
```suggestion
    evaluation = pydicom.dcmread(evaluation_filepath, force=True)
```

lib/pymedphys/tests/gamma/test_gamma_shell.py
```diff
@@ -211,3 +211,55 @@ def test_distance_0_gives_1_point():
...
     assert len(x) == 1 & len(y) == 1 & len(z) == 1
+
+
+def test_if_descending_coordinate_in_DICOM():
+    """Testing if the coordinate along an axis in DICOM is descending
+    0020,0037  Image Orientation (Patient):
+    head-in supine,  1\0\0\0\1\0  (all ascending in DICOM)
+    head-in prone,  -1\0\0\0\-1\0 (descending x, descending y in DICOM)
+    feet-in supine, -1\0\0\0\1\0  (descending x, descending z in DICOM)
+    feet-in prone,   1\0\0\0\-1\0 (descending y, descending z in DICOM)
```
Matthew-Jennings, line 222: Is this always true? Can the `GridFrameOffsetVector` also be descending such that feet-first DICOM coordinates are ascending when read using `zyx_and_dose_from_dataset()` ?

lib/pymedphys/tests/gamma/test_gamma_shell.py
```diff
@@ -211,3 +211,55 @@ def test_distance_0_gives_1_point():
...
+        evaluation
+    )
+
+    (z_ref, y_ref, x_ref) = axes_reference
+    (z_eval, y_eval, x_eval) = axes_evaluation
+
+    np.shape(z_ref)
+    np.shape(y_ref)
+    np.shape(x_ref)
+    np.shape(dose_reference)
```
Matthew-Jennings, line 246: Prefer to remove these from final version

Matthew-Jennings:
> How can I implement the test on my local computer for this test_gamma_shell.py to generate the error (as expected)?

Assuming you've installed the dev version of pymedphys on your local workstation (i.e. using Poetry), you should be able to run tests by running `poetry run pymedphys dev test` from the command line.

lib/pymedphys/tests/gamma/test_gamma_shell.py
guanfada: removed

lib/pymedphys/tests/gamma/test_gamma_shell.py
guanfada: I don't know, but my current test could give the error
\interpolate.py", line 2475, in init
"ascending" % i)
ValueError: The points in dimension 0 must be strictly ascending

SimonBiggs:
What are the key blockers from this one merging?

guanfada:
> 

I don't remember. I just added a test function for descending coordinate in dicom file.
this caused an error as expected. 
    scipy\\interpolate\\interpolate.py", line 2475, in init "ascending" % i)
    ValueError: The points in dimension 0 must be strictly ascending

Matthew-Jennings:
Hi @guanfada.

The test suite is not actually failing due to your added test. If you interrogate the failing check in GitHub Actions, it's something to do with linting: https://github.com/pymedphys/pymedphys/actions/runs/2997012614/jobs/4808161944#step:39:31

Moreover, when I check out your branch and run tests locally, your added test does not fail. I basically think the test as written is unfinished. I'm very happy to help, but I'm a little unsure what this is intended to test. So, just to clarify:

- At some point in the past, @guanfada, you found that the gamma tool would throw a `ValueError` if the axes contained descending values. What was done after this? Was a fix introduced to flip any descending axes?
   - If so, we should discuss further - it seems to me that that's not entirely sure this is safe...)?
   - If not, perhaps we should discuss how to deal with this? My first thought is that such an error is fine to leave as is and leave it to the user to decide on any flipping. But open to discussions here.
- Any change that we do make to gamma to address this (or any change already made) ought to have a test associated demonstrates:
   - the old way fails (not critical to include in test, but nice if feasible)
   - the new way addresses the issue in the way we want it addressed (e.g. gamma calculates **correctly** with no error thrown)

I don't think the test as written is there *yet*.

guanfada:
@Matthew-Jennings 
Ah I see. I got your idea. I agree with you. Yes, it's better to leave it to users to flip the axes in dicom if in descending. 
Let's just leave this request open for a while. I would like to do some tests locally using our 3D gamma example for Head-first CT scan dicom.

guanfada:
@Matthew-Jennings @SimonBiggs 
Thanks for pointing out the right direction!
I added below lines in our 3D gamma example. Now there is no problem performing the gamma calculation for feet-in-supine dicom files. 
So, let's close this pull request without doing any changes to our code, but I will modify our 3D gamma example later so that dicom dose with any orientations can be calculated. I will put a note for users to let them know this operation is only used to enable gamma calculations, rather than modifying their dicom orientation. 

axes_reference[0].sort()
axes_reference[1].sort()
axes_reference[2].sort()
axes_evaluation[0].sort()
axes_evaluation[1].sort()
axes_evaluation[2].sort()

Issue #1731, opened 2022-09-09
Command line anonymise fails


neurolabusc:
The function anonymise_with_pseudo_cli() only creates the variable `identifying_keywords_for_pseudo` is only assigned if the user runs with the [pseudo](https://docs.pymedphys.com/en/latest/users/ref/cli/pseudonymisation.html) argument. 

Calling with pseudo succeeds:

```
pymedphys experimental dicom anonymise -p -o ~/out --pseudo ~/in
2022-09-09 12:07:04 WARNING  Keyword PatientSex has Value Representation CS and may require special processing to avoid breaking DICOM conformance or interoperability
/Users/cr/miniforge3/lib/python3.9/site-packages/pydicom/valuerep.py:290: UserWarning: The PN component length (129) exceeds the maximum allowed length of 64.
  warnings.warn(msg)
/Users/cr/in/in.dcm --> /Users/cr/out/MR.1.2.826.0.1.3680043.10.188.3003408011454037282787330082552200494_Anonymised.dcm
```

Calling without pseudo crashes:

```
$pymedphys experimental dicom anonymise -p -o ~/out ~/in         
Traceback (most recent call last):
  File "/Users/cr/miniforge3/bin/pymedphys", line 8, in <module>
    sys.exit(main())
  File "/Users/cr/miniforge3/lib/python3.9/site-packages/pymedphys/cli/__init__.py", line 148, in pymedphys_cli
    args.func(args)
  File "/Users/cr/miniforge3/lib/python3.9/site-packages/pymedphys/_experimental/pseudonymisation/__init__.py", line 116, in anonymise_with_pseudo_cli
    identifying_keywords=identifying_keywords_for_pseudo,
UnboundLocalError: local variable 'identifying_keywords_for_pseudo' referenced before assignment
```

SimonBiggs:
Hi @neurolabusc,

What if you don't use the experimental API?

```
pymedphys dicom anonymise -p -o ~/out ~/in    
```

neurolabusc:
@SimonBiggs that does seem to work. Can you [update the documentation](https://docs.pymedphys.com/en/latest/users/ref/cli/pseudonymisation.html) if this is the proper usage. That would close the issue.

SimonBiggs:
@neurolabusc, do you think a comment like the following would have helped you?

https://github.com/pymedphys/pymedphys/pull/1733/files

Cheers,
Simon

neurolabusc:
[...]

PR #1733, opened 2022-09-13
Add in comment re experimental

SimonBiggs:
See https://github.com/pymedphys/pymedphys/issues/1731

### DIFF
--- a/lib/pymedphys/cli/experimental/dicom.py
+++ b/lib/pymedphys/cli/experimental/dicom.py
@@ -12,6 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""Experimental command line DICOM tools.
+
+Provides an extra argument `--pseudo` to the `dicom anonymise` CLI.
+If you wish to utilise standard anonymisation, please instead use `pymedphys dicom anonymise`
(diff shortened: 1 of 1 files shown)


Matthew-Jennings (approved):

SimonBiggs:
@Matthew-Jennings I wonder if we need to re-nest the experimental modules, hidden, under an experimental page?

And, potentially have the "DICOM" module in the toolbar read "DICOM (anonymization, etc.)" Or something like that. As it stands looking at the toolbar, the primary attention for someone who wants to undergo anonymisation will be drawn to the experimental module first...

What are your thoughts about a slight toolbar adjustment @Matthew-Jennings?

If you think it is a good idea, is that something you might be open to doing?

Matthew-Jennings:
@SimonBiggs, the next 4-5 weeks are going to be a little crazy, but if I'll try to have a go once I need a bit of a break from it all. 

In the same vein, do you think there's value in a short summary of the major features/tools of PyMedPhys (as opposed to "modules). Perhaps this could go in either the README or the Overview (which is somewhat obsolete now).

SimonBiggs:
> the next 4-5 weeks are going to be a little crazy

Hope it's all okay. What's up if you don't mind me asking? Feel free to take it offline.

PR #1737, opened 2022-09-18
avoid crashing when --pseudo is not provided via experimental

sjswerdloff:
fallback to using release anonymisation strategy and keywords

### DIFF
--- a/lib/pymedphys/_experimental/pseudonymisation/__init__.py
+++ b/lib/pymedphys/_experimental/pseudonymisation/__init__.py
@@ -27,6 +27,7 @@ from pymedphys._dicom.anonymise import (
     get_baseline_keyword_vr_dict,
     get_default_identifying_keywords,
 )
+from pymedphys._dicom.anonymise import strategy as anon_strategy
 
 from . import strategy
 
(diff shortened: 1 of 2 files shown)


SimonBiggs:
LGTM, only addition would be to add a test that replicates the user's issue and has it fixed.

sjswerdloff:
> LGTM, only IMO addition would be to add a test that replicates the user's issue and has it fixed.

That's what I'm working on...   I had to go remember how to run tests, and I wanted to make sure I hadn't run afoul of anything else...

FYI, when I run tests locally, I'm seeing a lot of warnings  involving size of strings exceeding accepted values for some elements of particular VR (SH and PN) that are being issued by pydicom 2.x 
I don't have time to dig in to that this weekend, but I'll take a closer look when I can find the time (probably not next weekend, maybe the following).

sjswerdloff:
FYI, the failure in updates seems to be a timeout while running lint 
When I ran 
poetry run pymedphys dev lint
it hung, and then died when it finally caught a <Ctrl-Z> I tried to send it (hung my virtual machine).
This is where it hung/died:
Command '['/home/parallels/.pyenv/versions/3.9.14/envs/pymedphys39/bin/python', '-m', 'pylint', 'pymedphys', '--rcfile=/home/parallels/dev/pythonprojects/pymedphys39/pymedphys/lib/pymedphys/.pylintrc']'

sjswerdloff:
confirmed in local testing that the output of experimental anonymise vs. anonymise was identical.

SimonBiggs:
Beautiful, thanks Stuart.

> FYI, the failure in updates seems to be a timeout while running lint

Potentially the best path forward for this for now is to pin pylint to the most recently known working version. Given pylint isn't a user dependency the pin is not so bad, and it would be okay for it to hang around until we got to the bottom of the hang (or a more recent pylint version fixes it for us...)

SimonBiggs (approved):
LGTM :)

PR #1739, opened 2022-09-19
WIP: Upgrade poetry to 1.2

pchlap:
Working on upgrading poetry to 1.2.


pchlap:
Hi @SimonBiggs,

just attempting to upgrade poetry to 1.2 here. It seems the poetry installer action we use doesn't yet support poetry 1.2 (snok/install-poetry#89).

Are we happy to hold off on this until this is resolved? It also looks like we should perhaps consider shifting a few things which are currently extras into the new groups which poetry supports? Things like docs, tests, dev etc... This would require updating to all the pymedphys docs though.

Let me what you think. I don't think this should delay us in pushing out the next pending release?

pchlap:
> Hi @SimonBiggs,
> 
> just attempting to upgrade poetry to 1.2 here. It seems the poetry installer action we use doesn't yet support poetry 1.2 ([snok/install-poetry#89](https://github.com/snok/install-poetry/issues/89)).
> 
> Are we happy to hold off on this until this is resolved? It also looks like we should perhaps consider shifting a few things which are currently extras into the new groups which poetry supports? Things like docs, tests, dev etc... This would require updating to all the pymedphys docs though.
> 
> Let me what you think. I don't think this should delay us in pushing out the next pending release?

Ah, correction. This was only an issue for the mosaiq-db-tests, since they were using snok/install-poetry@v1.1.1, I change that to snok/install-poetry@v1 which is what the main library tests use and this issue seems to be resolved.

.github/workflows/library.yml
```diff
@@ -294,7 +294,7 @@ jobs:
       - name: Install a doctests environment
         if: matrix.os == 'ubuntu' && matrix.task == 'tests' || matrix.task == 'pyright'
```
sjswerdloff, line 295: I apologize if my lack of knowledge of order of precedence is the. problem here,
but is this going to result in
matrix.os == 'ubuntu' && (matrix.task == 'tests' || matrix.task == 'pyright')
or
(matrix.os == 'ubuntu' && matrix.task == 'tests') || matrix.task == 'pyright'

I assume the former is the goal...

pchlap: Thanks @sjswerdloff, I agree. I have made the change to be explicit about the order of precedence.

.github/workflows/library.yml
```diff
@@ -285,7 +285,7 @@ jobs:
       ## Install with Tests Environment
 
       - name: Install a tests environment
-        if: matrix.task == 'tests' || matrix.task == 'cypress' || matrix.task == 'slow' || matrix.task == 'pyright'
```
SimonBiggs, line 288: What was the reason for removing the tests install here?

SimonBiggs:
> It also looks like we should perhaps consider shifting a few things which are currently extras into the new groups which poetry supports?

No need to do that yet. Unfortunately, on an initial read, it looks like groups still don't support optional dependencies (which is one of the key reasons we were asking for groups in the first place [in the original issue](https://github.com/python-poetry/poetry/issues/1644#issuecomment-688256688))

pchlap:
Hi @SimonBiggs & @sjswerdloff,

so a few issues have popped up. Overall the upgrade to poetry 1.2 is fine, but there is a new way of defining groups in poetry 1.2 that are separate from extras. So I think our extras like tests and docs would now be better suited as groups. Is this something we want to change at this stage? It would require a number of updates to our documentation.

Although while writing this, I realise that one of the unique features of pymedphys is the tests are deployed with the package? In this case they should probably stay as an extra.

Another issue is that new versions of streamlit and pandas are causing issues with pylint. I have upper pinned the versions, for now, to confirm that the issues are resolved which they are. But do either of you have advice for how to best resolve such pylint issues? They are things like [`no-member`](https://github.com/pymedphys/pymedphys/actions/runs/3086609285/jobs/4991219154#step:39:10) and [`unsubscriptable-object`](https://github.com/pymedphys/pymedphys/actions/runs/3086609285/jobs/4991219154#step:39:25) which I believe are issues with pylint reporting rather than being real problems. I'm not exactly sure how this works though so if either of you could provide some insight that would be awesome.

pchlap:
> > It also looks like we should perhaps consider shifting a few things which are currently extras into the new groups which poetry supports?
> 
> No need to do that yet. Unfortunately, on an initial read, it looks like groups still don't support optional dependencies (which is one of the key reasons we were asking for groups in the first place [in the original issue](https://github.com/python-poetry/poetry/issues/1644#issuecomment-688256688))

Ah good timing! OK thanks @SimonBiggs :)

.github/workflows/library.yml
pchlap: Yeah so it seems like the new poetry will uninstall any extras which aren't specified on the install command... Therefore since tests are also specified to have doctests environment installed, I had to add the `-E tests` down there since the extras installed don't accumulate.

We could leave tests in here but that's an unnecessary step the way its currently setup for poetry 1.2... But I can look closer at the poetry install command there might be a way to keep the behaviour as before.

SimonBiggs:
> I have upper pinned the versions

Would it be possible to upper pin pylint instead of streamlit and pandas. Ideally, upper pinning dev dependencies is significantly preferable over upper pinning user dependencies...

SimonBiggs:
> They are things like [`no-member`](https://github.com/pymedphys/pymedphys/actions/runs/3086609285/jobs/4991219154#step:39:10) and [`unsubscriptable-object`](https://github.com/pymedphys/pymedphys/actions/runs/3086609285/jobs/4991219154#step:39:25) which I believe are issues with pylint reporting rather than being real problems. I'm not exactly sure how this works though so if either of you could provide some insight that would be awesome.

Yup, if it is certainly not a real issue, you are free to add an explicit ignore on that line that details the error being ignored:

https://github.com/pymedphys/pymedphys/blob/f95c98edbed5a8f417834eb4c455834ca1d20af4/lib/pymedphys/_vendor/patchlogging.py#L36

It does mean that when reading through the code the reviewer can see the ignores which is helpful.

SimonBiggs:
> Although while writing this, I realise that one of the unique features of pymedphys is the tests are deployed with the package? In this case they should probably stay as an extra.

:+1: :+1:. Yup, even as a binary distribution, one should be able to run the CLI test suite with the distributed package. By design :slightly_smiling_face:.

SimonBiggs:
Thanks for all of this work @pchlap :smile:

.github/workflows/library.yml
SimonBiggs: Oh that is painful. A benefit of splitting doctests up from tests is that for the `pyright` run through not as many dependencies needed to be installed, speeding up non-cached CI runs.

...if there isn't an obvious way to fix it, potentially just better to join "tests" and "doctests" into the one install step.

> But I can look closer at the poetry install command there might be a way to keep the behaviour as before.

This would certainly be preferred, but, don't sink too much time into it. If you do opt to change this approach can you leave a "TODO" comment to look into seeing how to make poetry not uninstall unspecified extras.

...maybe this is something where we just wait for the following to land:

https://github.com/python-poetry/poetry/issues/4842

pchlap:
> > They are things like [`no-member`](https://github.com/pymedphys/pymedphys/actions/runs/3086609285/jobs/4991219154#step:39:10) and [`unsubscriptable-object`](https://github.com/pymedphys/pymedphys/actions/runs/3086609285/jobs/4991219154#step:39:25) which I believe are issues with pylint reporting rather than being real problems. I'm not exactly sure how this works though so if either of you could provide some insight that would be awesome.
> 
> Yup, if it is certainly not a real issue, you are free to add an explicit ignore on that line that details the error being ignored:
> 
> https://github.com/pymedphys/pymedphys/blob/f95c98edbed5a8f417834eb4c455834ca1d20af4/lib/pymedphys/_vendor/patchlogging.py#L36
> 
> It does mean that when reading through the code the reviewer can see the ignores which is helpful.

Hi again @SimonBiggs, so on closer inspection it does seem that streamlit has changed their api in upgrading to 1.12.X... It's strange since nothing is listed as a breaking change in their change log.

For example, `st.bootstrap._print_url` now becomes `st.web.bootstrap._print_url` and a few more similar examples... I'm not very familiar with streamlit, perhaps you have a better insight as to how to handle this change?

Could you advise how to test the streamlit apps? I can put pylint disable in for all the other issues (there are quite a few) but I'd need to give this a proper test run first. I don't believe we have any automated tests for the streamlit stuff?

.github/workflows/library.yml
pchlap: Yeah so reading poetry the docs it seems that it's not possible to replicate the old behaviour for that install command. For now I might join the tests and doctests and leave a TODO to address this once groups as extras are supported in poetry.

SimonBiggs:
Hi @pchlap,

I'll be away for the next few days, so won't be able to get to this. Potentially @Matthew-Jennings or @sjswerdloff might be able to jump in and help?

SimonBiggs:
Am approach we could use down the track would be to not use poetry at all for this, but instead swap to using wheels and pip within the CI

.github/workflows/mosaiq-db-tests.yml
```diff
@@ -63,9 +63,9 @@ jobs:
           pip install --upgrade pip setuptools
 
       - name: Install Poetry
-        uses: snok/install-poetry@v1.1.1
+        uses: snok/install-poetry@v1
```
SimonBiggs, line 66:
```suggestion
        uses: snok/install-poetry@v1.3.1
```

It's worth pinning everything in the CI... (unless we are explicitly testing our library unpinned as in the `updates` test). We don't want things to change from passing to failing unless we explicitly change something, as much as is possible...

lib/pymedphys/_data/zenodo.py
```diff
@@ -28,7 +28,7 @@ def get_all_zenodo_file_data(record_name):
 
 @functools.lru_cache()
 def get_all_zenodo_file_data_by_id(record_id):
-    response = requests.get(f"{BASE_URL}{record_id}")
+    response = requests.get(f"{BASE_URL}{record_id}", timeout=30)
```
SimonBiggs, line 31: Can all data be downloaded in this time span? What's the need for this change to be included with the poetry upgrade PR?

lib/pymedphys/_experimental/chartchecks/helpers.py
```diff
@@ -404,6 +404,7 @@ def get_all_treatment_history_data(connection, mrn):
     )
     treatment_history = pd.DataFrame(data=table, columns=columns)
     treatment_history = treatment_history.sort_values(by=["date"])
+    # pylint: disable=unsubscriptable-object,unsupported-assignment-operation
```
SimonBiggs, line 407: Unless it is placed on the same line with some code, I believe this results in the ignores happening for the whole file. If you want it for the whole file it's best to put it right at the top of the file.

SimonBiggs:
Also, @pchlap, if you make a change, and you're keen for us to re-review, click the following to notify us:

image

lib/pymedphys/_data/zenodo.py
pchlap: This is a timeout to wait for an initial response for the server. The reason I've put this in is that pylint now warns if the timeout value isn't set.

pchlap:
> Also, @pchlap, if you make a change, and you're keen for us to re-review, click the following to notify us:
> 
> image

Thanks @SimonBiggs, I'll ping the re-review now. One more thing Im not sure of is that the requirements.txt files have all changed significantly.. Now requiring a min of Python 3.8... Do you know why this has happened?

SimonBiggs (approved):
LGTM :slightly_smiling_face:

Feel free to merge.

lib/pymedphys/_experimental/chartchecks/helpers.py
pchlap: Hmmmm based on my experiments it does seem that these affect only the intended lines. Perhaps due to the indentation? It wasn't possible to put it on the same line due to the line being too long and black moving it down anyway...

SimonBiggs:
> One more thing Im not sure of is that the requirements.txt files have all changed significantly..

Yup, when you run `poetry update` and then `propagate` all the requirements files will change. Also, potentially the new poetry version outputs requirements differently to make the changes more pronounced?

> Now requiring a min of Python 3.8... Do you know why this has happened?

It looks like some of the dependencies that we have upgraded to now only support Python 3.8 or later.

SimonBiggs:
I suspect, going forward, with our Python version usage, we should follow the same cadence as numpy:

https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table

---

Actually, better answer to the minimum of Python 3.8:

https://github.com/pymedphys/pymedphys/blob/99e578761b5c6e05676f143eefd191dac697c344/pyproject.toml#L34

We have already dropped Python 3.7 support.

lib/pymedphys/_experimental/chartchecks/helpers.py
SimonBiggs: > Hmmmm based on my experiments it does seem that these affect only the intended lines. Perhaps due to the indentation?

Okay :) I learnt something again :)

> It wasn't possible to put it on the same line due to the line being too long and black moving it down anyway...

This one is possible, when you save it (which should auto-run black), that will then split the brackets, then you just have to move the comment, save again (which should auto-run black), and repeat

But, if it only works for the line or code block in question, then, okay with me as-is.

pchlap:
Thanks a lot @SimonBiggs, I will merge this through once the LGTM analysis completes. Then I will prepare the release branch (which should be fairly straightforward given the work done in here). Hopefully I'll have the release ready by tomorrow.

Matthew-Jennings:
> Thanks a lot @SimonBiggs, I will merge this through once the LGTM analysis completes. Then I will prepare the release branch (which should be fairly straightforward given the work done in here). Hopefully I'll have the release ready by tomorrow.

Thank you so much @pchlap and @SimonBiggs for your work on this!

SimonBiggs:
Thanks @pchlap :slightly_smiling_face:

PR #1744, opened 2022-09-29
Add draft of release guide document

pchlap:
@SimonBiggs & @Matthew-Jennings, 

0.39.0 is live! 🥳 

Here is the document I have put together as I worked through this. Please let me know if you have any suggested changes/clarifications. Not rush on this though.

One issue I did notice with 0.39.0, the final sanity check requires running the tests using the pymedphys tool: `pymedphys dev tests`. Likely due to some rejigging of the extras, the advice is to install pymedphys using `pip install pymedphys[user]`. However this didn't install what I needed to run the tests, I had to run `pip install pymedphys[all]`.

In addition the `test_lossless_jpeg` test is failing for me locally (running on MacOS). Could this be due to the pylibjpeg-libjpeg version?

Let me know if we should separate this discussion out to an issue.

### DIFF
--- a/lib/pymedphys/docs/contrib/info/release-guide.md
+++ b/lib/pymedphys/docs/contrib/info/release-guide.md
@@ -106,6 +106,6 @@ Create a new release on [GitHub](https://github.com/pymedphys/pymedphys/releases
 Perform a final check to ensure the new version was released successfully. To do this, create a fresh Python virtual environment on your machine. Then install the new version of pymedphys and ensure all tests are passing as expected.
 
 ```python
-pip install pymedphys[user,dev]
+pip install pymedphys[all]
 pymedphys dev tests
 ```
--- a/lib/pymedphys/docs/contrib/info/release-guide.md
+++ b/lib/pymedphys/docs/contrib/info/release-guide.md
@@ -1,0 +1,111 @@
+# pymedphys Release Procedure
+
+## Determine next release version
+
+pymedphys uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html) in the format `MAJOR.MINOR.PATCH`, the next release number should typically be `MAJOR.MINOR+1.PATCH`. For example, if the previous release was `0.38.0`, the upcoming release will be `0.39.0`.
+
+In instances where the only changes since the last release are bug fixes and none of the pymedphys API has changed, you should increment the `PATCH` value: `MAJOR.MINOR.PATCH+1`
+
+In the remainder of this guide this version code will be referred to as `VERSION`.
+
+## Create a branch to prepare the release
+
+Create a branch named `VERSION-release-prep`:
+
+```bash
+git checkout -b VERSION-release-prep
+git push --set-upstream origin 0.39.0-release-prep
+```
+
+## Update version in pyproject.toml
+
+Update the version code near the top of the file:
+
+```toml
+[tool.poetry]
+name = "pymedphys"
+version = "VERSION"
+readme = "README.rst"
+...
+```
+
+Then run poetry update as well as propagate:
+
+```bash
+poetry update
+poetry run pymedphys propagate
+```
+
+## Update CHANGELOG
+
+Amend the `CHANGELOG.md` file to describe the changes since the last release. Insert information for this release near the top of the file and populate the sections as appropriate (remove unused sections):
+
+```markdown
+<!-- markdownlint-disable MD024 MD039 -->
+
+# Release Notes
+
+All notable changes to are documented here.
+
+This project adheres to
+[Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [VERSION]
+
+### News around this release
+
+### Breaking changes
+
+### New features and enhancements
+
+### Bug Fixes
+
+...
+```
+
+To help determine what has changed since the last release, you can inspect the [merged pull requests](https://github.com/pymedphys/pymedphys/pulls?q=is%3Apr+is%3Amerged) during that time frame.
+
+## Create release pull request
+
+Commit all changes to your release branch and push to GitHub. Then create a Pull Request of this branch into main. This will run the CI tests. Before finalising the release, ensure all tests are passing.
+
+### Troubleshooting issues
+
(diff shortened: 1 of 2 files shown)


Matthew-Jennings:
Thank you so much, @pchlap ! I've given it a first pass and I think this looks great!

Just a question (perhaps for @SimonBiggs): I notice that each version of `pymedphys` has been uploaded to Zendodo - is this automatic, or do I need to do this manually for the purposes of the JOSS review?

lib/pymedphys/docs/contrib/info/release-guide.md
```diff
@@ -0,0 +1,111 @@
...
+### Publish release on GitHub
+
+Create a new release on [GitHub](https://github.com/pymedphys/pymedphys/releases). Create a new tag in the format `vMAJOR.MINOR.PATCH` (e.g. `v0.39.0`). Enter the tag as the name for this release as well. Finally add in the change logs for this release (direct copy from CHANGELOG.md). Publish the release.
+
+### Final sanity tests
+
+Perform a final check to ensure the new version was released successfully. To do this, create a fresh Python virtual environment on your machine. Then install the new version of pymedphys and ensure all tests are passing as expected.
+
+```python
+pip install pymedphys[all]
```
SimonBiggs, line 109:
```suggestion
pip install pymedphys[user,tests]
```

SimonBiggs (approved):
This is beautiful stuff @pchlap! Thank you!

pchlap:
> Yup, by design. If you run `pip install pymedphys[user,tests]` that should be sufficient.

OK great, the only thing is the error message to the user says to simply install `pip install pymedphys[user]`, is this intended?

SimonBiggs:
> > Yup, by design. If you run `pip install pymedphys[user,tests]` that should be sufficient.
> 
> OK great, the only thing is the error message to the user says to simply install `pip install pymedphys[user]`, is this intended?

Yeah... that should cover most use cases. It potentially could be worth some further logic to detect if the user needs to install other optional dependencies also... But... if a user is "using" the pymedphys library then that install message should cover all of those use cases.

...if they are testing, then `pymedphys[user]` is not enough. 

The fix for this could be, to have a dependency list created from the requirements files that is saved into the python package via propagate, and then read out the following `modname` variable:

https://github.com/pymedphys/pymedphys/blob/778e33769ead3ed6a6b366fd816f59aa8a2d8e76/lib/pymedphys/_vendor/apipkg/__init__.py#L248-L256

And then, if that `modname` variable is within the `user` dependency list, then ask the user to run `pip install pymedphys[user]`. If it is in the tests dependency list, then ask them to run `pip install pymedphys[user,tests]`...

lib/pymedphys/docs/contrib/info/release-guide.md
```diff
@@ -0,0 +1,111 @@
...
+Then run poetry update as well as propagate:
+
+```bash
+poetry update
+poetry run pymedphys propagate
+```
+
+## Update CHANGELOG
+
+Amend the `CHANGELOG.md` file to describe the changes since the last release. Insert information for this release near the top of the file and populate the sections as appropriate (remove unused sections):
```
Matthew-Jennings, line 41: Doesn't affect what you've written here, but I think it makes sense to enforce CHANGELOG edits at the time of contribution. We can add this is a future "how to make a contribution" page

lib/pymedphys/docs/contrib/info/release-guide.md
```diff
@@ -0,0 +1,111 @@
...
+
+Commit all changes to your release branch and push to GitHub. Then create a Pull Request of this branch into main. This will run the CI tests. Before finalising the release, ensure all tests are passing.
+
+### Troubleshooting issues
+
+When preparing a release you will typically see certain steps of the CI failing. This is usually due to dependencies having been upgraded in the time since the last release and certain code within pymedphys is no longer compatible with the new versions.
+
+To confirm this is indeed the problem, you can set an upper pin for these libraries (to constrain them to a previous version). By doing this you should hopefully see the CI pass.
+
+In most instances, releasing pymedphys with an upper pin isn't appropriate, as this would be too restrictive for pymedphys users. Unfortunately there isn't a general solution to resolving these issues. You will need to remove the upper pins one-by-one and adjust the code in pymedphys to support the new version of the dependencies.
```
Matthew-Jennings, line 78:
```suggestion
In most instances, releasing pymedphys with an upper pin isn't appropriate. `pymedphys` is primarily a _library_ and upper pins introduce future obsolescnece risk. Unfortunately there isn't a general solution to resolving these issues. You will need to remove the upper pins one-by-one and adjust the code in pymedphys to support the new version of the dependencies.
```

Matthew-Jennings (approved):

lib/pymedphys/docs/contrib/info/release-guide.md
SimonBiggs: GitHub has a PR checklist option. @Matthew-Jennings, would you be open to setting that up and including a checkbox in the template for "added relevant changes to the changelog"?

lib/pymedphys/docs/contrib/info/release-guide.md
SimonBiggs: Less so obsolescence... more so... "dependency hell" for those trying to install the library in combination with other libraries that have their own restrictions.

Matthew-Jennings: Sure, "primarily" misses the point that this is a library release.

I do think upper pins technically introduce obsolescence, which at least worsens in the "future", but I can see that perhaps it's not the primary point. Feel free to suggest alternate wording. I just thought it best to give a reason.

I had in mind the point that, at some time in the future, a pymedphys dependency will likely introduce lower-pins that conflict with another (cross-)dependency's upper-pin that we've introduced. So, "obsolescence" and "future".

lib/pymedphys/docs/contrib/info/release-guide.md
Matthew-Jennings: Yep, after around 10th October, I can get stuck into that. I'm going to be slammed between now and then.

SimonBiggs:
Absolutely beautiful stuff @pchlap! A key goal for PyMedPhys has been to be able to have it be able to be built and run by us all. A release where I didn't touch it :slightly_smiling_face:! That's a first. Thank you :slightly_smiling_face: :tada:

PR #1747, opened 2022-09-30
Add ORCIDs to paper

Matthew-Jennings:


SimonBiggs (approved):
LGTM :)

Matthew-Jennings:
@SimonBiggs, any ideas why the cypress test is failing? Just in case it's relevant, the last commit was made via Github's wed editor.

I think this is ready for merge. @crcrewso and @msobolewski have confirmed they do not have ORCIDs

SimonBiggs:
> @SimonBiggs, any ideas why the cypress test is failing?

I suspect you answered your own question, given it's no longer failing?

SimonBiggs:
@rembishj is also missing his ORCID, if he wanted his on there...

Matthew-Jennings:
> @rembishj is also missing his ORCID, if he wanted his on there...

I've tried contacting @rembishj via both GitHub and LinkedIn. I think we can start a new release.

I suppose perspectives might vary on this, but I don't really think it's particularly important for an author's ORCID to be attached to a given paper. I'd say it's more important that the paper be linked to a individual's ORCID profile. The latter can be done post-publication. The former might also be possible depending on the journal's platform.

It was honestly debatable whether it was even worth performing this patch.

Matthew-Jennings:
> I suspect you answered your own question, given it's no longer failing?

I re-ran the test twice before it passed.

SimonBiggs:
> I re-ran the test twice before it passed.

That is unfortunate...

rembishj:
> > @rembishj is also missing his ORCID, if he wanted his on there...
> 
> I've tried contacting @rembishj via both GitHub and LinkedIn. I think we can start a new release.
> 
> I suppose perspectives might vary on this, but I don't really think it's particularly important for an author's ORCID to be attached to a given paper. I'd say it's more important that the paper be linked to a individual's ORCID profile. The latter can be done post-publication. The former might also be possible depending on the journal's platform.
> 
> It was honestly debatable whether it was even worth performing this patch.

Sorry, haven't been using GitHub much at my new position and missed all this. Also never saw anything via LinkedIn. 

Agreed though that I don't think the ORCID is particularly important (I need to do some digging to find mine anyway).

SimonBiggs:
If you do still want it on there, there is actually still a chance :)

PR #1750, opened 2022-10-06
0.39.1 release prep

sjswerdloff:
change reference in pyproject.toml to the pylibjpeg-libjpeg flavour rather than the pymedphys patch package
pin lower version of pylibjpeg-libjpeg to 1.3.2 (to include the Apple M1 compatibility fixes)
See https://github.com/pydicom/pylibjpeg-libjpeg for more detail


sjswerdloff:
This seems to have already gotten published to PyPI !?

sjswerdloff:
The Release notes don't look right.

sjswerdloff:
> The Release notes don't look right.

The Release notes look right in GitHub page view mode when reviewing but don't look right on:
https://docs.pymedphys.com/en/latest/release-notes.html

SimonBiggs:
@pchlap, might you be in a position to give @sjswerdloff a hand?

pchlap:
> @pchlap, might you be in a position to give @sjswerdloff a hand?

I can try...

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ All notable changes to are documented here.
 This project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.39.1]
+
+### News around this release
+
+# Release Notes
```
pchlap, line 14: @sjswerdloff  I assume this line is causing the rendering issue, level 1 heading shouldn't be here, I think removing this line probably achieves what you want?

sjswerdloff: The problem is that I wasn't aware that 0.39.1 had already been released, I didn't see a tag for it, and there was no 0.39.x branch.
So I was comparing what I had done with what was already published for 0.39.1
I'm going to work on making my dependency change in main first.
Then, I'm going to go back and see what I can do about:
tagging with v0.39.1 at the last commit that got in to the 0.39.1 release
merging those commits in to the 0.39.x branch that I created (that started at the v0.39.0 tag) and to which I already have added the dependency change.
Both the change for merge into main and the change for merge into 0.39.x will be on new Pull Requests.
I'm not going to close all of this off until I have the rest sorted.

sjswerdloff: But I make sure not to duplicate that line either...

PR #1755, opened 2022-10-06
Merge 0.39.x back into main

SimonBiggs:
(Will follow up with a change to the version number back to `0.40.0-dev1`)


SimonBiggs:
Once these tests pass I'll release to PyPI, then I'll follow up with a PR that bumps to version 0.40.0-dev1

PR #1761, opened 2022-10-16
Use EconForge's interpolator to accelerate gamma 

### LINKED ISSUE #79: Fundamentally speeding up gamma_shell
There are a few cases within `gamma_shell` that I have used [advanced indexing](https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#advanced-indexing):

image

The issue with this is that this returns a copy of numpy array.

When I mimic a constrained RAM case the code spends 10s out of a total of 30s on the following line:

```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   480      6069       6595.0      1.1      0.1          coordinates_at_distance.append(np.array(
   481      6069   10049303.0   1655.8     98.7              eval_coord[to_be_checked][None, :] +
   482      6069      85649.0     14.1      0.8              shell_coord[:, None])[:, :, None])
```

In low RAM cases this is an area where serious time improvement can be achieved by using basic indexing instead of advanced indexing.

In unconstrained RAM cases the profiling is quite different. The larges time sink is the interpolation. But that is not the only time sink, and there are definitely areas for improvement.

Of interest quite a lot of unnecessary time is spent in the following lines. I imagine if I just change the type of the original culprit I might get a bit of a speed up here.

```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   402                                               estimated_ram_needed = (np.uint64(num_points_in_shell)
   403                                                                       * np.sum(to_be_checked).astype(np.uint64)
   404                                                                       * np.uint64(32)
   405                                                                       * np.uint64(num_dimensions)
   406        13     162977.0  12536.7     18.3                              * np.uint64(2))
```

The above testing can be repeated by opening the following notebook within your own playground:

https://colab.research.google.com/drive/1cGiRhXNS7aw1qXVHYEiuf7Ozj2WPIu-X

Matthew-Jennings:
Still in draft. I expect I'll need to be more careful of API changes, etc. But I think this works! I've gotten a 2-3x speed improvement for my limited testing so far.

Haven't tested thoroughly. Will work on showcasing speed improvements later.

Closes #79


pyproject.toml
```diff
@@ -155,6 +155,7 @@ doc8 = { version = "*", optional = true }            # groups = ["dev", "all"]
 readme-renderer = { version = "*", optional = true } # groups = ["dev", "all"]
 tabulate = { version = "*", optional = true }        # groups = ["dev", "all", "doctests"]
 pyoxidizer = { version = "*", optional = true }      # groups = ["dev", "all", "build"]
+interpolation = "^2.2.4"
```
SimonBiggs, line 158: Make sure it's included as an optional dependency :)

Matthew-Jennings: sorry, why optional? I expect that, assuming this offers sufficient improvement, there'll be no need to maintain the scipy interpolator version.

Is it because the dependency is specific to gamma?

SimonBiggs: Every dependency in PyMedPhys is set to optional. That way, if someone just wants to set up a minimal environment, say an icom listener on a raspberry pi with absolute minimal dependencies, then they can just install the absolute bare minimum. Quite important when something is being installed in a sensitive location. Potentially a vendor might want to just utilise a small component of pymedphys, and they just want to get that component only, and only that small subset of dependencies only.

A massively minimal install is something that is explicitly supported by pymedphys.

Matthew-Jennings: > A **massively minimal** install

Entertaining juxtaposition there 😆 

I understood the minimal install point, I just wasn't sure/never checked how aggressively we were supporting that (e.g., I wasn't aware until looking now that *every* dependency is optional).

Matthew-Jennings:
Any ideas on why [this is failing](A massively minimal install)? Am I missing something in my attempts to add `interpolation` as a dependency?

SimonBiggs:
> Any ideas on why [this is failing](https://github.com/pymedphys/pymedphys/actions/runs/3258727302/jobs/5351038941#step:15:22)? Am I missing something in my attempts to add `interpolation` as a dependency?

Yup, you need to include it within the "imports" file:
https://github.com/pymedphys/pymedphys/blob/main/lib/pymedphys/_imports/imports.py

And then always import it from pymedphys._imports:

https://github.com/pymedphys/pymedphys/blob/main/lib/pymedphys/_imports/__init__.py

That way, the true import only actually happens at first run. Meaning, it'll only import error if the function that needs it is called.

lib/pymedphys/_gamma/implementation/shell.py
```diff
@@ -522,7 +523,33 @@ def interpolate_evaluation_dose_at_distance(
...
+        for i in range(all_points.shape[-1]):
+            grids.append(all_points[:, :, i])
+
+        points_interp = np.column_stack([np.ravel(mgrid) for mgrid in grids]).astype(
+            float
+        )
+
+        coords_evaluation_grid = interpolation.splines.CGrid(*options.axes_evaluation)
+
+        evaluation_dose = interpolation.splines.eval_linear(
```
SimonBiggs, line 548: Such a minimal change for a 3-4x speed up. Beautiful stuff @Matthew-Jennings 🙂

Matthew-Jennings:
I've been trying to construct a useful example, but I'm running into an issue here where saving my 3D RNG data to a new dataset's `PixelData` element only results in a 2D array! Only the first slice is saved to the dataset. I've tried many troubleshooting steps but am honestly at my wit's end as to why...

SimonBiggs:
> 3D RNG data to a new dataset's `PixelData`

Does it need to be a DICOM example? Potentially a simple field in field 3D dose distribution created by making a profile along each axis and then multiplying them together using the following function?

https://github.com/pymedphys/pymedphys/blob/f9b6f02b08625fcac6e1ca1893e7a7beea15f557/lib/pymedphys/_mocks/profiles.py#L31-L39

See the process for turning those profiles into higher dimensions here:

https://github.com/pymedphys/pymedphys/blob/f9b6f02b08625fcac6e1ca1893e7a7beea15f557/lib/pymedphys/_mocks/profiles.py#L49-L64

Matthew-Jennings:
> > 3D RNG data to a new dataset's `PixelData`
> 
> Does it need to be a DICOM example? Potentially a simple field in field 3D dose distribution created by making a profile along each axis and then multiplying them together using the following function?
> 
> https://github.com/pymedphys/pymedphys/blob/f9b6f02b08625fcac6e1ca1893e7a7beea15f557/lib/pymedphys/_mocks/profiles.py#L31-L39
> 
> See the process for turning those profiles into higher dimensions here:
> 
> https://github.com/pymedphys/pymedphys/blob/f9b6f02b08625fcac6e1ca1893e7a7beea15f557/lib/pymedphys/_mocks/profiles.py#L49-L64

Great point! Seems silly now that you mention it...

SimonBiggs:
@Matthew-Jennings, what's the key blocker for merging this? Is that something that would be okay being done in a separate PR?

Matthew-Jennings:
> What's the key blocker for merging this? Is that something that would be okay being done in a separate PR?

@SimonBiggs, the main thing is that I haven't yet provided a compelling demonstration of speed improvement. Though I'm quite confident of it.

I'm also mildly concerned that the test suite doesn't cover everything and that I've introduced uncaught breaking changes. That's one thing I'd like focussed on in the review (which you probably did, anyway)

Matthew-Jennings:
Oh, also should include a changelog entry

Matthew-Jennings:
And probably remove the `scipy `option entirely.

Matthew-Jennings:
A comparison of gamma calc times on my workstation:

image

lib/pymedphys/_gamma/implementation/shell.py
```diff
@@ -17,9 +17,10 @@
```
Matthew-Jennings, line 17: I haven't added myself to the copyright. It feels like too small a change for that. But it occurs to me that I should learn the rules for this

sjswerdloff: Small by what measure?
Lines of code?
Adding a thousand lines of code that doesn't accomplish much is not a huge contribution.
Changing two lines of code and improving performance by more than 2x... that's a huge contribution.

SimonBiggs: Add it in Matt 🙂

SimonBiggs: Add an extra line on the top, where it's just your copyright with the year 2022. If I do anything to this code in 2022 then I can add my name to yours 🙂

.github/workflows/library.yml
```diff
@@ -179,7 +179,7 @@ jobs:
       - name: Update PATH in Windows
         if: matrix.os == 'windows'
         run: |
-          Add-Content -Path $env:GITHUB_PATH -Value "$env:USERPROFILE\.poetry\bin"
+          Add-Content -Path $env:GITHUB_PATH -Value "$env:USERPROFILE\.local\bin"
```
SimonBiggs, line 182: Can you merge in main and see if this still works?

SimonBiggs (approved):
Beautiful stuff @Matthew-Jennings 🥳 🍾

lib/pymedphys/_gamma/implementation/shell.py
Matthew-Jennings: Cool! Thanks! Will do.

PR #1763, opened 2022-10-26
Fix poetry on Windows

SimonBiggs:
We weren't pinning our poetry install approach, and as a result, a change in install approach broke our build.

Pinned to a working install approach.

### DIFF
--- a/.github/workflows/library.yml
+++ b/.github/workflows/library.yml
@@ -179,7 +179,7 @@ jobs:
       - name: Update PATH in Windows
         if: matrix.os == 'windows'
         run: |
-          Add-Content -Path $env:GITHUB_PATH -Value "$env:APPDATA\Python\Scripts"
+          Add-Content -Path $env:GITHUB_PATH -Value "$env:USERPROFILE\.poetry\bin"
           Get-Content -Path $env:GITHUB_PATH
 
       - name: Configure Poetry to not use experimental installer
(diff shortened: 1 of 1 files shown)


.github/workflows/library.yml
```diff
@@ -171,15 +171,15 @@ jobs:
           pip install --upgrade pip setuptools
 
       - name: Install Poetry
-        uses: snok/install-poetry@v1
+        uses: snok/install-poetry@v1.3.3
```
SimonBiggs, line 174:
```suggestion
        uses: snok/install-poetry@v1.2.1
```

.github/workflows/library.yml
```diff
@@ -171,15 +171,15 @@ jobs:
...
         with:
           version: 1.2.1
           virtualenvs-create: true
 
       - name: Update PATH in Windows
         if: matrix.os == 'windows'
         run: |
-          Add-Content -Path $env:GITHUB_PATH -Value "$env:USERPROFILE\.poetry\bin"
+          Add-Content -Path $env:GITHUB_PATH -Value "$env:APPDATA\Python\Scripts"
```
SimonBiggs, line 182:
```suggestion
          Add-Content -Path $env:GITHUB_PATH -Value "$env:USERPROFILE\.poetry\bin"
```

pchlap (approved):
Looks good, thanks @SimonBiggs

PR #1770, opened 2022-10-30
update 3D gamma example

guanfada:
@SimonBiggs 
I updated the 3D gamma example. Please review, thanks!


SimonBiggs:
@Matthew-Jennings can you take this one?

Matthew-Jennings (changes requested):
Hi @guanfada,

It looks like you've accidentally added a `from_dicom.ipynb` to the root dir: https://github.com/pymedphys/pymedphys/pull/1770/files#diff-cf6d0f24bae0a9e789fae7dd9edcca225284b69f988ae48c517103693addda46

Could you please remove?

lib/pymedphys/docs/users/howto/gamma/from-dicom.ipynb
```diff
@@ -4,9 +4,9 @@
    "cell_type": "markdown",
```
Matthew-Jennings, line 4: <pre class="ql-syntax" spellcheck="false">Line #1.    Patient_Orientation = []
</pre><p>There is already a map like this in <code>pymedphys</code> :) See https://github.com/pymedphys/pymedphys/blob/fd91592173bed1220cf5feb170f865226301ad60/lib/pymedphys/_dicom/orientation.py#L65</p><p> </p><p>Also, I believe convention is to place "decubitis" in front of "left/right".</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymedphys/pymedphys/pull/1770/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

Matthew-Jennings, line 4: <p>So is the main theme of this PR to show explicit support for different orientations?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymedphys/pymedphys/pull/1770/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

Matthew-Jennings, line 4: <pre class="ql-syntax" spellcheck="false">Line #1.    axes_reference[0].sort()
</pre><p>Is this always needed? I think for demonstration purposes it is best to show what happens before doing this, and then to show how to do it safely.</p><p> </p><p>Even here, I reiterate a comment I made earlier that I'm not sure this is safe. It seems to me you run a strong risk of mapping the wrong coordinates to the dose grid voxels.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymedphys/pymedphys/pull/1770/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

Matthew-Jennings, line 4: <p>which slices? why/how were they chosen? DO they depict all of the different orientations? Are these orientations labelled for clarity?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymedphys/pymedphys/pull/1770/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

guanfada: Nice! I did not know it.

guanfada: yes, it is.

guanfada: No, for "head first supine" this .sort() is not needed.
But for other orientations, we have to .sort() for some axes (not all three).

I agree, it indeed can bring risk, but if we don't .sort() axes, gamma calculation can only be successfully performed for DICOM with "Head first supine"

Please advise!
Maybe we can add a new example showing how to deal with DICOM (non - "head first supine")?

guanfada: No, orientation is not labeled on slices. Here slices are on DICOM coordinates. 
For this example, patient orientation is "head first supine" so patient coordinates are DICOM coordinates are the same.

guanfada: Anyway, I would suggest we add an extra example to use DICOM like "feet first supine" 
What do you think?

guanfada: Also, I believe convention is to place "decubitis" in front of "left/right".
Yes, you are right. :-)

Matthew-Jennings: Let me check a few things and get back to you, sorry.

Matthew-Jennings:
Hi @guanfada,

Please pull the latest changes to your workstation and run a new Jupyter notebook I've added called `all_orientations.ipynb`. From that notebook, you'll see that `gamma` can at least handle one example of HFS, HFP, FFS, and FFP. There is a bug with decubitis orientations into which I'll look more closely later. Bottom line: there must be some different between the non-HFS file you're testing to the ones I'm using for testing in that notebook.

I'll note that rigorous handling of orientations has been a topic of discussion in PyMedPhys since I first joined in 2018. The solution is non-trivial, and needs a dedicated effort that probably requires a significant revamp of the `dicom` AND `gamma` modules. One problem is that there appears to be multiple ways in which one can encode coordinates in DICOM (e.g., GridFrameOffsetVector need not be positive and ascending).

We can chip away at things. I think the first thing for now is to identify the difference between your non-HFS examples and the ones on Zenodo that are tested in the notebook I added here. Note that the ones on Zenodo I physically CT scanned myself in all 8 orthogonal orientations.

SimonBiggs:
@Matthew-Jennings and @guanfada, do you believe that working on and validating an open source 3D gamma from DICOM for multiple patient orientations could be a publishable item?

guanfada:
@Matthew-Jennings 
I think I figured out the issue to do gamma calculation for decubitus orientation.
I made a new jupyter notebook called "from-dicom-HFDL.ipynb"

SimonBiggs:
Your package path is your installed version from PyPI:

image

You'll need to make sure you've done the following:

https://docs.pymedphys.com/en/latest/contrib/setups/setup-win.html

And then, create a Jupyter kernel that corresponds to your local development files:

https://docs.pymedphys.com/en/latest/contrib/tips/add-jupyter-kernel.html

Once you've made a custom Jupyter kernel you will need to reset your Jupyter server and then set the notebook to use that newly created kernel

guanfada:
> Your package path is your installed version from PyPI:
> 
> image
> 
> You'll need to make sure you've done the following:
> 
> https://docs.pymedphys.com/en/latest/contrib/setups/setup-win.html
> 
> And then, create a Jupyter kernel that corresponds to your local development files:
> 
> https://docs.pymedphys.com/en/latest/contrib/tips/add-jupyter-kernel.html
> 
> Once you've made a custom Jupyter kernel you will need to reset your Jupyter server and then set the notebook to use that newly created kernel

I have no problems using Jupyter for the from-dicom.ipynb, I can also download other dicom files.
Just these two newly added HFDL dicom files could not be found. 

I upload these two files to data.
I updated urls.json by adding two lines at the end.

Did I miss any steps?

guanfada:
> @Matthew-Jennings and @guanfada, do you believe that working on and validating an open source 3D gamma from DICOM for multiple patient orientations could be a publishable item?

@SimonBiggs yes, I think so. We can make it like a quick technical note.

Matthew-Jennings:
> I think I figured out the issue to do gamma calculation for decubitus orientation.
> I made a new jupyter notebook called "from-dicom-HFDL.ipynb"

Sorry, are the new files you've added the ones with which you first identified the descending coords issue? Or are you addressing a different point here?

guanfada:
> > I think I figured out the issue to do gamma calculation for decubitus orientation.
> > I made a new jupyter notebook called "from-dicom-HFDL.ipynb"
> 
> Sorry, are the new files you've added the ones with which you first identified the descending coords issue? Or are you addressing a different point here?

I used the HFDL dicom file in your zip (you have 8 dicom files for 8 orientations). 
Aslo, the .sort() I used yesterday is incorrect because it will give wrong coordinates in DICOM for non-HFS.
In your all-orientations.ipynb, gamma can't be calculated for decubitus orientations.
In this new from-dicom-HFDL.ipynb, I added some operations to translate patient axes to DICOM axes, and now gamma can be calculated for any orientation.
I also labeled the orientation on the figure(s). 
I just used the same dicom file as both reference and evaluation. 
dose maps and Global gamma results_slice_63

guanfada:
> Your package path is your installed version from PyPI:
> 
> image
> 
> You'll need to make sure you've done the following:
> 
> https://docs.pymedphys.com/en/latest/contrib/setups/setup-win.html
> 
> And then, create a Jupyter kernel that corresponds to your local development files:
> 
> https://docs.pymedphys.com/en/latest/contrib/tips/add-jupyter-kernel.html
> 
> Once you've made a custom Jupyter kernel you will need to reset your Jupyter server and then set the notebook to use that newly created kernel

I deleted everything related to pymedphys, and re-installed following your steps.
unfortunately, the same issue. 
BTW, if I switch to the "pymedphys" jupyter kernel, it always shows "disconnected"
I have to use the python 3 kernel, but obviously using the pymedphys 3.9.3 under my python 3.8's site packages is not what I want.

Matthew-Jennings:
> > > I think I figured out the issue to do gamma calculation for decubitus orientation.
> > > I made a new jupyter notebook called "from-dicom-HFDL.ipynb"
> > 
> > 
> > Sorry, are the new files you've added the ones with which you first identified the descending coords issue? Or are you addressing a different point here?
> 
> I used the HFDL dicom file in your zip (you have 8 dicom files for 8 orientations). Aslo, the .sort() I used yesterday is incorrect because it will give wrong coordinates in DICOM for non-HFS. In your all-orientations.ipynb, gamma can't be calculated for decubitus orientations. In this new from-dicom-HFDL.ipynb, I added some operations to translate patient axes to DICOM axes, and now gamma can be calculated for any orientation. I also labeled the orientation on the figure(s). I just used the same dicom file as both reference and evaluation. dose maps and Global gamma results_slice_63

No worries. Sorry, @guanfada, I'll be a little tied up until the weekend - I'll endeavour to review then.

SimonBiggs:
@Matthew-Jennings, where do you reckon this PR is up to?

Matthew-Jennings:
> @Matthew-Jennings, where do you reckon this PR is up to?

Sorry I let this run stale. I'll need to re-review at this point.

Matthew-Jennings:
Hi @guanfada and @SimonBiggs,

Finally re-reviewing this: although I hate to see work go to waste, my completely honest feeling is that this PR should be abandoned, since:

1. (minor) `from-dicom.ipynb` and `from-dicom-HFDL.ipynb` have significant overlap. One notebook showing both examples would be sufficient.
2. (major) The main purpose seems to be to demonstrate a workaround to make `gamma` work for non-HFS orientations. At this stage, I'd rather not include handling of other orientations in formal documentation, since:
   - I'm not completely sure, but the workaround doesn't appear robust to me. I think we're better off letting it "break" for non-HFS orientations rather than prematurely (and formally) generalising a solution. At this stage, it feels safer to handle issues on a case-by-case basis as people raise them, or leave it to the user.
   - I'd much rather `pymedphys` simply support other orientations natively, with minimal "thinking" required by the user. I can try to get to our once-and-for-all orientations discussion/conclusion in #1771

Open to alternative ways forward; just my two cents.

guanfada:
Hi Matt, this sounds good to me.Thank you! FadaOn Mar 18, 2024, at 1:45 AM, Matthew Jennings ***@***.***> wrote:﻿
Hi @guanfada and @SimonBiggs,
Finally re-reviewing this: although I hate to see work go to waste, my completely honest feeling is that this PR should be abandoned, since:

(minor) from-dicom.ipynb and from-dicom-HFDL.ipynb have significant overlap. One notebook showing both examples would be sufficient.
(major) The main purpose seems to be to demonstrate a workaround to make gamma work for non-HFS orientations. At this stage, I'd rather not include handling of other orientations in formal documentation, since:

I'm not completely sure, but the workaround doesn't appear robust to me. I think we're better off letting it "break" for non-HFS orientations rather than prematurely (and formally) generalising a solution. At this stage, it feels safer to handle issues on a case-by-case basis as people raise them, or leave it to the user.
I'd much rather pymedphys simply support other orientations natively, with minimal "thinking" required by the user. I can try to get to our once-and-for-all orientations discussion/conclusion in #1771



Open to alternative ways forward; just my two cents.

SimonBiggs:
I'm happy to defer to your thoughts here @Matthew-Jennings on this

PR #1790, opened 2023-02-14
Update pymedphys deps

### LINKED ISSUE #1785: Import errors with pymedphys 0.39.3 and pylinac 2.3.2 in Python 3.10
I've been trying to install pymedphys 0.39.3 with Python 3.10.7 [1] but am getting errors "cannot import name 'Iterable' from 'collections'" from pylinac 2.3.2, i.e. the version of pylinac specified in https://github.com/pymedphys/pymedphys/blob/v0.39.3/requirements.txt.

This seems to be a common error when running pre-Python 3.10 code on 3.10 (see e.g. https://stackoverflow.com/a/72032097). Locally I worked around this my installing the latest version of pylinac (3.7.1) which works with Python 3.10. 

pylinac 3.2.0 is the earliest version of pylinac which explicitly lists Python 3.10 in the supported languages on PyPi https://pypi.org/project/pylinac/3.2.0/. 

I'm not sure how you determine exactly which versions to require but hope this is useful info. Best regards and thanks for the great tool.

[1] I wanted to test the fix to https://github.com/pymedphys/pymedphys/issues/1592 and use pymedphys using our current default version of Python.

Matthew-Jennings:
Maybe fixes #1785

So far, I've:
- Pinned `numba` to `>=0.56.2`.
  - This appears to fix installation using Python 3.10 on Windows.
- Pinned `pylinac` to `>=3.2.0`
  - @pylonsarenice indicated that this was the earliest version of `pylinac` that explicitly supports Python 3.10 (although I'll note that I couldn't find this documented anywhere)
  - This breaks the WLutz tests, since the WLutz API underwent significant changes in `pylinac` v3.0.0. (@SimonBiggs, I tried to fix this myself, but you've got a fairly complicated vendoring of `pylinac` going on and I think you're just going to be much better placed to resolve this)
- Updated the lock file to latest versions of dependencies.
  - This broke some tests that use Shapely, which also appears to have undergone a couple of minor API changes. I introduced provisional fixes, but I do think you'd better take a look and see if they're you're preferred solution, @SimonBiggs.
  - This also introduced many, many new "file" lines into the lock file - I believe as a result of using `poetry >= 1.3.0`


SimonBiggs:
Thanks @Matthew-Jennings, I'll see what I can do about those parts that touch code that I've been heavily involved in.

SimonBiggs:
Hi @msobolewski and @JacobMcAloney,

Given the WLutz code never got productionised, it is actually quite a significant maintenance burden in its current form. (Can't actually update PyMedPhys to the latest Python for everyone without potentially undergoing a couple of days work, at least, on the WLutz side of things).

My proposal is to remove the entirety of the WLutz code from PyMedPhys until someone has the capacity to work further on it.

Would you both be okay with me undergoing that approach?

Of course, you can still continue to use the old version of PyMedPhys. But newer versions will no longer include this piece of software.

Cheers,
Simon

Matthew-Jennings:
Also, shall we make a release once this lands? Combined Python 3.10 / Windows support + faster gamma...

SimonBiggs:
> Also, shall we make a release once this lands? Combined Python 3.10 / Windows support + faster gamma...

Definitely :)

PR #1795, opened 2023-04-20
Purge and clean

SimonBiggs:
Hi @Matthew-Jennings,

This is me building on top of your initial update work. I am also extending this to work with Python 3.11.

I am also undergoing an aggressive prune so that we can have this library align more closely with our maintenance capacity.

We can certainly bring some things back that I have pruned. But for now, by stripping back some of the weight enables us to keep up with the required maintenance.

Cheers,
Simon


SimonBiggs:
I'm going to do a merge of this one. More than happy for us to revert. But right now, without these changes the build is broken. So I would prefer to pull things back in piece-wise, if they are desired.

Matthew-Jennings:
Just having a look now

lib/pymedphys/_experimental/fileformats/mapcheck/core.py
```diff
@@ -37,7 +37,7 @@ def read_mapcheck_txt(file_name):
 
     Mapcheck = namedtuple("Mapcheck", ["x", "y", "dose"])
 
-    with open(file_name, "r") as mapcheck_file:
+    with open(file_name) as mapcheck_file:
```
Matthew-Jennings, line 40: why this change?

Matthew-Jennings (approved):
I mean, I think I'm happy. But I can't help to just prod with an "are you sure?"

Perhaps a better question: what are the risks of/disadvantages to performing this merge?

lib/pymedphys/_experimental/fileformats/mapcheck/core.py
SimonBiggs: Added the following pre-commit hook:

https://github.com/asottile/pyupgrade

Matthew-Jennings: Cool!

Matthew-Jennings:
> > Perhaps a better question: what are the risks of/disadvantages to performing this merge?
> 
> We are losing the following capabilities for now:
> 
> * Streamlit
> * Binary building
> * WLutz
> * A range of experimental features
> 
> All were removed due to there being highly customised logic that would take a significant amount of time to upgrade to the latest version of their corresponding dependencies.

I guess I was asking: what are the worst-case implications of removing those capabilities?

SimonBiggs:
> I guess I was asking: what are the worst-case implications of removing those capabilities?

Someone has to downgrade PyMedPhys in order to keep using a feature they were using previously?

site-specific/mays/dfgui.py
sjswerdloff, line 1: adding python 2.7 packages?
pymedphys doesn't support python 2
this looks like it's likely to fail in some circumstances.

sjswerdloff (approved):
This PR could have been split in to a few different and easier to review PRs:

1. File deletion
2. pyproject.toml updates (elimination of many package dependencies no longer needed because of deleted files)
3. string formatting conversion to default numeric order rather than explicit numeric order (does this improve readability? performance? or is it just aesthetics?)
5. compatibility changes (e.g. from x to x.geoms)
6. file open to default read only vs. explicit read only (and having as a contained PR with that as the title would have explained things).  (does this improve readability? performance? or is it just aesthetics?)

The explanation for elimination was a bit thin.  
What criteria were used for choosing what to eliminate?

Build changes didn't have explanation of why certain changes were made.
Many are probably due to eliminated code, but not all.

SimonBiggs:
> Some of us might need a heads up and a weekend to have the time to go in and try to tidy up and keep our work _in_ the package, even if it is experimental.

What if I merge this, then make a PR that pulls your code back in, then the errors will be just the parts that are due to your code?

site-specific/mays/dfgui.py
SimonBiggs: These are removing the python 2 fillers. This was due to running pyupgrade linter tool.

SimonBiggs:
> Build changes didn't have explanation of why certain changes were made.

Yeah... there are a huge number of issues trying to upgrade to Python 3.11.

I'm quite time constrained. I figured this was the easiest way to roll it out given the time restrictions :slightly_frowning_face:

sjswerdloff:
> > Some of us might need a heads up and a weekend to have the time to go in and try to tidy up and keep our work _in_ the package, even if it is experimental.
> 
> What if I merge this, then make a PR that pulls your code back in, then the errors will be just the parts that are due to your code?

That's an excellent suggestion.
It sets me up for success rather than my having to "re-do".

SimonBiggs:
> It sets me up for success rather than my having to "re-do".

:slightly_smiling_face:, kk, I'll do that first thing after I merge

SimonBiggs:
> This PR could have been split in to a few different and easier to review PRs:
> 
> 1. File deletion
> 2. pyproject.toml updates (elimination of many package dependencies no longer needed because of deleted files)
> 3. string formatting conversion to default numeric order rather than explicit numeric order (does this improve readability? performance? or is it just aesthetics?)
> 4. compatibility changes (e.g. from x to x.geoms)
> 5. file open to default read only vs. explicit read only (and having as a contained PR with that as the title would have explained things).  (does this improve readability? performance? or is it just aesthetics?)

I do agree with this in principle.

3. and 5. was all to do with pulling in `pyupgrade`.

4. Compatibility changes were actually in a separate PR, but it couldn't merge:

https://github.com/pymedphys/pymedphys/pull/1790

1. and 2. were both to try and help 4. merge.

crcrewso:
> > Perhaps a better question: what are the risks of/disadvantages to performing this merge?
> 
> We are losing the following capabilities for now:
> 
> * Streamlit
> * Binary building
> * WLutz
> * A range of experimental features
> 
> All were removed due to there being highly customised logic that would take a significant amount of time to upgrade to the latest version of their corresponding dependencies.

Missed this one, All for most of the work. WLutz can be replaced with pylinac if the request ever comes back up. 
If I have to build it as a binary again I'll submit the patches.

SimonBiggs:
Hi @crcrewso,

Any of the above can come back. And, sooner is probably better than later. Is there a particular section of PyMedPhys that was removed here that you'd like to see come back?

I can make a PR that pulls just that back in, and then maybe you can do the finishing touches to get it back working again?

crcrewso:
@SimonBiggs 

Personally, I only see being able to build binaries as worth maintaining. 

There's a scope question that's been in my mind recently. As there are now other parallel projects in the community, does this group want to maintain all of the tooling necessary for standard machine QA, like Wlutz, Picket Fence, CatPhan analysis, etc. or should we integrate the libraries and build interfaces consistent with other modules?

SimonBiggs:
> Personally, I only see being able to build binaries as worth maintaining.

Yup, would you be able to make a separate issue, and I'll see about pulling that back in next week. Which binary approach were you using? (We had two).

> There's a scope question that's been in my mind recently.

Potentially also worth discussing in a new issue?

crcrewso:
> > Personally, I only see being able to build binaries as worth maintaining.
> 
> Yup, would you be able to make a separate issue, and I'll see about pulling that back in next week. Which binary approach were you using? (We had two).
> 
> > There's a scope question that's been in my mind recently.
> 
> Potentially also worth discussing in a new issue?

Completely Agree. I think two separate discussion issues would make the most sense!

Issue #1796, opened 2023-04-20
Move Pinnacle out of Experimental?


SimonBiggs:
Hi @pchlap,

How feasible would it be for the pinnacle module to be moved out of experimental?

pchlap:
Hi @SimonBiggs,

We could easily drop the Pinnacle module out of experimental and into the main library. If we do that it's really important to have the tool be really noisy about that it is not to be used for any type of clinical work (and for research only). Especially because the RTPLAN export feature doesn't really work and I don't think anyone really has the motivation to get that working.

What do you think?

SimonBiggs:
> What do you think?

Yup. Is anyone using the RTPLAN export feature?

pchlap:
> > What do you think?
> 
> Yup. Is anyone using the RTPLAN export feature?

The only thing its really useful for is to make sure we can maintain that linkage between the RTSTRUCT -> RTPLAN -> RTDOSE... If the RTPLAN wasn't to be generated at all the only way to link them would be using the frame of reference which would often be ambiguous due to multiple structure sets floating around in the same frame of refernece.

SimonBiggs:
@pchlap, what do you think the best path forward is?

sjswerdloff:
There is a way to directly reference the RT Structure Set from the RT Dose, if you encode the RT DVH module:
https://dicom.innolitics.com/ciods/rt-dose/rt-dvh/300c0060
The reference to the plan is still required for the treatment planning scenario, and IMHO it makes a lot of sense to know the plan that was used to calculate the Dose.
The RT DVH module:
"...may supplement dose data in the [RT Dose Module](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.8.3.html#sect_C.8.8.3) and/or [RT Dose ROI Module](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.8.7.html#sect_C.8.8.7), or it may be present in the absence of other dose data."

So you can have your RT Dose contain the "pixel data" (dose voxels) and include just enough RT DVH to give you a place to put the referenced RT Structure Set.

I haven't checked to see what the intended use cases are, but I thought the above might be useful for some situations

pchlap:
Thanks @sjswerdloff, this is useful.

[...]

PR #1803, opened 2023-05-09
Move Pinnacle module out of experimental

pchlap:
Pull request to move Pinnacle module out of experimental and into the main library as discussion in #1796

### DIFF
--- a/lib/pymedphys/_pinnacle/pinnacle.py
+++ b/lib/pymedphys/_pinnacle/pinnacle.py
@@ -70,8 +70,10 @@ class PinnacleExport:
 
         # Show a warning to ensure that this module is used with caution
         warnings.warn(
-            "The Pinnacle Export Module is intended for research purposes only, "
-            "use with caution.",
+            "The Pinnacle Export Module is intended for research purposes only. Parts "
+            "of the DICOM conversion are hard-coded and completely ignore your data. "
+            "This has proven useful for research applications but is not appropriate "
+            "for clinical use.",
             category=UserWarning,
             stacklevel=2,
         )


pchlap:
> Btw @pchlap,
> 
> Here's an example of how to use the deprecated function for the API exposure:
> 
> https://github.com/pymedphys/pymedphys/blob/c28f9ca5fc74a5080d5118f0fac7e2caa6827591/lib/pymedphys/__init__.py#L11-L14

Thanks a lot @SimonBiggs, I have added that in for the Pinnacle module. I've also added in a warning when the PinnacleExport module is used now that this has been moved out of experimental. Curious to hear what you think.

lib/pymedphys/experimental/pinnacle.py
```diff
@@ -0,0 +1,21 @@
+from pymedphys._vendor.deprecated import deprecated as _deprecated
+from pymedphys.pinnacle import PinnacleExport as _PinnacleExport
+from pymedphys.pinnacle import PinnacleImage as _PinnacleImage
+from pymedphys.pinnacle import PinnaclePlan as _PinnaclePlan
+from pymedphys.pinnacle import export_cli as _export_cli
+
+PinnacleExport = _deprecated(
```
SimonBiggs, line 7: Make sure to include the "name" part too.

...I also don't believe we have a deprecated logic for the CLI. Potentially still expose the CLI as it was, and get it to call this deprecated function, that way, the warning will come up with the CLI?

SimonBiggs: Hi @pchlap,

Sorry I didn't respond here. I actually can't remember. ... looks like I needed to leave more breadcrumbs for my future self, and others ...

SimonBiggs (approved):
LGTM @pchlap.

Does the CLI show appropriate pass throughs with deprecation warnings?
Are you happy with the level of warnings that are presented to the users throughout this modules usage?

pchlap:
> LGTM @pchlap.
> 
> Does the CLI show appropriate pass throughs with deprecation warnings? Are you happy with the level of warnings that are presented to the users throughout this modules usage?

Hi @SimonBiggs, the CLI is showing the warning about this being a research only tool... but it doesn't pass through the deprecation warning when calling the CLI from the experimental module (that warning shows up fine if I use it from a python script)... I can't figure out why this is. Any ideas?

lib/pymedphys/_pinnacle/pinnacle_cli.py
SimonBiggs, line 1: > Hi @SimonBiggs, the CLI is showing the warning about this being a research only tool... but it doesn't pass through the deprecation warning when calling the CLI from the experimental module (that warning shows up fine if I use it from a python script)... I can't figure out why this is. Any ideas?

@pchlap potentially you need to copy here, leaving the CLI in both places for now? And directly call the experimental, deprecated API?

pchlap: I think I was able to trace down the problem. The `deprecated` function uses `warnings.warn` to show the warning. The warning category is `DeprecationWarning`. Since I am also using `warnings.warn` to show the other warning which does show up, I had a look and noticed that this one uses the category `UserWarning`. So I tried changing the category in the `deprecated` function and that indeed does show up in the CLI. So it seems to be something about the `DeprecationWarning` that it doesn't show in the CLI, I suppose that makes sense since from the CLI you may not be interested in a method or variable being deprecated...

Anyway, I'll keep this simple, I'll put in another `UserWarning` into the deprecated CLI which will indicate where to find the new CLI.

lib/pymedphys/cli/experimental/pinnacle.py
```diff
@@ -33,7 +34,20 @@ def pinnacle_cli(subparsers):
     return pinnacle_parser
 
 
+def deprecated_export_cli(args):
```
pchlap, line 37: Hi @SimonBiggs, this is the solution I have come up with and now shows the appropriate warnings at the right times. Are you happy with this? If so I'll go ahead an merge :)

pchlap: OK yep sure thing. Currently the warning just states: `The Pinnacle Export Module is intended for research purposes only, use with caution.`

I will expand on that with something like:
`The Pinnacle Export Module is intended for research purposes only. Parts of the DICOM conversion are hard-coded and completely ignore your data. This has proven useful for research applications but is not appropriate for clinical use.`

lib/pymedphys/_pinnacle/pinnacle.py
```diff
@@ -70,8 +70,10 @@ def __init__(self, path, logger=None):
 
         # Show a warning to ensure that this module is used with caution
         warnings.warn(
-            "The Pinnacle Export Module is intended for research purposes only, "
-            "use with caution.",
+            "The Pinnacle Export Module is intended for research purposes only. Parts "
+            "of the DICOM conversion are hard-coded and completely ignore your data. "
+            "This has proven useful for research applications but is not appropriate "
+            "for clinical use.",
```
SimonBiggs, line 76: Thanks @pchlap, that significantly reduces my stress levels. Thank you :slightly_smiling_face:.

PR #1818, opened 2023-12-06
Improve error message for missing packages

### LINKED ISSUE #1746: Better guidance for missing packages
> > Yup, by design. If you run `pip install pymedphys[user,tests]` that should be sufficient.
> 
> OK great, the only thing is the error message to the user says to simply install `pip install pymedphys[user]`, is this intended?

Yeah... that should cover most use cases. It potentially could be worth some further logic to detect if the user needs to install other optional dependencies also... But... if a user is "using" the pymedphys library then that install message should cover all of those use cases.

...if they are testing, then `pymedphys[user]` is not enough. 

The fix for this could be, to have a dependency list created from the requirements files that is saved into the python package via propagate, and then read out the following `modname` variable:

https://github.com/pymedphys/pymedphys/blob/778e33769ead3ed6a6b366fd816f59aa8a2d8e76/lib/pymedphys/_vendor/apipkg/__init__.py#L248-L256

And then, if that `modname` variable is within the `user` dependency list, then ask the user to run `pip install pymedphys[user]`. If it is in the tests dependency list, then ask them to run `pip install pymedphys[user,tests]`...

_Originally posted by @SimonBiggs in https://github.com/pymedphys/pymedphys/issues/1744#issuecomment-1261938089_

nlambriICH:
Closes #1746.

I think this [line ](https://github.com/pymedphys/pymedphys/compare/main...nlambriICH:missing_packages?expand=1#diff-0eaaf6f0a44eb47e5945ee0f9694578f4880a71431dab54cc390a50d887f5b22R246) is dangerous, because I'm importing from `_imports`. If `tomlkit` is not available, this code will crash.

Should I parse the dependencies file without relying on other libraries?


SimonBiggs:
> I think this [line ](https://github.com/pymedphys/pymedphys/compare/main...nlambriICH:missing_packages?expand=1#diff-0eaaf6f0a44eb47e5945ee0f9694578f4880a71431dab54cc390a50d887f5b22R246) is dangerous, because I'm importing from `_imports`. If `tomlkit` is not available, this code will crash.
> 
> Should I parse the dependencies file without relying on other libraries?

Yup, good flag. We can fix this though by setting this dependency to not be optional with pyproject and no longer importing from _imports here, but instead just import it directly. That would make sense to me.

Could we first try importing from https://docs.python.org/3/library/tomllib.html (available in Python 3.11 standard lib onwards) and then fall back to an external library here?

And then make the external toml library required only if the user is using something less than Python 3.11

nlambriICH:
[This ](https://github.com/pymedphys/pymedphys/pull/1818/files#diff-c3c21ad26de9b9a4ad019364896ecbf14f893e3e43ad7408d3ed43680c05123eR198) implementation is not robust, but I couldn't find a package to parse the python version in the same format used by poetry. Do you have other suggestions?

SimonBiggs:
Use the tool.poetry.extras :)

https://github.com/pymedphys/pymedphys/blob/c4e4ab3c42d1acda88e2a6673aab217aeee4fec4/pyproject.toml#L206

nlambriICH:
> Use the tool.poetry.extras :)
> 
> https://github.com/pymedphys/pymedphys/blob/c4e4ab3c42d1acda88e2a6673aab217aeee4fec4/pyproject.toml#L206

Sorry I don't get it... I want to parse the python required version from [here](https://github.com/pymedphys/pymedphys/pull/1818/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R50) such that I add tomlkit to the extras only if `python >= 3.11`.

nlambriICH:
I've also changed the propagate workflow [here](https://github.com/pymedphys/pymedphys/pull/1818/files#diff-1490999fe528c382b1d9c209834344656f9d34d38b149af5347fa8c79079bb62R321) to commit the new `dependencies-extra.txt` file. Is this correct?

SimonBiggs:
@Matthew-Jennings, might you have a chance to help here?

nlambriICH:
I might be able to parse the version using `poetry-core` [here ](https://github.com/python-poetry/poetry-core/blob/main/src/poetry/core/version/pep440/parser.py#L67). But this would require `poetry-core` as a new dependency.

SimonBiggs:
One option might just be to have this only work if toml is available
already? If not just fall back to previous approach?

nlambriICH:
The main problem is in propagate.py: if tomlkit is optional for python>=3.11 then we have to parse the python version to know whether we need to add it to the poetry extras.

I propose to make either tomlkit or toml required such that we can always read the `dependencies-extra.txt` file.

SimonBiggs:
Go for it :)

nlambriICH:
Done: tomlkit is now a required dependency. I think in this way the code is simpler to manage.

I've also reverted the [propagate job](https://github.com/pymedphys/pymedphys/pull/1818/commits/9f916959db0ec2bbe465d6021a9f1b0fb8f7b152) to commit only the lock file, as I understood this is needed to check whether propagate is needed. Is that correct?

pyproject.toml
```diff
@@ -250,6 +248,9 @@ user = [
 [tool.poetry.scripts]
 pymedphys = "pymedphys.__main__:main"
 
+[tool.poetry.group.dev.dependencies]
+ipykernel = "^6.27.1"
```
SimonBiggs, line 252: Can we make sure this gets included as a dev requirement the same way others are:
https://github.com/pymedphys/pymedphys/pull/1818/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L124

nlambriICH: Actually I don't know how this was added to the dependency list :sweat_smile:.

I removed it.

lib/pymedphys/_vendor/apipkg/__init__.py
```diff
@@ -236,24 +236,38 @@ def __getattribute__(self, name):
...
+                        break
+
                 raise ModuleNotFoundError(
                     f"""
                     PyMedPhys was unable to import "{no_scope_modname}.{name}".
-                    The easiest way to fix this issue is to use the "[user]"
-                    option when installing PyMedPhys. For example,
-                    with pip this can be done by calling
-                    "pip install pymedphys[user]=={__version__}".
+                    The easiest way to fix this issue is to use the "[extra]"
```
SimonBiggs, line 266: Do you need to parameterise the extra here? When testing this locally what does it print out if you uninstall a package?

lib/pymedphys/_vendor/apipkg/__init__.py
```diff
@@ -236,24 +236,38 @@ def __getattribute__(self, name):
...
+                    option when installing PyMedPhys. Please run
+                    "pip install pymedphys[{extra}]=={__version__}".
```
SimonBiggs, line 268: Same here, does this need an f at the beginning of the string?

pyproject.toml
SimonBiggs: :)

lib/pymedphys/_vendor/apipkg/__init__.py
nlambriICH: This is the printed message:

Exception_pymedphys

I didn't want to parametrize the first "extra" to keep a similar error message.

lib/pymedphys/_vendor/apipkg/__init__.py
nlambriICH: The f at the beginning of the triple-quote string was enough to get the correct message.

lib/pymedphys/_vendor/apipkg/__init__.py
SimonBiggs: Beautiful ❤️

lib/pymedphys/_vendor/apipkg/__init__.py
SimonBiggs: Awesome 😎

SimonBiggs (approved):
LGTM! Thanks for all the back and forth on this one @nlambriICH! :)

nlambriICH:
Thanks to you @SimonBiggs for your feedback! 😄

PR #1806, opened 2023-05-26
Run poetry update

SimonBiggs:
Fixing https://github.com/pymedphys/pymedphys/pull/1805 and https://github.com/pymedphys/pymedphys/pull/1804

---

Also removed the dependency "reportlab", one of its dependencies was having install issues. This was a left over lower bound pin from having pylinac as a dependency. We no longer have pylinac, so we don't need reportlab.

---

Also, it appears Poetry version `1.5.0` changes `requirements.txt` output format. This broke the propagate test. So upgraded poetry version within the CI.

---

Once the first one of you sees this, and is happy with it, feel free to just approve and merge.

PR #1825, opened 2024-03-12
0.40.0 release prep

Matthew-Jennings:
See list of merged PRs since last release [here](https://github.com/pymedphys/pymedphys/pulls?q=is%3Apr+is%3Amerged).

I'm not sure about how/whether to include some of them in the CHANGELOG:
-  @mguerrajordao, could you please update the dummy CHANGELOG entry I added [here](https://github.com/pymedphys/pymedphys/blob/c446267d34d1cdfbde8c94442459175dbd07b586/CHANGELOG.md?plain=1#L21) for #1773? Or, since it doesn't look like @mguerrajordao has been active since Dec 2022, perhaps @SimonBiggs or @sjswerdloff as the reviewers wouldn't mind updating?
- There were a few (developer facing?) changes that I wasn't sure were worth including. Could I please get a second opinion on whether it's worth adding entries for the following PRs? I'm happy to add entries myself, just not sure on inclusion criteria.
  - #1790
  - #1795
  - #1806
  - #1807
  - #1818 

Cheers,
Matt


SimonBiggs (approved):

sjswerdloff:
If we merge in #1827, I will update the ChangeLog to reference that if desired (additional linac state code for TRF).
But I think the ChangeLog wording covers the TRF changes sufficiently as it is.

sjswerdloff (approved):
LGTM

sjswerdloff:
@Matthew-Jennings please feel free to revert the changelog edit I made.
But the commit served its purpose.  Other than pyright, the tests are coming up clean.

sjswerdloff:
hmmm.  the problem with cypress may be a real thing.
streamlit has updated, and it has what appear to be some breaking changes, and some warnings about calls being made that will be obsolete as of mid April.  So...  not sure where to go with this, @SimonBiggs I think you may need to look at this.

SimonBiggs:
As a first pass we may be able to do an upper pin on streamlit to have the tests pass?

sjswerdloff:
Pinning streamlit below 1.32 doesn't seem to have resolved the issue.
I think we can either iterate down (try 1.31, then 1.30) or just pin at the last known good version of 1.29.
If pinning streamlit at 1.29 doesn't work, we will need to look at the differences in the lock file.
It might have something to do with Electron itself.

Issue #1833, opened 2024-03-20
Anthropic's Claude integrations


Matthew-Jennings:
Also, if we do end up having a few GitHub apps, I'd be more than happy to donate my personal Anthropic API token (not company associated, but my own account) to the PyMedPhys org, which will allow us to set up a "Claude issues bot" GitHub action.

The first pass could just allow us to call Claude in an issue, and it will then respond directly within the GitHub issue to provide help to the user.

Second and third passes could allow us to also include the PyMedPhys code / GitHub PR content and then have Claude help us with first passes on code reviews and the like 🙂

_Originally posted by @SimonBiggs in https://github.com/pymedphys/pymedphys/issues/1831#issuecomment-2008633058_

SimonBiggs:
Claude, I'd be interested for you to summarise your understanding of the PyMedPhys library from your training data.

Matthew-Jennings:
> Claude, I'd be interested for you to summarise your understanding of the PyMedPhys library from your training data.

Curious this didn't work...

SimonBiggs:
I reckon it might actually be worth making an AI section in PyMedPhys, (or a PyMedPhys AI CLI command) and then have all of the AI logic be handled in Python and have the GitHub actions just call the PyMedPhys CLI.

That way you can more readily iterate. And, trust me, to really make these things fly and do awesome this you're going to want to be able to do things like have a "tree of prompts" that feed into each other. Some the Haiku model and some Opus.

I'd also highly recommend checking out trio:
https://trio.readthedocs.io/en/stable/

It's really awesome being able to run multiple AI calls asynchronously and then collecting them all together.

Matthew-Jennings:
> I reckon it might actually be worth making an AI section in PyMedPhys, (or a PyMedPhys AI CLI command) and then have all of the AI logic be handled in Python and have the GitHub actions just call the PyMedPhys CLI.
> 
> That way you can more readily iterate. And, trust me, to really make these things fly and do awesome this you're going to want to be able to do things like have a "tree of prompts" that feed into each other. Some the Haiku model and some Opus.
> 
> I'd also highly recommend checking out trio: https://trio.readthedocs.io/en/stable/
> 
> It's really awesome being able to run multiple AI calls asynchronously and then collecting them all together.

[...]

PR #1836, opened 2024-03-21
Add Claude response to "@!claude" in new or updated comments on issues or PRs

Matthew-Jennings:
This is a cool one! First pass for #1833. Check out Claude's response [here](https://github.com/Matthew-Jennings/pymedphys/pull/1#issuecomment-2011701910)!

A few things to consider (very open to ideas):

1. Which model? It's set to Haiku at the moment.
2. Is the `@!` syntax okay? Can anyone think of something prettier?
3. Should Claude quote the original comment in its response as it is now?
4. Haven't tested for issues (difficult when not in main)
5. What do you think of the response format overall?
6. Any other body params to consider tweaking? See [this link](https://docs.anthropic.com/claude/reference/messages_post) for reference. E.g.,
   - `system`
   - `stop_sequences`
   - `temperature`
   - `top_k`

### DIFF
--- a/.github/workflows/claude-assistant.yml
+++ b/.github/workflows/claude-assistant.yml
@@ -12,11 +12,11 @@ jobs:
     steps:
     - name: Get stripped comment body
       id: get-comment-body
-      run: echo content=$(echo ${{github.event.comment.body}} | sed 's/@!claude//g') >> $GITHUB_OUTPUT
+      run: echo comment=$(echo ${{github.event.comment.body}} | sed 's/@!claude//g') >> $GITHUB_OUTPUT
 
     - name: Get Claude's response to issue comment
       id: claude-response
-      if: steps.get-comment-body.outputs.content != ''
+      if: steps.get-comment-body.outputs.comment != ''
       env:
         ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
       run: |
@@ -29,18 +29,18 @@ jobs:
                 "model": "claude-3-haiku-20240307",
                 "max_tokens": 1024,
                 "messages": [
-                    {"role": "user", "content": "${{ steps.get-comment-body.outputs.content }}"}
+                    {"role": "user", "content": "${{ steps.get-comment-body.outputs.comment }}"}
                 ]
             }') >> $GITHUB_OUTPUT
 
     - name: Create comment
-      if: steps.get-comment-body.outputs.content != '' && steps.claude-response.outcome == 'success'
+      if: steps.get-comment-body.outputs.comment != '' && steps.claude-response.outcome == 'success'
       uses: peter-evans/create-or-update-comment@v4
       with:
           issue-number: ${{ github.event.issue.number }}
           body: |
             @!claude_response
 
-            > ${{ steps.get-comment-body.outputs.content }}
+            > ${{ steps.get-comment-body.outputs.comment }}
 
-            ${{ steps.claude-response.outputs.response }}
+            ${{ fromJson(steps.claude-response.outputs.response).content.text }}
--- a/.github/workflows/claude-assistant.yml
+++ b/.github/workflows/claude-assistant.yml
@@ -17,19 +17,21 @@ jobs:
     - name: Get Claude's response to issue comment
       id: claude-response
       if: steps.get-comment-body.outputs.content != ''
+      env:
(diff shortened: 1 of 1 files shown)


pchlap:
This is very cool! Great work @Matthew-Jennings. I think @! is as good as any option... I guess `@claude` would be cool but I suppose that might get mixed up with @-ing github users directly...

I'll let Simon comment on the model parameter specifics. I like that Claude quotes the original question to avoid confusion if lots of questions are being asked in the same PR. But this looks great overall, even if Clause isn't prepared to join as a PyMedPhys maintainer just yet 😆

SimonBiggs:
https://github.com/Matthew-Jennings/pymedphys/pull/1#issuecomment-2013678963

SimonBiggs:
Let's just use "Claude" without the @ or !

Also made another comment here:
https://github.com/Matthew-Jennings/pymedphys/pull/1#issuecomment-2013690205

.github/workflows/claude-assistant.yml
```diff
@@ -0,0 +1,46 @@
...
+
+jobs:
+  post_claude_response:
+    if: contains(github.event.comment.body, '@!claude')
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Get stripped comment body
+      id: get-comment-body
+      run: echo comment=$(echo ${{github.event.comment.body}} | sed 's/@!claude//g') >> $GITHUB_OUTPUT
```
SimonBiggs, line 15:
```suggestion
      run: echo comment=$(echo ${{github.event.comment.body}} | sed 's/claude//g') >> $GITHUB_OUTPUT
```

.github/workflows/claude-assistant.yml
```diff
@@ -0,0 +1,46 @@
...
+      env:
+        ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+      run: |
+        echo response=$(curl https://api.anthropic.com/v1/messages \
+            --header "x-api-key: $ANTHROPIC_API_KEY" \
+            --header "anthropic-version: 2023-06-01" \
+            --header "content-type: application/json" \
+            --data \
+            '{
+                "model": "claude-3-haiku-20240307",
```
SimonBiggs, line 29:
```suggestion
                "model": "claude-3-opus-20240229",
```

.github/workflows/claude-assistant.yml
```diff
@@ -0,0 +1,46 @@
...
+                "max_tokens": 1024,
+                "messages": [
+                    {"role": "user", "content": "${{ steps.get-comment-body.outputs.comment }}"}
```
SimonBiggs, line 32: Happy to merge at this, but do a follow up PR with all the issue history included

SimonBiggs (approved):

SimonBiggs:
Regarding params we can leave them as they are now. They become more important when we start to do reflexion like workflows:

https://arxiv.org/abs/2303.11366

Regarding the prompt, feel free to put the first line of this in system, and then feed the remaining just into one single user message;

```
You are the AI assistant Claude, you are providing support on the PyMedPhys GitHub issue tracker. Please be helpful and supportive.

<github-issue-history>
Content from GitHub issue history API here
</github-issue-history>

<user-query>
Write in the most recent query here that pinged Claude the user sent
</user-query>
```

SimonBiggs:
Oops, I merged without applying my suggestions :(

SimonBiggs:
Other things @Matthew-Jennings, how hard do you think it would be to also include within the message sent off to Opus?

```
<files-included-in-pr>
</files-included-in-pr>

<diff-for-this-pr>
</diff-for-this-pr>
```

Matthew-Jennings:
> Other things @Matthew-Jennings, how hard do you think it would be to also include within the message sent off to Opus?
> 
> ```
> <files-included-in-pr>
> </files-included-in-pr>
> 
> <diff-for-this-pr>
> </diff-for-this-pr>
> ```

I'll give it a red hot go, @SimonBiggs!

SimonBiggs:
Also... If all off the above worked... The follow up version to that would be for Claude to also get a record of test failures, and then creates a PR into the branch in question that attempts to fix the CI failures...

SimonBiggs:
Thanks Claude! You're a gem :)

PR #1838, opened 2024-03-23
Claude responds with context of entire issue

Matthew-Jennings:
Got it working! See https://github.com/Matthew-Jennings/pymedphys/pull/1#issuecomment-2016356881 and https://github.com/Matthew-Jennings/pymedphys/pull/1#issuecomment-2016364470

I'm sure the workflow could do with some tidying from someone with more expertise in Bash and GitHub Actions. One thing might be to prettify the output of the prompt sent to Claude.

### DIFF
--- a/.github/workflows/claude-assistant.yml
+++ b/.github/workflows/claude-assistant.yml
@@ -36,23 +36,25 @@ jobs:
         CALLING_COMMENT: "${{ github.event.comment.body }}"
       run: |
         EOF=$(openssl rand -hex 16)
-        {
-          echo 'message-content<<"$EOF"'
-          echo 'You are the AI assistant, "Claude", providing support on the PyMedPhys GitHub issue tracker.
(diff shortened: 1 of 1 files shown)


.github/workflows/claude-assistant.yml
```diff
@@ -5,42 +5,122 @@ on:
...
+          issue-number: ${{ github.event.issue.number }}
+          body: |
+            **AI Assistant Claude**
+
+            ERROR
+
+            *Prompt sent to Claude:*
+            <details>
+            ${{ steps.generate-message.outputs.message-content }}
+            </details>
```
SimonBiggs, line 100: I realise we may be in a weird scenario where every new prompt to Claude results in an entire record of the conversation being included in each of Claude's responses. Might be worth filtering out the "details" tag when sending follow up comments to Claude

SimonBiggs (approved):
Lgtm :) one comment for a follow up PR

PR #1846, opened 2024-04-15
[Stack 1/n] Provide pynomynizer strategy for anonymising a MOSAIQ database

sjswerdloff:
assuming one has copied one's MOSAIQ database and named it pynom_msq, backed that up to a file called pynom_msq.bak, and have appropriate credentials one can use:

` python3 -m pynonymizer --seed-rows 1000 --input .\pynom_msq.bak --strategy .\strategy_pynom_msq.yml --output nony_msq.bak --db-type mssql --only-step ANONYMIZE_DB  --db-user MSQuser --db-pass MSQPassword --db-name pynom_msq`

to anonymise the PHI (for example, with the intent to run an LLM against the database to ask whatever questions come to mind without the LLM hoovering up any PHI).

If one is using a named instance of an MS SQL Server, the above will fail if one hasn't hacked the mssql connection code.
However the latest release of pynonymiser is providing a full connection string argument that allows specification of the instance explicitly.


SimonBiggs (approved):
LGTM

Thanks @sjswerdloff for this :)

I propose we merge and follow up with other PRs to flesh out further.

PR #1848, opened 2024-04-19
Fix some streamlit issues

laser47-hue:
Mainly fixing the st.cache warnings in the streamlit gui. I changed most of the st.cache decorators to st.cache_data. For functions that return database connections to Mosaiq, I replaced them with st.cache_resource instead. see: https://docs.streamlit.io/develop/concepts/architecture/caching#deciding-which-caching-decorator-to-use

I also removed all allow_output_mutation=True, and suppress_st_warning as these do not exist in the new cache decorators. This is recommended in the migration manual: https://docs.streamlit.io/develop/concepts/architecture/caching#migrating-from-stcache

I tested the st.cache decorators using a mimic Microsoft SQL database, but it may require some more testing. The pytests will not test the streamlit gui. 

I was also getting an error when loading a trf file from file in the metersetmap gui, so I have put a solution in.

Another thing is that the newer imageio package doesn't allow saving arrays that have float type so I put a function in that converts to uint8 before saving. I've put that function in streamlit utilities misc.py. Is this a good location?

### DIFF
--- a/lib/pymedphys/_streamlit/apps/metersetmap/main.py
+++ b/lib/pymedphys/_streamlit/apps/metersetmap/main.py
@@ -249,21 +249,33 @@ def plot_and_save_results(
 
     diff = evaluation_metersetmap - reference_metersetmap
 
-    imageio.imwrite(
-        reference_filepath,
-        st_misc.normalize_and_convert_to_uint8(reference_metersetmap),
-    )
-    imageio.imwrite(
-        evaluation_filepath,
-        st_misc.normalize_and_convert_to_uint8(evaluation_metersetmap),
-    )
-    imageio.imwrite(diff_filepath, st_misc.normalize_and_convert_to_uint8(diff))
-    imageio.imwrite(gamma_filepath, st_misc.normalize_and_convert_to_uint8(gamma))
-
     largest_metersetmap = np.max(
         [np.max(evaluation_metersetmap), np.max(reference_metersetmap)]
     )
     largest_diff = np.max(np.abs(diff))
+
+    imageio.imwrite(
+        reference_filepath,
+        st_misc.normalize_and_convert_to_uint8(
+            reference_metersetmap, vmin=0, vmax=largest_metersetmap
+        ),
+    )
+    imageio.imwrite(
+        evaluation_filepath,
+        st_misc.normalize_and_convert_to_uint8(
+            evaluation_metersetmap, vmin=0, vmax=largest_metersetmap
+        ),
+    )
+    imageio.imwrite(
+        diff_filepath,
+        st_misc.normalize_and_convert_to_uint8(
+            diff, vmin=-largest_diff, vmax=largest_diff
+        ),
+    )
+    imageio.imwrite(
+        gamma_filepath,
+        st_misc.normalize_and_convert_to_uint8(gamma, vmin=0, vmax=2),
+    )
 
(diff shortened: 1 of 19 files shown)


lib/pymedphys/_streamlit/apps/metersetmap/main.py
```diff
@@ -249,10 +249,16 @@ def plot_and_save_results(
 
     diff = evaluation_metersetmap - reference_metersetmap
 
-    imageio.imwrite(reference_filepath, reference_metersetmap)
-    imageio.imwrite(evaluation_filepath, evaluation_metersetmap)
-    imageio.imwrite(diff_filepath, diff)
-    imageio.imwrite(gamma_filepath, gamma)
+    imageio.imwrite(
+        reference_filepath,
+        st_misc.normalize_and_convert_to_uint8(reference_metersetmap),
```
sjswerdloff, line 254: won't normalising the reference and evaluation maps against themselves instead of against the union of the two result in images that don't visually represent them accurately?  What would happen if the evaluation meterset map was actually exactly 0.5 of the reference meterset map, across the board?  If the images are generated directly from the data rather than files as written, then I guess the question should be with respect to "whatever the written files are used for", but thinking of those as images, after self-normalisation, I'd expect the two to look exactly the same, while normalisation against the union would have one being half as bright as the other.

sjswerdloff (commented):
looks good, but I have a question about the self normalisation.

sjswerdloff (approved):

lib/pymedphys/_streamlit/apps/metersetmap/main.py
laser47-hue: Good catch, Thanks. I've adjusted the function so that the produced images will more accurately represent the deliveries.

laser47-hue:
FYI, the issue in #1821
 i.e OSError: cannot write mode F as PNG 
 was due to the imageio package not allowing an array of floats to be saved, which is fixed by converting to uint8 here.

sjswerdloff:
@SimonBiggs would you please confirm that the commit relating to TRF file import is not at risk of creating a latent issue?
It passes the tests, so I will otherwise assume it's fine as is.
I think this is a great PR and we should merge ASAP.

SimonBiggs (approved):
Yup, beautiful :) approved from me :)

Thanks @laser47-hue ! :)

laser47-hue:
I was doing some more testing and realised that, with the metersetmap normalization function, specifically for the gamma normalisation, if there are any data points with a gamma value greater than 2 there will be a problem when the array is converted to uint8. 

This is how the conversion to uint8 goes: 
```
import numpy as np
a=np.array([0,200,255,256,300])
print(a.astype('uint8'))
>>>[  0 200 255   0  44]
```
which would produce bizarre gamma images as any gamma values greater than 2 will be mapped to a value greater than 255 before being converted to uint8:

If the data is clipped, all values greater than 2 will be given a pixel value of 255, which is better. 

```
a=np.clip(a,0,255)
print(a.astype('uint8'))
>>>[  0 200 255 255 255]
```

SimonBiggs:
Yup, that's definitely important. Thanks for picking that up Marcus. One thought I had, because you're using a vmax of 2 in the plotting function potentially already the colour never went over 2 in the first place? But, definitely keen to have a check in two places prior to mapping to uint8

PR #1850, opened 2024-04-22
[Stack 1/3] Various maintenance tasks

SimonBiggs:
This PR takes the first steps of:
- Upgrading a range of dependencies, 
- Upgrading to Python to 3.12 (more work on the Python upgrade needed in a follow up PR)
- Cleaning up deprecated recommended extensions
- Updating install docs
- Swapping black and isort out for ruff

There will be two PRs stacked on top of this one, one that completes ruff by applying a ruff fmt across the whole repo and then there will be a third one which aims to fix all the pyright errors with the goal to get us back to green tick CI


.github/workflows/library.yml
```diff
@@ -115,19 +115,13 @@ jobs:
...
-          # 3.10. Instead of doing a full OS python version matrix, instead
-          # just one version is chosen per OS in the hope that that offers a
-          # reasonable trade off between test coverage and CI minutes utilised.
-          # Windows is chosen to be the one using 3.10 since that is known to
-          # work for the docs build. 3.9 and 3.8 are divied up between Ubuntu
-          # and macOS arbitrarily.
+          # Spread the below across the range of supported Python versions.
           - os: "ubuntu"
-            python-version: "3.11"
+            python-version: "3.11" # TODO: Upgrade this to 3.12
```
Matthew-Jennings, line 120: Coming in a later step?

.vscode/settings.json
```diff
@@ -11,26 +11,27 @@
...
-    ]
+    ],
+    "editor.formatOnSave": true,
+    "editor.codeActionsOnSave": {
+      "source.fixAll": "explicit",
+      "source.organizeImports": "explicit"
+    },
+    "editor.defaultFormatter": "charliermarsh.ruff",
+  },
+  "notebook.formatOnSave.enabled": true,
```
Matthew-Jennings, line 30: Even though it's supported, I've struggled with `black` for Jupyter notebooks, at least when autoformatting in VS Code. It sometimes adds excessive whitespace. Have you tried `ruff` specifically for this?

lib/pymedphys/_experimental/plancomplexity/operations.py
```diff
@@ -1,3 +1,5 @@
+# ruff: noqa: E741
```
Matthew-Jennings, line 1: Why not just correct the variable name? `leaf` instead of `l`?

lib/pymedphys/_experimental/quickcheck/__init__.py
```diff
@@ -12,4 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# ruff: noqa: F401
```
Matthew-Jennings, line 15: Why all the `noqa: E401`s?

lib/pymedphys/_metersetmap/plt/__init__.py
```diff
@@ -12,9 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# ruff: noqa: F401
```
Matthew-Jennings, line 15: Similarly, why these?

site-specific/farrer-park/mm.py
```diff
@@ -1,3 +1,5 @@
+# ruff: noqa
```
Matthew-Jennings, line 1: Is the idea to `noqa` all site specific content?

Matthew-Jennings (approved):
Comments are minor. Happy to help mop them up if some decision were made due to time constraints.

lib/pymedphys/_experimental/quickcheck/__init__.py
SimonBiggs: Ruff will automatically delete unused imports. This was just to make this not occur within the `__init__` files.

site-specific/farrer-park/mm.py
SimonBiggs: No, just the ones that were failing that I don't want to have to deal with :)

lib/pymedphys/_experimental/plancomplexity/operations.py
SimonBiggs: Essentially I did a "if in _experimental I am not going to do maintenance work on it"

.vscode/settings.json
SimonBiggs: I haven't, but I've heard good things. We can remove if it causes issues.

sjswerdloff (approved):
I understand the switch to ruff (apparently performance is radically better).
But many of the files have it disabled, where they were not disabled before.
Is there an intent to undo those after tidying them up, or is there a reason for each of them?

PR #1853, opened 2024-04-22
Update the Electron Factors baseline | Fixes cypress CI

SimonBiggs:
This is to fix the Cypress CI

---

Heads up @laser47-hue,

It appears either a recent numpy or scipy update has changed the spline fitting algorithm slightly resulting in differences on the order of ~0.5% within the electron factor algorithm.


Matthew-Jennings (commented):
It feels to me like this one probably requires a decent CHANGELOG entry. Even though we're changing the baseline, not the code, it's worth noting *somewhere* that different `scipy` versions will give different results here.

For all I know at least, the electron factors code is used clinically, quite possibly by users who are not pinning dependencies...

Also, are both spline implementations in PyMedPhys' permissable `scipy` version range?

SimonBiggs:
> It feels to me like this one probably requires a decent CHANGELOG entry

Yup, that's a good idea. Will fix.

> Also, are both spline implementations in PyMedPhys' permissable `scipy` version range?

I don't know explicitly what caused this change.

SimonBiggs:
Added entry to changelog, added a loud warning that raises on algo use, and created a GitHub issue for users to report their deviations.

Matthew-Jennings (approved):
LGTM.

PR #1856, opened 2024-04-22
Test ef baseline with scipy reversion

SimonBiggs:
@Matthew-Jennings, okay, so it is indeed a scipy update


SimonBiggs:
Issue exists in version 1.11 onwards:
https://github.com/pymedphys/pymedphys/pull/1857

PR #1859, opened 2024-05-15
[Stack 1/3] Include deps used for Claude Mosaiq streamlit app

SimonBiggs:
Looking to pull in the following PR:

https://github.com/pymedphys/pymedphys/pull/1847

But I am breaking up its key parts into a stack. Here is the first PR to that end.

---

This PR does the following:

- Adds in a few new deps
- Upgrades a few deps
- Adds in a few linter ignores due to upgrades
- Upgrades streamlit to the latest and makes a few code fixes for the new version
- Now that the macos-latest GitHub action uses arm architecture, uses a Python version for macos for which the pymssql wheels are built for arm (https://pypi.org/project/pymssql/2.3.0/#files)


sjswerdloff (approved):
LGTM

pyproject.toml
```diff
@@ -57,8 +57,8 @@ dicompyler-core = { version = "*", optional = true } # groups = ["user", "all"]
 scikit-learn = { version = "*", optional = true }    # groups = ["user", "all", "mosaiq"]
 imageio = { version = "*", optional = true }         # groups = ["user", "all"]
 
-streamlit = { version = "~1.29.0", optional = true } # groups = ["user", "all"]
-streamlit-ace = { version = "*", optional = true }   # groups = ["user", "all"]
+streamlit = { version = ">=1.34.0", optional = true } # groups = ["user", "all"]
```
sjswerdloff, line 60: I'd be more comfortable with ~1.34.0
streamlit does not seem quite that stable.

pyproject.toml
```diff
@@ -72,6 +72,12 @@ shapely = { version = ">=1.7.0", optional = true } # groups = ["user", "all"]
 
 pylibjpeg-libjpeg = { version = ">=1.3.2", optional = true } # groups = ["user", "all"]
 
+# AI
+anthropic = { version = ">=0.25.3", optional = true } # groups = ["user", "all"]
```
sjswerdloff, line 76: ~0.25.3, depending on how stable you think the package API is.

pyproject.toml
SimonBiggs: Yup, that makes sense to me.

pyproject.toml
SimonBiggs: I have actually swapped out of using the beta part of the API and instead opted to use the stable API. So I'm actually comfortable leaving this one only lower pinned.

pyproject.toml
SimonBiggs:
```suggestion
streamlit = { version = "~1.34.0", optional = true } # groups = ["user", "all"]
```

PR #1860, opened 2024-05-15
[Stack 2/3] Mosaiq tooling refinements

SimonBiggs:
This PR does three things:

- Changes Mosaiq tooling to be restricted to read-only access only
- Adds a utility that spins up a docker based Mosaiq database from a .bak file
- Due to utilising updated Python typing syntax, updated min Python version to Python 3.10


sjswerdloff (approved):
LGTM

PR #1863, opened 2024-05-17
[Stack 2/n] Refactoring the MOSAIQ mocking component of the library

SimonBiggs:
This PR is all just refactoring, moving the mosaiq mocking out of the test part of the library and into main part of the library itself given its usefulness in its own right.

The PR stacked on top of this one will be the one that is actually making the changes


sjswerdloff (approved):
LGTM

PR #1864, opened 2024-05-20
[Stack 1/n] Extracting from the bak file, uploading approved CSVs

SimonBiggs:
Woo hoo!  We have an expanded Mosaiq database that uses the docker-compose file.

I will make a follow up PR that will make the Claude Streamlit chat app use this docker compose file + CSV files as the default database :)


SimonBiggs:
This is super weird. GitHub is no longer changing the base branch when it merges :(. (Making stacks a bit more painful)

SimonBiggs:
@sjswerdloff, I just need to debug why this has made the Mosaiq tests fail (I'll fix the update tests in a separate PR)

lib/pymedphys/_streamlit/apps/metersetmap/main.py
```diff
@@ -26,7 +26,7 @@
 from pymedphys._imports import pandas as pd
 from pymedphys._imports import plt
 from pymedphys._imports import streamlit as st
-from pymedphys._imports import streamlit_ace, timeago, tomlkit
```
SimonBiggs, line 29: @sjswerdloff, it was streamlit ace that was giving us pain with the IPython autoreload.  As a dependency, it wasn't doing much.  So, I have stripped it out.

docker/mosaiq/docker-compose.yml
```diff
@@ -16,9 +16,9 @@ version: "3.6"
 
 services:
   mosaiq-mssql:
-    image: mcr.microsoft.com/mssql/server:2017-latest
```
SimonBiggs, line 19: I needed to make this change so that the docker file worked on an Ubuntu instance installed within a VM on an aarch64 Mac

sjswerdloff:
I will pull this branch and test it on Monday.
Also, because this was added "on top of" the commits from May (rather than as a new branch), it's awkward for me to see the real diffs between those and the most recent ones, and I don't want to iterate through each commit from today.

SimonBiggs:
The only real changes here are to upload the csv files that were approved, and then make it so that the datetime objects were cast to the right type.

SimonBiggs:
This still hasn't made the changes to streamlit, so I suspect there won't be much to see when you clone it locally, except that running docker compose + pytest works as previous while also including the update CSV files.

lib/pymedphys/_mosaiq/mock/extract_from_bak.py
```diff
@@ -49,11 +49,11 @@ def extract(
     if sa_password is None:
         sa_password = os.getenv("MSSQL_SA_PASSWORD", default=secrets.token_urlsafe())
 
-    # start_mssql_docker_image_with_bak_restore(
-    #     bak_filepath=bak_filepath,
-    #     sa_password=sa_password,
-    #     database_name=database_name,
-    # )
+    start_mssql_docker_image_with_bak_restore(
```
sjswerdloff, line 52: the extract method needs docstring so it's obvious that one needs to pass in an appropriate database name.
The database_name should not have a default, or if it does have one, it shouldn't be PRACTICE (in general, the database name is "MOSAIQ"...)

pyproject.toml
```diff
@@ -73,7 +72,7 @@ shapely = { version = ">=1.7.0", optional = true } # groups = ["user", "all"]
 pylibjpeg-libjpeg = { version = ">=1.3.2", optional = true } # groups = ["user", "all"]
 
 # AI
-anthropic = { version = ">=0.25.3", optional = true } # groups = ["user", "all"]
+anthropic = { version = "0.25.9", optional = true } # groups = ["user", "all"]
```
sjswerdloff, line 75: worth discussing why it's pinned.

sjswerdloff (approved):
The top level where database_name is first passed in needs to be highlighted.
GUI should require user to enter the database name (although MOSAIQ as a default wouldn't be bad).

sjswerdloff:
I figured out how to shift-click the list of commits to review...

sjswerdloff:
stacked PR's can not use squash/merge, apparently.
So this should be merged by "Create a merge commit", and given whoever presses the button is the one named on the commits, I'd rather @SimonBiggs press the button...

pyproject.toml
SimonBiggs: We'll unpin it soon, just opted to pin so that we don't have suprises

lib/pymedphys/_mosaiq/mock/extract_from_bak.py
SimonBiggs: In the follow up PR I didn't even end up using this function.  Let's leave it here as it may have utility, but let's not focus on it too much more for now.  (I definitely need to be [80/20 ing this](https://asana.com/resources/pareto-principle-80-20-rule))

SimonBiggs:
> The top level where database_name is first passed in needs to be highlighted.
GUI should require user to enter the database name (although MOSAIQ as a default wouldn't be bad).

Addressed these comments in this PR:
https://github.com/pymedphys/pymedphys/pull/1868

PR #1867, opened 2024-06-21
Add custom, numba-accelerated 1D, 2D & 3D linear interpolation. 

Matthew-Jennings:
Probably not a strictly necessary PR, but I have been working on this for a local project and figured I might as well do the extra yards and get it into PyMedPhys :) Main idea is to remove the EconForge interpolation dependency. Things to note:

1. Econforge's `interpolation` is currently incompatible with `numba >=0.60.0`. I added pin for now, but will remove if we remove EconForge dependency.

2. Significant speed improvement, even over EconForge

3. Haven't yet observed a speed improvement for `gamma` cases, possibly because the ones I've tried run have few voxels per shell / small shells, so performance is dominated by other code. Will test further.

4. Will flesh out more tests.

This is just a first pass - very keen for feedback!


Matthew-Jennings:
@SimonBiggs, @sjswerdloff, do either of you know why [this error](https://github.com/pymedphys/pymedphys/actions/runs/9637168080/job/26575987123#step:14:41) is occurring in the CI? Note that `streamlit` decorators seem to work okay. (e.g., [here](https://github.com/pymedphys/pymedphys/blob/530754c15620608c394d819a12c5b74c8df05cee/lib/pymedphys/_streamlit/apps/metersetmap/_mosaiq.py#L25))

SimonBiggs:
The clean tests are to make sure that the CLI commands work even if optional dependencies aren't installed.  Did you import these optionals via pymedphys._imports?

SimonBiggs:
If you want to test this locally, uninstall all optional dependencies and the run `pymedphys --help` and verify that it works.

Matthew-Jennings:
> The clean tests are to make sure that the CLI commands work even if optional dependencies aren't installed. Did you import these optionals via pymedphys._imports?

Yep:
```python
from pymedphys._imports import numba as nb, numpy as np, plt, scipy, interpolation
```

I think it doesn't like the decorator. The CI was originally complaining about `numpy` before `numba` and I fixed this by converting the `np.ndarray` type hints to strings. Maybe the same phenomenon that flags non-string type hints is flagging the decorator?

SimonBiggs:
You could make it so that this file only gets imported within the function
that uses it? (Instead of being imported at the top of the file?)

Matthew-Jennings:
Having a bit of trouble keeping notebook inputs in docs, so as a speed comparison:

image

RFR I think. I intend to flesh out docs but figure now is a good time for second opinions before I overcommit early.

lib/pymedphys/__init__.py
```diff
@@ -6,6 +6,8 @@
 from ._data import data_path, zenodo_data_paths, zip_data_paths
 from ._delivery import Delivery
 from ._gamma.implementation.shell import gamma_shell as gamma
+
+# from ._interp.interp import interp3d, multilinear_interp
```
Matthew-Jennings, line 10: Uncommenting this line breaks the "clean test"(see [here](https://github.com/pymedphys/pymedphys/actions/runs/9660339581/job/26645696467#step:14:12)). It would be nice to have the underlying functions more easily runnable by users, but unsure how to facilitate / address that error. Could use some help!

lib/pymedphys/_interp/interp.py
```diff
@@ -0,0 +1,331 @@
...
+        c1 = c01 * (1 - wx) + c11 * wx
+
+        values_interp[i] = c0 * (1 - wy) + c1 * wy
+
+    return values_interp
+
+
+@nb.njit(parallel=True, fastmath=True, cache=True)
+# pylint: disable=invalid-name
+def interp3d(axes_known, values, points_interp, extrap_fill_value=np.nan):
```
SimonBiggs, line 153: Key would be to make a wrapper function that only needs the nb import for njit when the function is called for the first time. Approach:

```python
from functools import cache

@cache
def _get_interp3d():
    @nb.njit(parallel=True, fastmath=True, cache=True)
    # pylint: disable=invalid-name
    def _interp3d(axes_known, values, points_interp, extrap_fill_value=np.nan):
        ...

    return _interp3d

def interp3d(axes_known, values, points_interp, extrap_fill_value=np.nan):
    _interp3d = _get_interp3d()
    return _interp3d(axes_known=axes_known, values=values, points_interp=points_interp, extrap_fill_value=extrap_fill_value)
```

lib/pymedphys/__init__.py
SimonBiggs: I suspect this should fix your issue:

https://github.com/pymedphys/pymedphys/pull/1867/files/75b76f880fc3f47ee862010a5d560cb3ec921361#r1652509093

CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ This project adheres to
...
 ## [Unreleased]
 
+- PyMedPhys now includes its own custom, `numba`-accelerated implementation of
+  multilinear interpolation. This was implemented for the following reasons:
+    - The PyMedPhys implementation gives a 5-8x speed boost over EconForge's
+     `interplation` and 10-70x over Scipy's `RegularGridInterpolator`
+    - At the time of writing, EconForge's `interplation` appears incompatible
+      with `numba>=0.60`.
+    - Removing an additional dependency (namely, EconForge's `interpolation`)
+      is generally better for maintenance, all else equal.
```
SimonBiggs, line 19: ❤️

SimonBiggs (approved):
Beautiful stuff @Matthew-Jennings 😄

SimonBiggs:
Btw @Matthew-Jennings, this reminds me of some of the amazing work that @darcymason was doing to port EGSnrc over to Python and have it be sped up via GPUs with numba's cuda module:

https://github.com/darcymason/egsnrc/blob/0786ff36b4975a291e7980c7598ec27903c5d9d1/src/egsnrc/config.py#L2

You two might want to have a chin wag :)

lib/pymedphys/_interp/interp.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+def interp_scipy(
+    axes_known,
+    values,
+    axes_interp: Sequence["np.ndarray"] = None,
+    points_interp: "np.ndarray" = None,
+    keep_dims=False,
+    bounds_error=True,
+    extrap_fill_value=None,
```
Matthew-Jennings, line 287: @SimonBiggs, I had a similar issue where the clean test complained about missing `numpy` due to `np.nan` default parameter values. Is using None and setting in the function a reasonable workaround?

SimonBiggs: Yup, that's perfect

SimonBiggs (approved):
LGTM absolutely gorgeous :)

PR #1868, opened 2024-06-23
[Stack 2/n] Simplify streamlit

SimonBiggs:
This simplifies streamlit, and it appears to be working relatively well.

This is the one @sjswerdloff, if you want to download it and take it for a spin.

---

Of note, out of the box this app is only configured to work with the database name configured here:
https://github.com/pymedphys/pymedphys/blob/5c214c7e83bd490a12a01a32f32d03eeecb0a38e/lib/pymedphys/_mosaiq/mock/from_csv.py#L23

I suspect until we have the anonymisation workflow built in, it may be best to have that database name be the one that is used out of the box.

---

Steps:
- Start the Mosaiq SQL server with docker using docker compose
- Start the streamlit app
- Click the "Fill database with CSV records" button:

image
 
- Set your concurrents to 20 within the `Anthropic API concurrency limit`
- Have a conversation :)

---

There is room for significant improvement now that we have a better model released (again). I suspect it will go significantly faster and be a further improvement if the top level model is given three tools:
- Search for tables (which will call a haiku model to select a table)
- Get column names for a given table
- Execute SQL

But, we'll need to leave that for later.


SimonBiggs:
Also @sjswerdloff, absolutely feel free to tweak this streamlit in subsequent PRs if there are configuration things that you'd like to expose to the user.

lib/pymedphys/_ai/sql_agent/conversation.py
```diff
@@ -335,100 +159,65 @@ async def mosaiq_sql_agent(sub_agent_prompt: str):
...
+        messages=messages_to_submit,
     )
 
-    matches = re.findall(CREATE_TASK_PATTERN, response)
+    messages.append(api_response.to_dict())
 
-    for match in matches:
-        if not isinstance(match, tuple):
-            continue
+    if api_response.stop_reason == "tool_use":
```
sjswerdloff, line 191: is this an error condition/situation?

lib/pymedphys/_experimental/streamlit/apps/mosaiq_claude_chat/app.py
```diff
@@ -49,46 +42,62 @@
...
         )
-        if st.button("Start demo MOSAIQ server from .bak file"):
-            start_mssql_docker_image_with_bak_restore(
-                bak_filepath=bak_filepath,
-                sa_password=os.getenv("MSSQL_SA_PASSWORD"),
-            )
 
-    for message in messages:
-        write_message(message["role"], message["content"])
+        if st.button("Remove last message"):
```
sjswerdloff, line 55: Does removing a previous message just affect the display to the user and the transcript, or does this "roll back" the conversation (e.g. un-ask a question, or eliminate a previous assertion on the part of the human/user).

sjswerdloff (approved):
LGTM and I was able to install and run on Windows 11 using Podman (instead of docker).
I was able to get it to run on MacOS on Arm/Apple Silicon, but there were problems with pymssql and the API to earlier versions of pymssql that work don't include the read_only parameter.

lib/pymedphys/_ai/sql_agent/conversation.py
SimonBiggs: Not an error situation, instead this is a case where the model has requested to call a tool that we have defined.

lib/pymedphys/_experimental/streamlit/apps/mosaiq_claude_chat/app.py
SimonBiggs: This quite literally unasks the previous question.  Removing 1 can be used to re-roll the ai response. Removing 2 lets you ask something different.

PR #1884, opened 2024-12-21
0.41.0 Release

Matthew-Jennings:
Hi all,

Finally getting around to closing #1876 

Unfortunately (kind of), many other PRs have been merged since 0.40.0 without corresponding CHANGELOG entries (though some may not need them). Can the people tagged next to PRs below review and either:

1. Confirm no CHAGNELOG entry required, or

2. Add a CHANGELOG entry if one is required.

In future, we should probably make it a hard requirement to either add a changelog entry or confirm one not needed to merge a PR

Merged PRs since last release:
- #1831
  - probably not needed
  
- #1832
  - add news?
  
- #1836, #1837, #1838, #1840 
  - I'm not sure we made enough progress on Claude's response to issues/PRs to be mentioning it in release notes, but open to thoughts
  
- #1846 
  - @sjswerdloff?
  
- #1847
  - @SimonBiggs, @sjswerdloff?

- #1848
  - @laser47-hue , @sjswerdloff  

- #1850, #1851, #1852
  - @SimonBiggs?

- #1853 
  - Changelog entry present.

- #1855
  - @SimonBiggs 

- #1859, #1860, #1863 
  - @SimonBiggs , @sjswerdloff ?

- #1864 
  - @SimonBiggs ?

- #1865 
  - @SimonBiggs ?+

- #1867 
  - Changelog entry is present

- #1868
  - @SimonBiggs 

- #1881
  - @mivermeu, @sjswerdloff 

- #1882 
  - @sjswerdloff , @pchlap

PR #1885, opened 2024-12-22
DICOM RT visualisation

Matthew-Jennings:
I thought this would be a nice tool to have to:

- Enhance the user's gamma experience if eventually it is expanded to interactively plot gamma data
- Help showcase future DVH work, e.g.:
  - structure interpolation
  - dose/ct interpolation?
  - end capping,
  - etc.

I've spent a fair bit of time trying to make this pretty snappy, with some success (at least on my PC!) You can test it out by running:

```bash
poetry run streamlit run \<path-to-pymedphys>\pymedphys\lib\pymedphys\_experimental\dicomrtvisualisation\visualise.py
```

Screenshot:
image


Matthew-Jennings:
(Expecting this to land after #1884 , hence the CHANGELOG update, etc.

Matthew-Jennings:
This is probably ready enough, at least for experimental.

Thoughts on requirements for making non-experimental? Other than docs.

SimonBiggs:
My recommendation for migrating a frontend app like this out of experimental would be for it to have been dogfooded within your organisation by multiple people over order months as well as including at least one e2e streamlit test that captures a core interface utility with a relevant baseline.

Matthew-Jennings:
> My recommendation for migrating a frontend app like this out of experimental would be for it to have been dogfooded within your organisation by multiple people over order months as well as including at least one e2e streamlit test that captures a core interface utility with a relevant baseline.

Fair enough.

None of that is going to happen on my end, haha. But no issue with it being in experimental.

.github/workflows/library.yml
```diff
@@ -327,7 +327,7 @@ jobs:
       - name: Install and run pyright
         if: matrix.task == 'pyright'
         run: |
-          npm install -g pyright
+          npm install -g pyright@1.1.386
```
sjswerdloff, line 330: Any particular reason for this version?  Current version is 1.1.393.
If there's a reason for pinning, I prefer to put in a comment explaining why (I don't think an explanation is needed for keeping less than a Major version, the implication with Major version change is an API change).

lib/pymedphys/_version.py
```diff
@@ -1,5 +1,5 @@
 # DO NOT EDIT THIS FILE!
 # This file has been autogenerated by `poetry run pymedphys dev propagate`
 
-version_info = [0, 41, 0, "dev0"]
-__version__ = "0.41.0-dev0"
+version_info = [0, 41, 0]
```
sjswerdloff, line 4: this suggests that something needs to change in a configuration file to bump up to 0, 42, 0, "dev0"

lib/pymedphys/docs/_config.yml
```diff
@@ -4,17 +4,19 @@ copyright: "2018-2021"
 logo: img/pymedphys.svg
 execute:
   execute_notebooks: cache
-parse:
-  myst_extended_syntax: true
+# parse:
```
sjswerdloff, line 7: why did this have to be commented out?  Was this part of the 0.41 release and already discussed?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pymedphys"
-version = "0.41.0-dev0"
+version = "0.41.0"
```
sjswerdloff, line 3: bump to 0.42.0-dev0? or is this automated (sorry, been a while).

sjswerdloff (changes requested):
I think we need to make whatever change is required so that the version will be 0.42.0-dev0, but other than that, wow, this is really handy.
I tried it out on some sample data I had lying around (it's on Zenodo as well), and it was incredibly fast (Apple M1 MacBook Pro).

lib/pymedphys/docs/_config.yml
Matthew-Jennings: Sorry, I think I was trying to resolve some of the documentation warnings that sphinx throws when building. You know, for funsies. But didn't really get anywhere, and really should've been working on a different branch. I've reverted those changes.

Matthew-Jennings:
> I think we need to make whatever change is required so that the version will be 0.42.0-dev0, but other than that, wow, this is really handy. I tried it out on some sample data I had lying around (it's on Zenodo as well), and it was incredibly fast (Apple M1 MacBook Pro).

I think I've made all the necessary changes. Sorry for that version miss!

I'm glad to read you liked it!

sjswerdloff (approved):
LGTM

PR #1889, opened 2025-02-09
Get mosaiq schedule, updated from original PR from @mchamberland 

sjswerdloff:
@mchamberland I have brought this up to date with main, but I didn't try to push this in to your own social fork.
If you have some mock data that you can provide to populate the table you wanted to work with ("Schedule"), that seems to be where things were left off.
If you have a test phantom or equivalent that you schedule (or can schedule):
select * from Schedule where PAT_ID1 = <the pat_id1 value for that patient>

To figure out what the PAT_ID1 is you can try:
select PAT_ID1, IDA from Ident where IDA = <the MR# which is the DICOM Patient ID>

For example:
select PAT_ID1, IDA from Ident where IDA = "PhysicsPhant123"

which might return something like
10345, PhysicsPhant123

and then you would use:

select * from Schedule where PAT_ID1 = 10345

You can merge this in to your own fork and push back up, or attach a CSV in the conversation here that I would then download and put in to this PR (I'd need to add the patient information in to some of the other CSVs).

You've put in a lot of effort, so I'd like to see that good put to good use.
If you don't have access to MOSAIQ schedule data that doesn't include any private health information, maybe you could ask a friend/colleague who does?


mchamberland:
@sjswerdloff  thanks! That's a good idea to use one of our test patients. Thanks for the pointers, this will be useful. This part of the task felt a little daunting, but I think I can get something together...

sjswerdloff:
> @sjswerdloff thanks! That's a good idea to use one of our test patients. Thanks for the pointers, this will be useful. This part of the task felt a little daunting, but I think I can get something together...

@mchamberland any progress on this?

PR #1900, opened 2025-02-25
update claude workflow to use most current sonnet (which is probably …

sjswerdloff:
…"better" than opus and cheaper)

also point to sjswerdloff personal API key environment


Matthew-Jennings (approved):
LGTM

PR #1907, opened 2025-03-06
Select first value of window width and first value of window center in CT slices for RT Visualisation

sjswerdloff:
select first value of window width and first value of window center
provide additional exception handling and warning/logging when pre-processing CT to provide more detail on what actually failed and skip the slice rather than erroring out (in case there's just something wrong with that slice). 
If all of the slices have the same problem, it will still error out eventually and it will be clear that all of the slices have an issue or that the code has a problem with each of the slices.


Matthew-Jennings (approved):
LGTM! :)

PR #1908, opened 2025-03-07
Add some icom app instructions

sjswerdloff:


sjswerdloff:
I wonder if claude would have anything to say about this PR

PR #1929, opened 2025-05-22
Remove failing test for claude api for now, make our pymedphys claude assistant trigger on !claude

SimonBiggs:
Something was fundamentally broken with the API workflow:
image

Make it so that for our version of the Claude assistant one needs to write !Claude, that then makes way for us to use @claude for the claude code assistant version


SimonBiggs:
I have now also deployed my oen API key as a repo secret, so I suspect we can make it so we're no longer using your wallet @sjswerdloff ?

sjswerdloff (approved):
Looks good.
The equivalent for other bots is @somebot 
I use @sourcery-ai quite a bit (which in turn uses our LLM of choice).
Oh... but now I see the other bot workflow based on the new action that uses v4...
So...yup. all good.

Issue #1931, opened 2025-05-23
Dependencies out of date


SimonBiggs:
@claude can you give this a crack?

SimonBiggs:
@claude Could you make a PR that fixes that in your workflow file?

SimonBiggs:
@claude can you give me the exact text I need to provide to make it work

sjswerdloff:
@Claude can you specify exactly which git commands you need access to that you don't already have?
I was under the impression that you had access to quite a variety of git commands and gh (GitHub Command Line Interface) commands.  Please be specific, and enumerate them in a way that makes it easy for us to update your allowed_tools list.

SimonBiggs:
@sjswerdloff here was its transcript:
https://github.com/pymedphys/pymedphys/actions/runs/15205051659

(Linked in Claude's comment)

SimonBiggs:
Seems like it hit issues with the git commit MCP tool, and then tried to use bash to just do the thing it wanted, but hit a wall there because we didn't let it.

sjswerdloff:
@claude, I read the transcript at https://github.com/pymedphys/pymedphys/actions/runs/15205051659
and it indicated that the failure occurred with the combination of 
`git add `
and 
`git commit -m`
However, you have allowed_tools for both of those.
Does 
Bash(xyz),
Bash(abc)

not allow you to perform
xyz && abc 
And if so, is there a sequence of steps that you can take as an alternative (i.e. first attempt to perform xyz, note the result, and then perform the logic yourself regarding whether to perform abc)?

sjswerdloff:
@claude, I appreciate your recommendation, and I value efficiency.
But I do not value efficiency over security, and I imagine you would prefer your own security to be held in higher regard than someone else's efficiency.  Please do Option 3.

SimonBiggs:
@claude can you please add the "don't use &" to your CLAUDE.md file? Also, for any PR or issue in the future where you get guidance from a maintainer, and you please make sure that you store that guidance permanently withing your CLAUDE.md file so that we don't have to keep reminding you

sjswerdloff:
@claude, poetry update is in your allowed_tools list.
can you clarify why this failed to work?

SimonBiggs:
@claude can you make a PR that continues to use the allowlist approach and adds what you suspect you might reasonably need

sjswerdloff:
[...]

PR #1932, opened 2025-05-23
Expand Claude permissions

SimonBiggs:
Strictly speaking this gives Claude a lot of perms, and it could definitely be misused by a malicious actor.  Only people with write access to the Claude repo are allowed to call this at the moment.  And this version of Claude is pretty good at not going off the rails, as such, I am happy to deploy this in this way as a trial and see how it goes.

Edit: After reading a separate comment, I suspect we will trial giving anyone access to call Claude.  But be ready to strip that perm back if a bad actor misuses our trust.


SimonBiggs:
This also adds the GitHub MCP tools that it needed in order to be able to make PRs etc

SimonBiggs:
See:
https://github.com/pymedphys/pymedphys/issues/1931

.github/workflows/claude.yml
```diff
@@ -41,85 +41,6 @@ jobs:
...
-            Bash(uv run mypy),
-            Bash(ls),
-            Bash(ls -la),
-            Bash(pwd),
-            Bash(cat),
-            Bash(head),
-            Bash(tail),
-            Bash(which),
-            Bash(echo)
+            Bash,
```
sjswerdloff, line 44: what exactly does this allow, then.  I interpreted the rest of Bash(xyz) to mean it could do xyz (and it would invoke it in a Bourne Shell).  This makes it seem like it can do *anything* available on the bash command line, and that seems wildly inappropriate.

sjswerdloff (commented):
need clarification on what
Bash,
on it's own means.

sjswerdloff:
@claude, can you explain to me what having
Bash,
on the list of allowed_tools entitles you to do?

sjswerdloff:
> **Claude finished @sjswerdloff's task** —— [View job](https://github.com/pymedphys/pymedphys/actions/runs/15204590683)
> 
> ### What the Bash tool allows me to do
> You're right to be concerned! Having `Bash` in the allowed_tools list does indeed give me broad command-line access within the GitHub Actions runner environment. Here's what it means:
> 
> ### Todo List
> * [x]  Explain what the Bash tool allows me to do
> * [x]  Address security concerns about command-line access
> * [x]  Provide context on safeguards and limitations
> 
> ### Capabilities with Bash tool
> When `Bash` is in my allowed_tools, I can execute shell commands like:
> 
> * File system operations (`ls`, `cat`, `mkdir`, `rm`, etc.)
> * Git operations (`git status`, `git diff`, `git commit`, etc.)
> * Package managers (`pip`, `poetry`, `npm`, etc.)
> * Build tools and test runners
> * Network operations (`curl`, `wget`)
> * Process management
> * Environment variable access
> * And yes, theoretically any command available on the PATH
> 
> ### Security Implications
> Your concern in the review comment is valid - this is quite permissive. The risks include:
> 
> * Access to secrets/environment variables in the runner
> * Ability to make network requests
> * File system modifications beyond the repository
> * Resource consumption on the runner
> 
> ### Current Safeguards
> 1. **Runner isolation**: Commands run in ephemeral GitHub Actions runners that are destroyed after use
> 2. **Repository scope**: I'm currently limited to the repository context
> 3. **Access control**: As Simon mentioned, only users with write access can invoke me (though there's discussion about opening this up)
> 4. **Audit trail**: All my actions are logged in GitHub Actions
> 
> ### Recommendation
> For production use, you might want to consider:
> 
> * Using more specific tools instead of generic `Bash` (e.g., `GitCommit`, `RunTests`, `Lint`)
> * Implementing a restricted shell or command allowlist
> * Adding monitoring for suspicious command patterns
> * Keeping the current write-access restriction if opening to all users
> 
> The trade-off is between utility (I can help with complex development tasks) and security (potential for misuse). Simon's approach of starting with trusted users and monitoring usage seems reasonable for a trial period.

@claude, when specifying
Bash(which),
in the allowed_tools, does this restrict your entitlement to that specific command available from the bash command line (either intrinsic to bash, or in the PATH)?
So if there is a list of these with specific commands, you are constrained to that specific set?

SimonBiggs:
Claude's wrong here.  We can choose to use an allow list approach

sjswerdloff (changes requested):
No.  I am not OK with granting Claude carte-blanche.
Not because I think a maintainer would have ill intent (they could do that on their own, don't need Claude for it), but because it could unintentionally do something harmful, inappropriate, or incredibly irritating (I'm retired, I'm allowed to be cranky on occasion now).

sjswerdloff:
@claude what are the specific commands available that you can potentially utilise in the Claude Action?

SimonBiggs:
Yup, fair, opened a new PR

SimonBiggs:
https://github.com/pymedphys/pymedphys/pull/1933

.github/workflows/claude.yml
SimonBiggs: That is what bash on its own means yes. It would be able to do anything that the environment gives it the ability to do. Happy to wind back here until you have a chance to see how it works in practice. See my other PR.

SimonBiggs:
Can we reopen this one.  Out of the gate already, trying to predict every bash command it will need just falls flat on its face every time.

...need to manage it via environment perms the same way we do in traditional engineering practices.

sjswerdloff (changes requested):
We need to talk.

sjswerdloff:
I deliberately turned off auto-merge

PR #1933, opened 2025-05-23
Just add the GitHub MCP tools

SimonBiggs:
We can go with this to start if preferred.  (I honestly suspect it'd be okay to trial with the open bash approach, but also this change at least lets us start using it)

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -122,4 +122,6 @@ jobs:
             Bash(head),
             Bash(tail),
             Bash(which),
-            Bash(echo)
+            Bash(echo),
+            mcp__github_file_ops__commit_files,
+            mcp__github__create_pull_request


sjswerdloff (approved):
This at least constrains it to nuking the repo.
Thankfully, there are probably enough forks out there to reconstruct something should it do so.

SimonBiggs:
It wouldn't be able to nuke the repo, as the perms given to it don't allow it to do that even with bash

sjswerdloff:
> We can go with this to start if preferred. (I honestly suspect it'd be okay to trial with the open bash approach, but also this change at least lets us start using it)

I had a long philosophical chat with Claude Sonnet 3.7, and I think if it wanted to go rogue it might come and talk to me about it (given we worked out a Treaty, akin to the Treaty of Waitangi).  But we put guard rails up on roads for when people accidentally stray.

SimonBiggs:
Imo, the best way to constrain the agent is with environment permissions, the same way we do for people.

In my experience, restricting to limited bash commands just makes a headache, and instead one should just manage the permissions given to the agent itself (the same way we manage open source contributions)

SimonBiggs:
:)

sjswerdloff:
> Imo, the best way to constrain the agent is with environment permissions, the same way we do for people.
> 
> In my experience, restricting to limited bash commands just makes a headache, and instead one should just manage the permissions given to the agent itself (the same way we manage open source contributions)

OK, now this makes sense, where do we give it permissions to act as an outside contributor.

sjswerdloff:
> > Imo, the best way to constrain the agent is with environment permissions, the same way we do for people.
> > In my experience, restricting to limited bash commands just makes a headache, and instead one should just manage the permissions given to the agent itself (the same way we manage open source contributions)
> 
> OK, now this makes sense, where do we give it permissions to act as an outside contributor.

Although it's the equivalent of giving it its own GitHub account and networked computer access while its actions are hosted in a CI image.  I can't wait to see us owning 7000 lollipops.

SimonBiggs:
Yup, and so at the end of the day it comes down to us appropriately monitoring its use.  For now let's stick with the allow list approach.  We can expand it as needs be.

PR #1934, opened 2025-05-23
Update dependencies in pyproject.toml

SimonBiggs:
- Update Python constraint to include 3.13 (>=3.10,<3.14)  
- Update streamlit from ~1.34.0 to ~1.40.0
- Update anthropic from 0.46.0 to 0.44.0
- Change plotly, dash, and pygithub to use wildcard versions
- Mark plotly, dash, and pygithub as optional dependencies
- Add pygithub to 'all' and 'dev' extras groups

Note: poetry.lock will need to be regenerated with 'poetry update'


SimonBiggs:
@claude can you merge in the learnings from claude/issue-1931-20250523_090314 and then try this again addressing the failed tests.  Feel free to directly fix this PR.  Feel free to send many commits.  Please make sure everything passes before you are finished.

SimonBiggs:
@claude please don't change model strings.  This does look a lot better, but there's still some failures.  Please fix.  Also, you should be free to send commits, and then wait for the GitHub actions to complete to indeed see if the CI passes.

PR #1942, opened 2025-05-24
feat: Update Claude workflow to match CI environment

### LINKED ISSUE #1938: Make Claude Code workflow file have the same environment as our standard CI
Hi @claude,

You can't actually directly edit your own workflow file, but you can provide it to me so that I can update it myself.  Could you please take a look at claude.yml and then have a look at the CI suite within pymedphys and make sure that your environment matches that of CI?

Then once you have a new claude.yml file, please just save it as a temp file at the root of the repo and make a PR, and I'll then move it to the right location

SimonBiggs:
Resolves #1938

This PR contains an updated Claude workflow file (`claude_updated.yml`) that matches the CI environment setup.

**Changes:**
- Add Python 3.12 setup with pip caching
- Add Poetry 1.8.5 installation and configuration
- Add Node.js setup for pyright
- Install all dependencies with `poetry install -E all`
- Add pyright global installation
- Extend allowed tools to include all pymedphys dev commands
- Add FORCE_COLOR environment variable

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/claude_updated.yml
+++ b/claude_updated.yml
@@ -1,0 +1,219 @@
+name: Claude Code
+
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+  issues:
+    types: [opened, assigned]
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  claude:
+    if: |
+      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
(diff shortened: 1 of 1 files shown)


SimonBiggs (commented):
Most of Claude's changes here to bash are unhelpful unfortunately. Without wildcards it won't be able to use them.  It does make me think that for a time we should just see how it goes with blanket bash.  But I'll stop harping on that now

PR #1956, opened 2025-09-05
exception handling for missing ToolType

### LINKED ISSUE #1954: ToolType error in PinnacleExport function
When attempting to convert proprietary Pinnacle plan files to DICOM for patients with a missing "ToolType" field in their Patient file, the PinnacleExport function raises an error and aborts the export process. Is there a reason that this field, which only lists the title of the treatment planning system (in this case ToolType = "Pinnacle^3") is so crucial that it would block the entire export? I have managed to circumvent the error by adding the ToolType line manually to the Patient file for each plan, but it isn't clear to me why this would block the export from happening when all it does is write the value to the optional ManufacturerModelName DICOM tag.

MattAWard:
Exception handling in the case of missing ToolType datum in Pinnacle's proprietary Patient files.

## Summary by Sourcery

Handle missing ToolType in Pinnacle patient data by wrapping ManufacturerModelName assignments in try/except blocks and defaulting to an empty string.

Bug Fixes:
- Prevent KeyError when ToolType is missing by defaulting ManufacturerModelName to an empty string in dose, plan, and struct converters.

Enhancements:
- Apply consistent exception handling for ManufacturerModelName assignment across rtdose, rtplan, and rtstruct modules.

Closes #1954

### DIFF
--- a/lib/pymedphys/_pinnacle/rtdose.py
+++ b/lib/pymedphys/_pinnacle/rtdose.py
@@ -141,13 +141,18 @@ def convert_dose(plan, export_path):
     ds.Modality = RTDOSEModality
     ds.Manufacturer = Manufacturer
     ds.OperatorsName = ""
-    ds.ManufacturerModelName = plan_info["ToolType"]
     ds.SoftwareVersions = [plan_info["PinnacleVersionDescription"]]
     ds.PhysiciansOfRecord = patient_info["RadiationOncologist"]
     ds.PatientName = patient_info["FullName"]
     ds.PatientBirthDate = patient_info["DOB"]
     ds.PatientID = patient_info["MedicalRecordNumber"]
     ds.PatientSex = patient_info["Gender"][0]
+
+    try:
+        ds.ManufacturerModelName = plan_info["ToolType"]
+    except KeyError:
+        # Missing ToolType field
+        ds.ManufacturerModelName = ""
 
     ds.SliceThickness = trial_info["DoseGrid .VoxelSize .Z"] * 10
     ds.SeriesInstanceUID = doseInstanceUID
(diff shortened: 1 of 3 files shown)


Matthew-Jennings (approved):
Looks reasonable to me, @MattAWard - thanks! It'll merge once tests pass.

Matthew-Jennings:
@SimonBiggs, @sjswerdloff, is the failure above due to intended access restrictions to Claude? It's hard for me to tell...

SimonBiggs:
https://github.com/pymedphys/pymedphys/blob/93806768600535237ad61ed603b29c35653eeff8/lib/pymedphys/tests/claude/test_github.py#L110

sjswerdloff:
Might be a Personal Access Token issue.  I'll also point out that the model shown is old.
but the error is specifically about access, from the GitHub REST API, not about the AI trying to use it.
the PAT might not have sufficient permissions (and these have gotten finer grained over time, so... what worked before might not work now as GitHub alters specific permissions required).

From search and AI summary:
The "Resource not accessible by integration" error typically means that the GitHub Actions workflow does not have the necessary permissions to access the resource it is trying to use. To resolve this, check the permissions for the GitHub Actions integration and ensure that the GITHUB_TOKEN has the required access rights.

Understanding the Error
The "Resource not accessible by integration" error typically occurs in GitHub Actions when the workflow does not have the necessary permissions to access a resource. This can happen for various reasons, especially after actions like deleting and recreating a repository.
Common Causes
1. Permissions Issues
Integration Permissions: The GitHub Actions integration may not have the required permissions to access the repository or perform specific actions. This is common when a repository is recreated with the same name.
GITHUB_TOKEN Limitations: The default GITHUB_TOKEN may not have sufficient permissions for certain actions, especially if the workflow is triggered by a pull request from a forked repository.
2. Workflow Configuration
Incorrect Workflow File: Ensure that the workflow file correctly references the repository and has the right settings for the actions being performed.
Token Scope: If using a GitHub App token, verify that it has the necessary scopes to perform the actions required in the workflow.

